PR #3, opened 2015-12-03
[RFC] Who tests the tests?

modocache:
First of all, thanks so much for open-sourcing this!! :cupid:

I noticed this project doesn't have GitHub issues enabled, so I hope you'll consider this an RFC where we can discuss the issue further. Please let me know if I should email the swift-corelibs-dev mailing list instead.

I noticed that this project isn't *itself* unit tested. It'd be nice to add some tests, since those would enable us to work on it while avoiding regressions.

I could think of several ways to do test this project. The simplest would be running this version of XCTest on some sample programs and examining the output. Those programs would call `XCTMain()` from within their `main()` function. Test output would be printed to stdout via the `print()` function. Tests for this project would then verify that what we expected was indeed printed to stdout.

Another approach would be what's currently in this pull request: allowing consumers of XCTest to pass a list of reporters to their `XCTMain()` function. These reporters would respond to a set of messages, similar to `XCTestObservation.h` in the Apple version of XCTest. From within this project's unit tests, we could then pass a `TestReporter` that captures the messages it's sent. We can then verify it was sent the expected messages.

Let me know your preferred approach, or any other thoughts you may have. Thanks!

### DIFF
--- a/XCTest/XCTest.swift
+++ b/XCTest/XCTest.swift
@@ -17,6 +17,31 @@ import Glibc
 import Darwin
 #endif
 
+/// Not everyone wants to be notified of test results in the same way.
+///
+/// Some people prefer a concise view that fits on a single line, with
+/// dots indicating success and "F" to signify failure: ".....FF..F"
+///
+/// Some prefer a detailed view, containing data on how long each test took
+/// to run, etc.
+///
+/// Objects that conform to the reporter protocol are responsible for
+/// formatting output related to test execution.
+public protocol Reporter {
+    /// Sent when XCTest wishes to log arbitrary information regarding the
+    /// a test event.
+    ///
+    /// - Parameter message: The message to log.
+    func log(message: String)
+}
+
+/// This reporter simply prints each message it is given to stdout.
+private struct StdOutReporter: Reporter {
+    func log(message: String) {
+        print(message)
+    }
+}
+
 public protocol XCTestCaseProvider {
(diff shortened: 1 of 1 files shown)


parkera:
Testing the test framework is absolutely something we need to get started.

Let's discuss this on the `swift-corelibs-dev` list. You can also open an issue on our [issue tracker](https://bugs.swift.org).

modocache:
Gotcha! I sent an email to the mailing list with the title "[xctest] Who tests the tests?". Feel free to close this, or we can keep it open as I push more commits to it. Thanks! :100:

mike-ferris-apple:
Thanks for your interest, and I agree with Tony that getting testing of the testing framework should be a strong goal. For the pull requests, I think we would rather see small targeted and orthogonal changes. For example, I'd be happy to take the change to make the indent prefs explicit in the Xcode project immediately. The notion of reporters can be treated separately, and if there's other things you want to propose and discuss we can treat those separately as well.

modocache:
> I'd be happy to take the change to make the indent prefs explicit in the Xcode project immediately. 

Thanks for clarifying, @mike-ferris-apple! I was worried you might consider the Xcode project file change too trivial to warrant a pull request, but now that I see that's not the case, I've submitted it as #5.

parkera:
I'm going to go ahead and close this pull request for now while the discussion continues on the mailing list.

modocache:
Good call, thanks!

PR #4, opened 2015-12-04
[build_script] Use argparse for required options

modocache:
Using `assert` to check that optional arguments like `--swiftc` are
provided has two downsides:

1. The user sees a stack trace, instead of a helpful error message.
2. The build script must write its own logic to check for arguments,
   rather than argparse's declarative API.

Replace the asserts with argparse's `required` parameter: https://docs.python.org/3/library/argparse.html#required

> Note that the argparse documentation considers `required` options bad
  form. However, I believe they are justified in our case, since
  otherwise the build script would take three unnamed arguments, which
  would be hard to understand.

### DIFF
--- a/build_script.py
+++ b/build_script.py
@@ -29,18 +29,21 @@ def main():
                         metavar="PATH",
                         action="store",
                         dest="swiftc",
+                        required=True,
                         default=None)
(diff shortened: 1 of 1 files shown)


parkera:
Looks good, thanks.

PR #6, opened 2015-12-04
[README] Use correct build script name

modocache:
The script used to build just XCTest is named "build_script.py", not "build_xctest.py".


parkera:
Good catch, thanks.

PR #7, opened 2015-12-05
XCTAssertEqualWithAccuracy behavor does not match true XCTest.framework

mattrajca:
The accuracy threshold check only works with unsigned numbers, however the `distanceTo` method we get from `Strideable` works in terms of offsets which can be negative. We need to take the absolute value of the distance.

This issue is causing tests for https://github.com/apple/swift-corelibs-foundation/pull/37 to fail.


parkera:
Thanks!

mattrajca:
@parkera: As an aside, I find it really weird that `distanceTo` can return a negative number. This tripped me up in some of my work in artificial intelligence as well. I would expect a distance function to always return a non-negative value. Perhaps it should be renamed to something like `offsetTo`: rdar://problem/20667836

parkera:
@mattrajca This is great feedback for the discussion on Swift 3 naming conventions. Part of that effort is to audit the Swift standard library for names like this. I think swift-dev or swift-evolution would be a great place to discuss it.

PR #9, opened 2015-12-06
Implement throw/no throw assertions

davelyon:
Implement throw/no throw assertions

Adds implementation of the following:

``` swift
func notThrowing() throws { }
func willThrow() throws { throw XCTestError(code: 1) }

// Fails with any error thrown
XCTAssertNoThrows(try notThrowing())

// Fails with any error of the specific Type thrown
XCTAssertNoThrowsSpecific(try notThrowing(), XCTestError.self)

// Fails with an instance of the Type that passes a check for equality
XCTAssertNoThrowsSpecific(try notThrowing(), XCTestError(code: 1))

// Fails without an error thrown
XCTAssertThrows(try willThrow())

// Fails without an error thrown, or an error of a different Type
XCTAssertThrowsSpecific(try willThrow(), XCTestError.self)

// Fails unless a specific error is raised, and passes an equality check
XCTAssertThrowsSpecific(try willThrow(), XCTestError(code: 1))

```

This differs from the existing Objective-C API as the model for error handling in Swift is different.

`XCTAssertNoThrowSpecificNamed` isn't as useful in Swift where an error could be an `enum`, `struct`, or `class`. However, by requiring `Equatable` for the more specific `ThrowsSpecific` similar (potentially more flexible) functionality is enabled.

I was able to test these changes locally (manually), however if there is further testing I could add, or other checks to pass I'm happy to update this pull request.

Thanks!

### DIFF
--- a/XCTest/XCTest.swift
+++ b/XCTest/XCTest.swift
@@ -239,3 +239,57 @@ public func XCTAssertTrue(@autoclosure expression: () -> BooleanType, _ message:
 public func XCTFail(message: String = "", file: StaticString = __FILE__, line: UInt = __LINE__) {
     XCTAssert(false, message, file: file, line: line)
 }
+
+public func XCTAssertNoThrows(@autoclosure expression: () throws -> Void, file: StaticString = __FILE__, line: UInt = __LINE__) {
+    do {
+        try expression()
+    } catch let error {
+        XCTFail("Expected no error, got \(error)", file: file, line: line)
+    }
+}
+
+public func XCTAssertNoThrowsSpecific<T: ErrorType>(@autoclosure expression: () throws -> Void, _ errorType: T.Type, file: StaticString = __FILE__, line: UInt = __LINE__) {
+    do {
+        try expression()
+    } catch {
+        XCTAssertFalse(error is T, "Expected not \(T.self), got \(T.self)", file: file, line: line)
+    }
+}
+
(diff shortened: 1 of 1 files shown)


mrackwitz:
:+1: As I have a test library [CatchingFire](https://github.com/mrackwitz/CatchingFire), which implements similar functionality and implemented that in Nimble as well, I was about to propose the same changes as I think that it is really needed.

I think though it would be preferable here to break the convention of the parameter order used in all other assert functions and have for these functions the `expression` to test as last argument consequently, so that the trailing closure syntax can be used. e.g.
```swift
XCTAssertThrowsSpecific(XCTestError.self) {
    let product = try notThrowingProducer()
    try throwingConsumer(product)
}
```

mike-ferris-apple:
I think this is going to need some discussion on the dev list. As you note, throwing in Swift is different from throwing in Obj-C and in the interests of remaining API compatible between the corelibs version and original versions of XCTest, we may need to consider new API here.

As it happens, we are currently considering new API specifically for dealing with Swift error handling in the original XCTest frameworks. And we consider this a pretty high priority area to address. I will make sure that the folks here who are working on this will participate in discussion here about the best course of action.

Mike

parkera:
Following up on what @mike-ferris-apple said -- closing this pull request for now pending further discussion on the `swift-corelibs-dev` and `swift-evolution` mailing lists.

PR #12, opened 2015-12-07
Splitting implementation into several files in anticipation of growth of the code base.

mike-ferris-apple:
My general goal here is to go by the general pattern of defining things in their own files. I have split the assertion functions into XCTAssert.swift, I have also created separate a file for XCTestCaseProvider and another for the XCTestCase protocol and its extension. Finally XCTest.swift remains with the XCTMain function and some attendant global state and the structs used to track it.


parkera:
Foundation's test build & pass on OS X and Linux with this change.

PR #22, opened 2015-12-23
Print additional details when emitting assertion failure messages

briancroom:
This brings the output in line with what Darwin's XCTest outputs

Take a failing assertion like:
```swift
func testAssertEqual() {
    XCTAssertEqual(1, 2, "message", file: "test.swift")
}
```
The **current behavior** of `swift-corelibs-xctest` is to produce a failure message like:
```
test.swift:36: error: Test.testAssertEqual : message
```
**Darwin's XCTest** produces:
```
test.swift:36: error: -[Test.Test testAssertEqual] : XCTAssertEqual failed: ("Optional(1)") is not equal to ("Optional(2)") - message
```
**After this patch**, the output from `swift-corelibs-xctest` would be:
```
test.swift:36: error: Test.testAssertEqual : XCTAssertEqual failed: ("Optional(1)") is not equal to ("Optional(2)") - message
```

Note that this now includes the name of the failing assertion function and includes details about the asserted expressions when appropriate.

This change just screams for test coverage, but I guess that will have to wait until #20 or similar gets merged. In the meantime, though, [here](https://gist.github.com/briancroom/fd6029ad29a429596a9a) is a test fixture that I wrote while working on this, and have been using to manually verify the new behavior.

### DIFF
--- a/Sources/XCTest/XCTAssert.swift
+++ b/Sources/XCTest/XCTAssert.swift
@@ -10,9 +10,88 @@
 //  XCTAssert.swift
 //
 
-/// The primitive assertion function for XCTest. All other XCTAssert* functions
-/// are implemented in terms of this. This function emits a test failure if the
-/// general Bool expression passed to it evaluates to false.
+private enum _XCTAssertion {
+    case Equal(Any, Any)
+    case EqualWithAccuracy(Any, Any, accuracy: Any)
+    case GreaterThan(Any, Any)
+    case GreaterThanOrEqual(Any, Any)
+    case LessThan(Any, Any)
+    case LessThanOrEqual(Any, Any)
+    case NotEqual(Any, Any)
+    case NotEqualWithAccuracy(Any, Any, accuracy: Any)
+    case Nil(Any)
+    case NotNil
+    case True
+    case False
+    case Fail
+
+    var name: String? {
+        switch(self) {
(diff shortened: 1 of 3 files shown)


ddunbar:
It would be great to get this in, the current failure messages are very unfortunate when we see CI failures using a Linux XCTest.

parkera:
This looks good to me. @mike-ferris-apple should take a look too.

briancroom:
I've just rebased this now.

mike-ferris-apple:
This looks good and some testing with Xcode's XCTest shows that the output is not almost the same for these. The one exception seems to be XCTAssertNil where you have the failureDetails returning "\"not nil\"".  To get the same output as Xcode's XCTest, this should return nil. Can you make that tweak? Then I'll definitely merge this! Thanks for providing this, it will make the output a ton more useful.

briancroom:
Good catch @mike-ferris-apple! I think I confused myself by choosing `"notnil"` as the object in the fixture I was working off of. :grin: 

I went back to closely compare the output from Xcode's XCTest again, and realized that *does* include the actual value of the expression in the message, e.g.:
```
test.swift:80: error: -[Test.Test testAssertNil] : XCTAssertNil failed: "MyString" - message
```
I have updated the PR to properly capture and emit this expression value in the failure message.

mike-ferris-apple:
Thanks!

PR #21, opened 2015-12-22
Prevent div-by-0 when producing test execution summary

briancroom:
This is a fix for [SR-334](https://bugs.swift.org/browse/SR-334), in which a `NaN` value would appear in the final summary line at the end of test execution:

```
Total executed 0 tests, with 0 failures (0 unexpected) in -nan (0.0) seconds
```
This is a simple fix for the issue, but I have greater doubts about the calculation of the time values that are being printed. The current code prints `... in {average test duration} ({total test duration}) seconds`, however the two figures printed by the original XCTest seem to be something closer to `... in {total time spent executing tests} ({total elapsed time since XCTest began running}) seconds`, or similar. Does anyone have more insight into what XCTest provides here?

@modocache

PR #23, opened 2015-12-23
Adjust duration output from the execution summary lines

briancroom:
Instead of outputting the average and summed execution time for each test
run, output the summed execution time of each test run and the total
elapsed time between the beginning and end of the test case's execution.
The new output follows the behavior of Darwin's XCTest framework.

Old behavior example excerpt:
```
Executed 22 tests, with 22 failures (0 unexpected) in 0.091 (2.008) seconds
Total executed 22 tests, with 22 failures (0 unexpected) in 0.091 (2.008) seconds
```
New behavior example excerpt (matching Darwin XCTest):
```
Executed 22 tests, with 22 failures (0 unexpected) in 2.006 (2.015) seconds
Total executed 22 tests, with 22 failures (0 unexpected) in 2.006 (2.016) seconds
```

Other implementation changes:
* Add a typealias `TimeInterval` for `Double` to increase explicitness
* Factor out a helper for measuring the time it takes to execute a piece of code

If this PR is accepted, it makes #21 obsolete, and also resolves [SR-334](https://bugs.swift.org/browse/SR-334).

### DIFF
--- a/XCTest/XCTimeUtilities.swift
+++ b/XCTest/XCTimeUtilities.swift
@@ -17,18 +17,28 @@
     import Darwin
 #endif
 
+internal typealias TimeInterval = Double
+
 /// Returns the number of seconds since the reference time as a Double.
-internal func currentTimeIntervalSinceReferenceTime() -> Double {
+private func currentTimeIntervalSinceReferenceTime() -> TimeInterval {
     var tv = timeval()
-    let currentTime = withUnsafeMutablePointer(&tv, { (t: UnsafeMutablePointer<timeval>) -> Double in
+    let currentTime = withUnsafeMutablePointer(&tv, { (t: UnsafeMutablePointer<timeval>) -> TimeInterval in
         gettimeofday(t, nil)
-        return Double(t.memory.tv_sec) + Double(t.memory.tv_usec) / 1000000.0
+        return TimeInterval(t.memory.tv_sec) + TimeInterval(t.memory.tv_usec) / 1000000.0
     })
     return currentTime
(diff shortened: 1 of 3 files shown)


modocache:
I definitely prefer the new output! I find the new output less surprising--the original discrepancy between `0.091` and `2.008` was confusing. Also, it matches the output of Apple XCTest. Thanks!! :100:

XCTest/XCTimeUtilities.swift
```diff
@@ -17,18 +17,28 @@
...
     return currentTime
 }
 
+/// Execute the given block and return the time spent during execution
+internal func measureTimeExecutingBlock(@noescape block: () -> Void) -> TimeInterval {
+    let start = currentTimeIntervalSinceReferenceTime()
+    block()
+    let end = currentTimeIntervalSinceReferenceTime()
+
+    return end-start
```
modocache: nit: Other binary operations in this codebase, such as `+=`, have a space on each side of the operator. For consistency this should be `return end - start`, with spaces.

briancroom: Fixed!

mike-ferris-apple:
Brian, thanks for the patches that fix up the output of the summaries and of the assertion failures. I'll look at them in detail as soon as I return!

Mike

briancroom:
👍 enjoy the holidays!

mike-ferris-apple:
this looks good to me. Presumably this subsumes the change in pull request #21. Can you please close that one?

PR #24, opened 2015-12-23
[XCTAssert] Document file and line params, as well as where assert functions may be used

modocache:
Although it is rare for people to specify these parameters manually, it is not so rare as to not require documentation; see https://twitter.com/mxcl/status/679719145567588353. Add a note to the documentation that explains when they are useful.

Also adopt documentation conventions from the Swift standard library: use `///` over `/** */`, and only have 80 characters per line.

In addition, the XCTAssert* family of functions require that XCTCurrentTestCase is set, so this pull request documents that as well.

### DIFF
--- a/XCTest/XCTAssert.swift
+++ b/XCTest/XCTAssert.swift
@@ -10,13 +10,57 @@
 //  XCTAssert.swift
 //
 
-/**
-The primitive assertion function for XCTest. All other XCTAssert* functions are implemented in terms of this. This function emits a test failure if the general Bool expression passed to it evaluates to false.
-- Parameter expression: A boolean test. If it evaluates to false, the assertion fails and emits a test failure.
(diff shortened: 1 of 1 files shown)


XCTest/XCTAssert.swift
```diff
@@ -10,13 +10,63 @@
...
+/// general Bool expression passed to it evaluates to false.
+///
+/// - Requires: This and all other XCTAssert* functions must be called from
+///   within a test method, as indicated by `XCTestCaseProvider.allTests`.
+///   Assertion failures that occur outside of a test method will *not* be
+///   reported as failures.
+///
+/// - Parameter expression: A boolean test. If it evaluates to false, the
+///   assertion fails and emits a test failure.
+/// - Parameter message: An optional message to use in the failure if the
```
modocache: This change conflicts with #22--@briancroom and I both fix the `assetion` typo :blush:

PR #25, opened 2016-01-01
Move source files to support swift package manager

keith:
This pull request adds support for swift package manager to corelibs-xctest. I'm submitting this just to show how easy this is in response to this thread I started yesterday on the mailing list which is pending conversation https://lists.swift.org/pipermail/swift-corelibs-dev/Week-of-Mon-20151228/000285.html


Package.swift
```diff
@@ -0,0 +1,3 @@
+import PackageDescription
+
+let package = Package(name: "XCTest")
```
keith: This probably wouldn't be the package name desired for both OS X and Linux. We can update this if we ever want to merge this PR.

ddunbar:
This makes sense to me, but I think we should go ahead and move to the Souces/XCTest/<file> layout, to avoid another rename in the future if we want multiple targets in the project (which seems inevitable).

/cc @mike-ferris-apple

keith:
I've updated this PR to nest the sources in `Sources/XCTest` instead!

mike-ferris-apple:
I don't see a any diff here for the linux build script. Presumably it will also need to be updated.

keith:
I've updated the build script to fix what I believe is the only specific path reference. Building on linux now to be sure.

mike-ferris-apple:
Tanks for fixing the build script and rebasing so it merges clean!

keith:
Definitely! Thank you!

PR #26, opened 2016-01-03
AddsSetUp and tearDown to the XCTestCase protocol.

brownleej:
The XCTestCase invokeTest method calls setUp and tearDown before and after the test case, but these calls are statically bound to the empty versions of setUp and tearDown provided by XCTestCase. If a real test case provides its own setUp and tearDown methods, they don't get called automatically. Adding them to the protocol should change this to a dynamic invocation and call the verisons provided by the real test case, which seems to be the intention behind calling them in the first place.

### DIFF
--- a/XCTest/XCTestCase.swift
+++ b/XCTest/XCTestCase.swift
@@ -12,7 +12,8 @@
 //
 
 public protocol XCTestCase : XCTestCaseProvider {
-    
+    func setUp()
+    func tearDown()
 }
 
 extension XCTestCase {


briancroom:
Hey @brownleej, thanks for opening this PR! I actually encountered this problem myself last weekend and ended up with the same fix as you. I would love to see this get merged.

On this note though, I'll mention that I do have other concerns with the `XCTestCase` protocol which I have written up [here](https://bugs.swift.org/browse/SR-460)

PR #27, opened 2016-01-04
Don't use StaticString for XCTAssert*

briancroom:
This addresses [SR-308](https://bugs.swift.org/browse/SR-308), bringing us in line with Apple's XCTest in using `String` instead of `StaticString` as the type of the `file` parameter on assertions.

Included is a minor refactor of how failure handling is structured in order to allow proper failure reporting before termination in the case of `continueAfterFailure` being `false`. I managed to reduce global state a little in the process.

### DIFF
--- a/XCTest/XCTAssert.swift
+++ b/XCTest/XCTAssert.swift
@@ -17,7 +17,7 @@ The primitive assertion function for XCTest. All other XCTAssert* functions are 
 - Parameter file: The file name to use in the error message if the assertion fails. Default is the file containing the call to this function. It is rare to provide this parameter when calling this function.
 - Parameter line: The line number to use in the error message if the assertion fails. Default is the line number of the call to this function in the calling file. It is rare to provide this parameter when calling this function.
 */
-public func XCTAssert(@autoclosure expression: () -> BooleanType, _ message: String = "", file: StaticString = __FILE__, line: UInt = __LINE__) {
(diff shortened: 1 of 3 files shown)


gribozavr:
Why not change the XCTest overlay instead?  `StaticString` is inherently cheaper than `String`.

modocache:
> Why not change the XCTest overlay instead? 

Oh, this would be awesome.

Slightly related: https://github.com/apple/swift/pull/785. It would be cool to expand the overlay's test suite so that it could catch any regressions in the `StaticString` change.

mike-ferris-apple:
I agree with @gribozavr... I would rather see this redone as a pull request for the XCTest overlay in the swift project instead of this change.

mike-ferris-apple:
Actually I see you already did. I merged #888 in the swift project, so let's go ahead and close this one.

briancroom:
@mike-ferris-apple how do you feel about the refactoring in https://github.com/briancroom/swift-corelibs-xctest/commit/cc2ae5d2c6cace77a70f3d658275ccfd0a7c98ee ? Would you consider merging it if I open a separate PR?

I consider it an improvement due to the reduction of global vars and the removal of the public testFailure function.

mike-ferris-apple:
Ah, I didn’t notice that. Yeah, that seems like an improvement. It also gets rid of that weird public function that shouldn’t be public (and probably doesn’t need to be anymore, but I’ve never gotten around to verifying that…)

Go ahead and do a new PR for that if you want!


Mike

PR #30, opened 2016-01-08
[build_script] Expand user home directory

modocache:
Allow users of the build script to use the `~` shorthand when
specifying paths to `--swiftc` and `--build-dir`.

### DIFF
--- a/build_script.py
+++ b/build_script.py
@@ -71,9 +71,9 @@ def main():
                         default="debug")
     args = parser.parse_args()
 
-    swiftc = os.path.abspath(args.swiftc)
-    build_dir = os.path.abspath(args.build_dir)
-    swift_build_dir = os.path.abspath(args.swift_build_dir)
+    swiftc = os.path.abspath(os.path.expanduser(args.swiftc))
+    build_dir = os.path.abspath(os.path.expanduser(args.build_dir))
+    swift_build_dir = os.path.abspath(os.path.expanduser(args.swift_build_dir))
 
     if not os.path.exists(build_dir):
         run("mkdir -p {}".format(build_dir))


gribozavr:
Why?  The shell should be expanding it, if it is passed unquoted on the command line.  (Try `echo ~`.)

modocache:
```
modocache@ubuntu:~/GitHub/apple/swift-corelibs-xctest$ echo ~
/home/modocache
```

Now here's the interesting part:

```
modocache@ubuntu:~/GitHub/apple/swift-corelibs-xctest$ python
Python 2.7.10 (default, Oct 14 2015, 16:09:02) 
[GCC 5.2.1 20151010] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.path.abspath('~/foo')
'/Users/bgesiak/GitHub/apple/swift-corelibs-xctest/~/foo'
```

ddunbar:
Dmitri's point was that if you call the script like:
  build_script --something ~/somepath
then shell will expand it. Note that if you do
  build_script --something=~/somepath
then the shell won't.

 - Daniel

gribozavr:
```
>>> os.path.abspath('~/foo')
```

Of course.  My point is that, if this string came from the command line, it would have been expanded already.  If it wasn't expanded, it was quoted, which means that the user doesn't want it expanded (e.g., they have a directory named `~`).

modocache:
Oh, I see. I didn't quote my arguments, but I did specify them using `--swiftc=~/foo`:

```
modocache@ubuntu:~/GitHub/apple/swift-corelibs-xctest$ ./build_script.py --swiftc=~/GitHub/apple/build-trunk/Ninja-ReleaseAssert/swift-linux-x86_64/bin/swiftc --swift-build-dir=/home/modocache/GitHub/apple/build-trunk/Ninja-ReleaseAssert/swift-linux-x86_84 --build-dir=/home/modocache/GitHub/tmp/XCTest
xctest-build: /home/modocache/GitHub/apple/swift-corelibs-xctest/~/GitHub/apple/build-trunk/Ninja-ReleaseAssert/swift-linux-x86_64/bin/swiftc -c -g -emit-object /home/modocache/GitHub/apple/swift-corelibs-xctest/Sources/XCTest/XCTAssert.swift /home/modocache/GitHub/apple/swift-corelibs-xctest/Sources/XCTest/XCTestCaseProvider.swift /home/modocache/GitHub/apple/swift-corelibs-xctest/Sources/XCTest/XCTestCase.swift /home/modocache/GitHub/apple/swift-corelibs-xctest/Sources/XCTest/XCTimeUtilities.swift /home/modocache/GitHub/apple/swift-corelibs-xctest/Sources/XCTest/XCTestMain.swift -module-name XCTest -parse-as-library -emit-module -emit-module-path /home/modocache/GitHub/tmp/XCTest/XCTest.swiftmodule -o /home/modocache/GitHub/tmp/XCTest/XCTest.o -force-single-frontend-invocation -module-link-name XCTest
/bin/sh: 1: /home/modocache/GitHub/apple/swift-corelibs-xctest/~/GitHub/apple/build-trunk/Ninja-ReleaseAssert/swift-linux-x86_64/bin/swiftc: not found
```

It appears this differs from `--swiftc ~/foo`, which is expanded. Is this an `argparse` quirk?

modocache:
Ah, @ddunbar beat me to it :laughing: 

Well, if that's the expected behavior then sounds good--I'll close this. Thanks!

ddunbar:
> It appears this differs from --swiftc ~/foo, which is expanded. Is this an argparse quirk?

No, this is just how the shell works.

Compare:
  echo ~/foo
with
  echo --hello=~/foo

ddunbar:
FWIW, I also frequently use things like:
  foo --swiftc=$(echo ~/fooo)
in order to get shell tilde expansion in places where it wouldn't otherwise do it. "$()" here is shell magic to replace with the output of the contents (run as a command).

PR #33, opened 2016-01-09
Refactor to rely less on global state for failure reporting

briancroom:
* Also allows removing a weird public method from XCTestCase

This was originally a piece of #27 but it makes sense to apply standalone.

### DIFF
--- a/Sources/XCTest/XCTestCase.swift
+++ b/Sources/XCTest/XCTestCase.swift
@@ -34,8 +34,18 @@ extension XCTestCase {
         var unexpectedFailures = 0
         let overallDuration = measureTimeExecutingBlock {
             for (name, test) in tests {
-                XCTCurrentTestCase = self
                 let method = "\(self.dynamicType).\(name)"
+
+                var failures = [XCTFailure]()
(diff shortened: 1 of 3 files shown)


Sources/XCTest/XCTestCase.swift
```diff
@@ -45,21 +55,19 @@ extension XCTestCase {
...
+                for failure in failures {
                     failure.emit(method)
                     totalFailures += 1
                     if !failure.expected {
                         unexpectedFailures += 1
                     }
+
+                    let result = failures.count > 0 ? "failed" : "passed"
+
+                    print("Test Case '\(method)' \(result) (\(printableStringForTimeInterval(duration)) seconds).")
```
modocache: I believe this is a regression that causes "passed" to never be output.

Before, the following line appeared **outside** of the `XCTCurrentFailures` loop:

```swift
print("Test Case '\(method)' \(result) (\(printableStringForTimeInterval(duration)) seconds).")
```

Now, it appears **within** the failures for loop. As a result, only the results of *failing* tests are printed.

PR #34, opened 2016-01-11
Osx build script

dunkelstern:
This adds OSX support to the `build_script.py` to allow using the same tests on OSX as on Linux.

Added paragraph to Readme with build instructions on OSX.

`--test` won't work on OSX currently, use the Xcode test target for testing on OSX

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,18 @@ install XCTest in your active version of Swift:
     --test
 ```
 
-To run the tests on OS X, build and run the `SwiftXCTestFunctionalTests` target in the Xcode project.
+Building and installing on OSX:
+
+```sh
+sudo ./build_script.py \
+    --swiftc="/Library/Developer/Toolchains/swift-latest.xctoolchain/usr/bin/swiftc" \
+    --build-dir="/tmp/XCTest_build" \
+    --swift-build-dir="/Library/Developer/Toolchains/swift-latest.xctoolchain/usr" \
+    --library-install-path="/Library/Developer/Toolchains/swift-latest.xctoolchain/usr/lib/swift/macosx" \
+    --module-install-path="/Library/Developer/Toolchains/swift-latest.xctoolchain/usr/lib/swift/macosx/x86_64"
+```
+
+To run the tests on OS X, build and run the `SwiftXCTestFunctionalTests` target in the Xcode project, the `--test` parameter for the build script will not work.
 
 You may add tests for XCTest by including them in the `Tests/Functional/` directory. For an example, see `Tests/Functional/SingleFailingTestCase`.
 
--- a/build_script.py
+++ b/build_script.py
@@ -12,7 +12,7 @@
 # Here is a nice way to invoke this script if you are building locally, and Swift is installed at /, and you want to install XCTest back there
 # sudo ./build_script.py --swiftc="/usr/bin/swiftc" --build-dir="/tmp/XCTest_build" --swift-build-dir="/usr" --library-install-path="/usr/lib/swift/linux" --module-install-path="/usr/lib/swift/linux/x86_64"
 
-import os, subprocess, argparse
+import os, shlex, subprocess, argparse, platform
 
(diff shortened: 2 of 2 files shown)


mike-ferris-apple:
Why not use Xcode (or xcodebuild) on OS X? I am reluctant to take a lot of change to support different ways of building and testing this on OS X, especially since there's no real intent for this library to actually be used on OS X. Having the Xcode project is a nice convenience for working on this for those of us who spend most of our time in OS X, even though the project itself is really meant for use on Linux. But I guess I am curious what problem adding this support to the build script is solving.

dunkelstern:
There is a simple intent: just having a unified build system on OSX and Linux.

If I want to implement tests that run on linux I have to use this XCTest library, if i try to run the same tests on OSX it fails because XCTest on OSX is not the same lib as XCTest on Linux.

Primarily I want my `swift build` command that builds a testing binary work on OSX too.

XCTest on OS X does not know `XCTMain()` and `XCTestCase` on OSX works differently in detecting test cases because of the ObjC runtime that is not available on Linux. (`var allTests : [(String, () -> Void)] { return [ ... ] }`)

And to make installation work the same on OSX and Linux I think that the build script should run on both platforms.

parkera:
Our intent here was to allow the API to be compatible, so that developers may use XCTest on OS X/iOS and swift-corelibs-xctest on other platforms.

parkera:
The `allTests` API in swift-corelibs is something we acknowledge is different, but it would be ignored on XCTest on Darwin.

dunkelstern:
Correct, but currently the API it is not compatible completely and before someone comes up with a autodetection idea for tests it will stay that way, so why not allow using the lib on OSX until this has been sorted out?

parkera:
Which part is not API compatible?

If your tests have to add something *additional*, which is used only on Linux (for example), but otherwise the same code can be built against the Xcode XCTest, then I see that as ok.

I really don't want to start us down the road of a fork at this time.

mike-ferris-apple:
For the aspect of this related to swift build (the package manager), I think this discussion is probably best had with the package manager folks involved. They are currently working on (and discussing early proposals for) how testing support will be integrated with the package manager.

I would like to hear more detail about the API not being compatible between Xcode XCTest and corelibs XCTest. As @parkera mentions, the goal here is for Xcode's XCTest to be used on OS X and corelibs XCTest to be used on other platforms.

I know there are rough edges here, but I also know that the package manager's tests, for example, are currently set up such that this works for them.

dunkelstern:
See the other pull request, i would call that an incompatability :( (`setUp`, `tearDown` and the `override` mess)

modocache:
Thanks for this, @dunkelstern. Sorry it's been sitting around for a while! I think we may have addressed your concerns in the meantime:

- The `setUp` and `tearDown` incompatibility was addressed by #40; now `override` is used by both Apple and corelibs XCTest.
- You may reference the Swift build script to see how you can use `xcodebuild` to "install" SwiftXCTest.framework on OS X: https://github.com/apple/swift/blob/0ae11d4216a6763ff0bc7645a55eb0e9db05dc34/utils/build-script-impl#L1884. By "install", I mean you can pass arguments on the command line to `xcodebuild` to specify where build products are placed.

By linking SwiftXCTest.framework in your Xcode project target's "Link Binary with Libraries" build phase, it's possible to use this framework on OS X--in fact, the SwiftXCTestFunctionalTests target in this repository does just that! You can take a look there for an example.

Keep in mind, however, that this is **not recommended**, and that you **cannot use this framework from within a "test target" in Xcode**. Test targets are run automatically via Xcode and Apple XCTest. Instead, we recommend using the swift-package-manager and its new `swift test` command to run tests on OS X and Linux.

Feel free to comment here or send an email to the swift-dev or swift-corelibs-dev mailing lists if you have any additional comments/questions. In the meantime, we wish to continue using `xcodebuild` on OS X and `build_script.py` on other platforms. The changes in this pull request don't reflect that position, so unfortunately I'm going to close this. :bow:

PR #35, opened 2016-01-12
[XCTestCase] Output "test passed"

modocache:
Fix a regression introduced in #33 that causes the "test passed"
message to never be output. Add a test to prevent future regressions.


briancroom:
Thanks for catching this and cleaning up after me @modocache! Big 👍 for the test coverage.

modocache:
@briancroom My pleasure! Thanks for fighting the good fight to get rid of those globals :muscle:

PR #36, opened 2016-01-12
[xcodeproj] Share test scheme

modocache:
- Allow the `SwiftXCTestFunctionalTests` scheme to be run from the
  command-line, by marking it as a "shared" scheme in Xcode.
- Add instructions on how to run the tests via the command line to the
  README.

Allowing the tests to be run via the command-line on all platforms is
necessary for integrating swift-corelibs-xctest tests into the greater
Swift test suite.
(See: https://github.com/apple/swift/blob/162becc91e1eab9118a17c34f994e79d5983819f/utils/build-script-impl#L1968-L1974)

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,11 @@ install XCTest in your active version of Swift:
     --test
 ```
 
-To run the tests on OS X, build and run the `SwiftXCTestFunctionalTests` target in the Xcode project.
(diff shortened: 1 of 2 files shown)


mike-ferris-apple:
This should be fine. In general shared schemes are not required just for xcodebuild, but without them, a user has to have opened to project at least once to get scheme auto creation to kick in for them.

I am generally in favor of having shared schemes for the standard development workflows, though, and shared schemes are definitely required if anyone were to want to set up CI for the project at some point.

PR #37, opened 2016-01-12
[xcodeproj] Better lit output

modocache:
Because Xcode treats the default lit output as error, simply running
`lit Tests/Functional` causes build failures in Xcode. For this reason,
the script was run with the `-q` quiet parameter. However, this made it
difficult to understand why a test failed.

This commit borrows a page from LLVM (see: https://github.com/apple/swift-llvm/blob/3ebdbb2c7e5ce577363994fd0aa0f8409bc68490/CMakeLists.txt#L329-L332)
and uses `-s` (summarize results), `-v` (verbose), and
`--no-progress-bar`. This combination emits diagnostics that are not
treated as errors by Xcode, *and* they provide a great deal of detail
when tests fail--the best of both worlds!

### DIFF
--- a/XCTest.xcodeproj/project.pbxproj
+++ b/XCTest.xcodeproj/project.pbxproj
@@ -219,7 +219,7 @@
 /* Begin PBXLegacyTarget section */
 		DAA333B51C267AD6000CC115 /* SwiftXCTestFunctionalTests */ = {
 			isa = PBXLegacyTarget;
-			buildArgumentsString = "-q Tests/Functional";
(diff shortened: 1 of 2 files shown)


mike-ferris-apple:
Should this change be made in the linux build script as well?

modocache:
@mike-ferris-apple Good call; I amended the commit to include the same flags for Linux.

I'd love to include swift-corelibs-xctest tests in the greater Swift test suite. At that time, I'll try and figure out a way we can share the `lit` flags used when invoking the Swift test suite, and use them for these tests as well.

PR #39, opened 2016-01-14
Add regression test for XCTAssert* failure messages

briancroom:
This backfills tests for the behavior introduced in #22.

It also fixes a minor formatting regression introduced in https://github.com/apple/swift-corelibs-xctest/commit/e5cf2f401867ab90b1b307b4b825f2e7fafddf51 in which the output of `XCTAssertNil` incorrectly prints the actual value as an optional.

### DIFF
--- a/Sources/XCTest/XCTAssert.swift
+++ b/Sources/XCTest/XCTAssert.swift
@@ -286,7 +286,7 @@ public func XCTAssertNil(@autoclosure expression: () throws -> Any?, @autoclosur
         if value == nil {
             return .Success
         } else {
-            return .ExpectedFailure("\"\(value)\"")
+            return .ExpectedFailure("\"\(value!)\"")
         }
     }
 }
--- a/Tests/Functional/FailureMessagesTestCase/main.swift
+++ b/Tests/Functional/FailureMessagesTestCase/main.swift
@@ -1,0 +1,197 @@
+// RUN: %{swiftc} %s -o %{built_tests_dir}/FailureMessagesTestCase
+// RUN: %{built_tests_dir}/FailureMessagesTestCase > %t || true
+// RUN: %{xctest_checker} %t %s
+// CHECK: Test Case 'FailureMessagesTestCase.testAssert' started.
+// CHECK: test.swift:109: error: FailureMessagesTestCase.testAssert : XCTAssertTrue failed - message
+// CHECK: Test Case 'FailureMessagesTestCase.testAssert' failed \(\d+\.\d+ seconds\).
+// CHECK: Test Case 'FailureMessagesTestCase.testAssertEqualOptionals' started.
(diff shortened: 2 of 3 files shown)


Sources/XCTest/XCTAssert.swift
```diff
@@ -267,7 +267,7 @@ public func XCTAssertNil(@autoclosure expression: () -> Any?, @autoclosure _ mes
         if value == nil {
             return .Success
         } else {
-            return .ExpectedFailure("\"\(value)\"")
+            return .ExpectedFailure("\"\(value!)\"")
```
modocache: Could we reverse the conditional here, to avoid the `!`? Exclamation points make me nervous. :sweat_smile:

How about:

```swift
if let value = expression() {
    return .ExpectedFailure("\"\(value)\"")
} else {
    return .Success
}
```

Tests/Functional/FailureMessagesTestCase/main.swift
```diff
@@ -0,0 +1,197 @@
...
+
+#if os(Linux) || os(FreeBSD)
+    import XCTest
+#else
+    import SwiftXCTest
+#endif
+
+// Regression test for https://github.com/apple/swift-corelibs-xctest/pull/22
+class FailureMessagesTestCase: XCTestCase/*, XCTestCaseProvider*/ {
+    var allTests : [(String, () -> Void)] {
```
modocache: You'll need to rebase onto e5c0cb36bc445d0522f7ec28a8342b2681f68666 and change this to `[(String, () throws -> Void)]`.

Sources/XCTest/XCTAssert.swift
briancroom: I had initially chosen to not reorder it that way because I liked the symmetry of the conditionals between the various assertion functions, but I have no problem flipping it.

Tests/Functional/FailureMessagesTestCase/main.swift
briancroom: Oh nice! Looks like that landed while I was putting this commit together. :confetti_ball:

briancroom:
Rebased and fixed up now.

eschaton:
I strongly prefer the `if` predicate to “match” the intent of the assertion; that’s why I wrote them as `if _ { return .Success } else { return .ExpectedFailure(“_”) }`.

Sources/XCTest/XCTAssert.swift
briancroom: I just reverted it to the original ordering at @eschaton's request. AFAIK there is no clean way to avoid the force-unwrap when doing the `nil` check first like this. Using `guard` could allow keeping the ordering the same, but I feel it be a pretty bad abuse of the expected semantics of that keyword.

Is there an alternative I'm overlooking? @modocache

modocache: Fine with me! I defer to him. :smile:

PR #40, opened 2016-01-14
[SR-460] Provide XCTestCase a class to allow overriding setUp() and tearDown()

briancroom:
### What is it?
This PR renames the `XCTestCase` protocol to `XCTestCaseType` and introduces a new `XCTestCase` class in its place which individual test cases may derive from. A relevant test case is also included (see below).

### Motivation
Providing `XCTestCase` as just a protocol rather than a class is a great concept, however it presents an insurmountable barrier to achieving source-level compatibility with Darwin XCTest with respect to defining `setUp()` and `tearDown()` methods in test cases, because the `override` keyword is required when *inheriting* (Darwin XCTest) but is prohibited when just *conforming to* the protocol (Corelibs XCTest). With this patch, it becomes possible to write a test case utilizing `setUp()` and `tearDown()` which builds with both frameworks. Here is the example from [SR-460](https://bugs.swift.org/browse/SR-460):

```swift
import XCTest

class Test: XCTestCase {
    var allTests: [(String, () -> Void)] { return [("testExample", testExample)] }

    override func setUp() {
        super.setUp()
    }

    override func tearDown() {
        super.tearDown()
    }

    func testExample() {
        XCTAssertEqual(1, 1)
    }
}
```

### Drawbacks
There is one significant caveat to this change: test cases deriving from the new `XCTestCase` class must still explicitly declare conformance to the `XCTestCaseType` protocol before they can be passed to `XCTMain`, which makes this a breaking change for current users of Corelibs XCTest. (The `XCTestCase` base class can't conform to `XCTestCaseType` without losing compile-time checking that individual test cases provide an `allTests` property for their own tests.) I see two reasonable approaches to this for cross-platform tests:

1. Conditionally redeclare `XCTestCaseType` on Apple platforms, and conform to the protocol in-line for each test case. This is essentially what swiftpm is currently doing (see [here](https://github.com/apple/swift-package-manager/blob/master/Tests/XCTestCaseProvider/XCTestCaseProvider.swift#L25) and [here](https://github.com/apple/swift-package-manager/blob/master/Tests/dep/DependencyGraphTests.swift#L16)).
2. Add `XCTestCaseType` conformance in a separate source file (perhaps the `main.swift` entry point which contains the call to `XCTMain`). I can easily imagine running a script as a part of the build process which dynamically generates a source file adding this conformance to each test case class. Such a script could even be made to generate the `allTests` property implementation automatically as well.

### Rejected Alternatives
I considered a couple of variants before settling on the current contents of this patch. Note that neither of these include the `XCTestCaseType` protocol:
* Convert the `XCTestCase` protocol into a class which conforms to `XCTestCaseProvider`. In this version, the base `XCTestCase` class included an empty `allTests` definition. This was rejected both because it required individual test cases to mark their `allTests` definitions as `override`, and because there would be no compile-time checking that an overridden `allTests` was present at all.
* Convert the `XCTestCase` protocol into a class which does *not* conform to `XCTestCaseProvider`. In this implementation, `XCTMain` remained unchanged, and accepted an array of `XCTestCase` classes. This fared poorly because it required a runtime check that individual test cases conform to `XCTestCaseProvider`, which felt like a missed opportunity to leverage the type system.

**I look forward to getting feedback and iterating on this!**
@mike-ferris-apple @eschaton @modocache


modocache:
Oh man, *thank you* for this proposal. The other day I was trying to implement `XCTestObservation` in swift-corelibs-xctest, and realized that it doesn't really work unless we have `XCTestCase` instances, and those have `XCTestRun` attributes. I think this direction would open up a lot of possibilities.

A few questions:

1. In Darwin XCTest, a new test case instance is instantiated for each test run. How would you feel about doing so here as well?
2. Is there an equivalent to `NS_REQUIRES_SUPER` in Swift? It would be nice to remind users to call `super.setUp()` and `super.tearDown()`.
3. It is a shame test cases must inferit/conform to two things: `XCTestCase, XCTestCaseType` is a lot of typing. Upon first glance your second "rejected alternative", of checking at runtime, seems like it has potential... I'll need to stew on this more.

I'm also very interested in what the core team thinks here.

mike-ferris-apple:
I have been intending to make a similar change as well.

If we have XCTestCase as a base class, I am wondering what advantage we really get having the XCTestCase protocol (XCTestCaseType in your proposal), though, and even what benefit we’re really getting from XCTestCaseProvider. I had more or less planned to remove both the protocols in favor of implementing an XCTestCase base class.

For me, these changes were motivated by wanting to implement XCTestObservation (as Brian G. talks about below.)



> On Jan 14, 2016, at 11:49 AM, Brian Gesiak <notifications@github.com> wrote:
> 
> Oh man, thank you for this proposal. The other day I was trying to implement XCTestObservation in swift-corelibs-xctest, and realized that it doesn't really work unless we have XCTestCase instances, and those have XCTestRun attributes. I think this direction would open up a lot of possibilities.
> 
> 
In my change for this I had been avoiding introducing XCTestRun.
> A few questions:
> 
> In Darwin XCTest, a new test case instance is instantiated for each test run. How would you feel about doing so here as well?

For XCTestObservation to be useful, we will need to do this. This will also most likely require that XCTMain change to take XCTestCase classes instead of instances. (And that will require some coordination of changes in the two main projects that are clients of XCTest… corelibs Foundation and the package manager).

> Is there an equivalent to NS_REQUIRES_SUPER in Swift? It would be nice to remind users to call super.setUp() and super.tearDown().
> It is a shame test cases must inferit/conform to two things: XCTestCase, XCTestCaseType is a lot of typing. Upon first glance your second "rejected alternative", of checking at runtime, seems like it has potential... I'll need to stew on this more.
> I'm also very interested in what the core team thinks here.
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/apple/swift-corelibs-xctest/pull/40#issuecomment-171759090>.
> 
Mike

briancroom:
Cool. I had been trying to preserve the protocol to avoid locking things down too much, but it sounds like things are headed towards a place that requires a tighter interface. I've now pushed some modified commits which do away with the protocols as @mike-ferris-apple suggested. The original branch with the `XCTestCaseType` protocol is still available [here](https://github.com/briancroom/swift-corelibs-xctest/tree/XCTestCaseType) for reference.

I have started prototyping some changes that would allow a new XCTestCase instance to be used for each test, but none of that is included in this PR.

mike-ferris-apple:
When we take this pull request, we're going to need existing clients to update code, right? (At least they'll be adding the "override" on their allTests, and on setUp() and tearDown() if they have them.)

We're going to have to coordinate with the foundation and package manager projects to get patches for their tests to go in at the same time these changes go.

When we change the execution semantics to an instance per test, we will also need to make another source breaking change. I think it might be good to do that at the same time so we will hopefully not have to do another coordinated patch of foundation and package manager at that time.

I would recommend we change the signature of XCTMain() to be:

    func XCTMain(testCaseClasses: [XCTestCase.Type])

In addition, I found it useful to change allTests into a dictionary when I was prototyping the change to the execution model as well:

    public var allTests: [String: () -> Void]

How do you feel about making these changes pre-emotively as well to minimize the number of changes we're having to coordinate with the two major clients?

mike-ferris-apple:
One additional detail about allTests  as a dictionary... this removes ordering from allTests. To replace that, I was sorting the tests by name. This makes it less practical to order tests specifically, but actually that's arguably a bad thing to do, especially if/when a test harness starts allowing executing subsets of test cases. Allowing ordering dependencies between test cases work against having suites that can be decomposed cleanly.

parkera:
We can react quickly in Foundation once you make this change, so don't worry about that part (cc @phausler). @ddunbar probably should just have a heads up as well for the package manager.

modocache:
> I found it useful to change allTests into a dictionary

This forces all tests to have unique names. An interesting aspect of the current tuple-based API is that tests need not be uniquely named within a test case. I actually consider this to be a feature; XCTest/Xcode integration implicitly requires test selectors to be unique, which causes subtle problems when dynamically generating test methods. See: https://github.com/specta/specta/pull/83#issuecomment-45304370

I'm not against requiring each test method name to be unique within a test case, provided there's a benefit to doing so. What's the upside here?

phausler:
The one problem with this is that the tests won't be portable to Darwin. Because there will be an override that isn't overriding anything.

I find it quite nice that I can copy/paste most tests across and run them against the Foundation that ships with Mac OS X.

mike-ferris-apple:
Arg!  Good point. Note that part of the motivation here was to solve the opposite problem with setUp() and tearDown().

Not sure what to do here. If we keep that in the TestProvider protocol then all tests need to do the double inheritance (and actually that probably creates its own problems with the portability of tests).

One possibility would be to add allTests to XCTestCase in the overlay so it’s an override in Xcode as well.  I am not thrilled with that and would want to talk it through more here and with the rest of the Xcode XCTest folks as well…

Any other options?

Mike

parkera:
Can setUp and tearDown come from inheritance and allTests come from protocol conformance?

parkera:
We would probably then require an empty protocol of the same name on Darwin to avoid having clients `#ifdef`. That's not really that great either.

mike-ferris-apple:
Perhaps there’s a better way to do this, but in order to prototype the execution model I wound up needing to instantiate an instance just to get the test list. Then as the tests are being run an instance has to be created to actually run each test case. The problem was that you need to use the closure from allTests for the instance that is going to run each test, so the initial instance gathered up the names and then passed each one into the instance that was going to run an individual test and it would look the closure up in allTests for that instance to get a block bound to the right self.  Having the dictionary allowed that lookup not to be linear.

It may be that there’s a cleverer way than I could think of to do this better. It sucks to have to create N+1 instances as well…  If we can’t come up with a different approach, we could keep allTests as an array if we were willing to search it linearly, but that winds up being n-squared on the number of tests since we have to search once for each test.

Mike

mike-ferris-apple:
Yeah, if we’re doing that, I’d rather add allTests to XCTestCase in Xcode… At least then people wouldn’t need to have two things in their inheritance list every time.

Mike

modocache:
> in order to prototype the execution model I wound up needing to instantiate an instance just to get the test list

I believe this is why, in Apple XCTest, `+[XCTestCase testInvocations]` is a class method. That way, you don't need to instantiate a test case in order to gather its tests. Apple XCTest then uses the `-[XCTestCase initWithInvocation:]` initializer to instantiate the test case to run the test. Could we change `allTests` to a static `var` or `func`?

> The problem was that you need to use the closure from allTests for the instance that is going to run each test, so the initial instance gathered up the names and then passed each one into the instance that was going to run an individual test and it would look the closure up in allTests for that instance to get a block bound to the right self.  Having the dictionary allowed that lookup not to be linear.

In my experience, the most time-expensive aspect of Apple XCTest was XPC communication between the XCTest driver, Xcode, and the XCTest host on the iOS/tvOS/watchOS device/simulator. My gut is to not worry too much about linear lookups until they demonstrate themselves to be a performance bottleneck--but that's just my two cents! :money_with_wings: 

> We would probably then require an empty protocol of the same name on Darwin to avoid having clients `#ifdef`. That's not really that great either.

Uh-oh, somebody better tell [Joar](https://twitter.com/joar_at_work). :wink:

I'm hesitant to advocate adding this API to Apple XCTest. Perhaps we could migrate away from `+[XCTestCase testInvocations]` to something that we could use here, but even then, I imagine any API churn on Apple XCTest will be an extremely hard sell.

mike-ferris-apple:
> On Jan 19, 2016, at 11:02 AM, Brian Gesiak <notifications@github.com> wrote:
> 
> in order to prototype the execution model I wound up needing to instantiate an instance just to get the test list
> 
> I believe this is why, in Apple XCTest, +[XCTestCase testInvocations] is a class method. That way, you don't need to instantiate a test case in order to gather its tests. Apple XCTest then uses the -[XCTestCase initWithInvocation:] initializer to instantiate the test case to run the test. Could we change allTests to a static var or func?
> 
NSInvocation has the -setTarget: method, though. In Swift, if you implement allTests as a class property it returns unbound test closures which changes the signature. And, not only that, but getting the typing right proved to be problematic…  You would want testClosure(self: XCTestCase.Type) but when you take a closure for a method on MyTestCaseSubclass you get a closure of type testClosure(self: MyTestCaseSubclass.Type) which is not compatible with the first signature.

Someone with more generic Swift-fu than I might be able to come up with a way around this…

> The problem was that you need to use the closure from allTests for the instance that is going to run each test, so the initial instance gathered up the names and then passed each one into the instance that was going to run an individual test and it would look the closure up in allTests for that instance to get a block bound to the right self. Having the dictionary allowed that lookup not to be linear.
> 
> In my experience, the most time-expensive aspect of Apple XCTest was XPC communication between the XCTest driver, Xcode, and the XCTest host on the iOS/tvOS/watchOS device/simulator. My gut is to not worry too much about linear lookups until they demonstrate themselves to be a performance bottleneck--but that's just my two cents! 
> 
> We would probably then require an empty protocol of the same name on Darwin to avoid having clients #ifdef. That's not really that great either.
> 
> Uh-oh, somebody better tell Joar <https://twitter.com/joar_at_work>. 
> 
> I'm hesitant to advocate adding this API to Apple XCTest. Perhaps we could migrate away from +[XCTestCase testInvocations] to something that we could use here, but even then, I imagine any API churn on Apple XCTest will be an extremely hard sell.
> 
> 
I wouldn’t mind migrating Xcode’s XCTest toward something more swift-y than +testInvocations if we could figure out a good approach. Probably something more block-y. And potentially this could be staged in as a new primitive that is implemented in terms of +testInvocations to avoid disruption, but the first thing would be figuring out what we do want.

Mike

briancroom:
Thanks for the discussion around this guys. I completely agree that we should aim to consolidate the breaking changes required for these two features.

When I was hacking on the instance-per-test concept before ([WIP here](https://github.com/briancroom/swift-corelibs-xctest/commits/new_instance_for_each_test)), I ran into precisely the same issues that @mike-ferris-apple describes. I still have a couple of other ideas to work around the generics-related issues when making `allTests` a class variable - probably involving a thunk/type eraser of some sort when calling into `XCTMain`. I'll update once (if?) I figure out a way to make it syntactically palatable.

briancroom:
Sorry for the slow turnaround on this.

After much wrangling with the type system, I have ended up with two approaches, either of which I would be happy to see merged. I've pushed my preferred option to this PR's branch, and the alternative one to [another branch](https://github.com/briancroom/swift-corelibs-xctest/tree/new_instance_for_each_test). Both branches include additional test cases covering the newly-introduced behaviors.

To recap, the requirements I was aiming to meet are:
* Source code compatibility by allow test cases to override `setUp()` and `tearDown()`, including `super` calls, with same syntax as Xcode's XCTest
* Source code compatibility by ensuring that the `allTests` property need not be marked as `override`
* Behavioral compatibility by executing each test method with a new instance of its test case object

One set of changes is shared by the two branches. The two source code compatibility requirements demand that `XCTestCase` be a class which defines `setUp()` and `tearDown()`, but not `allTests`. However since the test runner requires access to all three of these, I have defined an `XCTestCaseType` protocol which unifies them. This *does* require test cases to declare an extra protocol conformance, however I know of no way around that. I additionally dropped `XCTestCaseProvider` as I failed to see any continued benefit to its existence as an independent protocol.

`XCTMain`'s signature then becomes: `@noreturn public func XCTMain(testCases: [XCTestCaseType.Type])`

The two branches vary in their handling of the `allTests` property.

---
This current branch achieves the goal of turning `allTests` into a static property. It solves the generics-related challenges that @mike-ferris-apple described so clearly above by providing a generic adapter function which performs the type erasure necessary to keep `allTests` non-generic. Test cases' `allTests` declarations thus take the following form:
```swift
static var allTests: [(String, XCTestCaseType throws -> Void)] {
    return [
        ("test_passes", test(test_passes)),
    ]
}
```
In my opinion the only drawback to this approach is that it requires the use of this new `test` function when declaring `allTests`, however I feel that this is fairly minor, considering that this is in a block of code which is specific to Corelibs XCTest users anyway. (I also still think that we should investigate the feasibility of providing a build-time tool that would automatically generate a file adding `allTests` and `XCTestCaseType` conformances, but that's a separate issue.)

---
The [other branch](https://github.com/briancroom/swift-corelibs-xctest/tree/new_instance_for_each_test) keeps `allTests` as an instance property, and is very similar to what I assume @mike-ferris-apple prototype (mentioned above) must have looked like. It suffers from the problem described above of requiring N+1 instantiations of the test case, and needing re-execute `allTests` on each test case instance and looking up tests by name within the returned collection. This branch keeps `allTests` as an array (with the associated extra lookup cost), however it would be trivial to change it to be a dictionary as previously described.

---
I'm looking forward to getting feedback on these branches. Note that in either case, documentation updates (particularly in the `README`) would still be required before any merge would be completed.

mike-ferris-apple:
I wanted to at least reply as it may be a few days before I can look at these in detail.

Conceptually the notion of saying test(testMethod) in the allTests seems like a small price to pay. I’ll try to shop this idea around with some folks here who may not be following this discussion as closely including the foundation and package manager folks.

Thanks Brian!

Mike

briancroom:
Thanks for the note, Mike. 

When you do get a chance to take a closer look, please consider whether these changes are compatible with preparing for the XCTestObservation work you mentioned in your first reply. I wasn't explicitly targeting that because I didn't fully understand the requirements you had in mind. But, for example, might it be necessary to make XCTestCaseType a class protocol?

Tests/Functional/FailingTestSuite/main.swift
```diff
@@ -55,7 +55,9 @@ class FailingTestCase: XCTestCase {
     }
 }
 
+extension PassingTestCase: XCTestCaseType {}
+extension FailingTestCase: XCTestCaseType {}
```
modocache: In the grand scheme of things, I feel this is fairly painless for end-users. :+1: 

If they're already writing cross-platform tests (like the swift-corelibs-foundation and swift-package-manager projects are), then they have a conditionally-compiled file named that's used to call `XCTMain()`. This pull request simply requires those projects to add extensions like these in that file, one for each of their test cases.

Once we can use runtime reflection to automatically discover tests, we can simplify this quite a bit--we won't need to jump through hoops to ensure we don't `override` `allTests`. In the meantime, this seems like the best way to proceed.

Sources/XCTest/XCTestCaseType.swift
```diff
@@ -0,0 +1,119 @@
...
+        }
+    }
+
+    public static func invokeTest() {
+        let tests = self.allTests
+        var totalDuration = 0.0
+        var totalFailures = 0
+        var unexpectedFailures = 0
+        let overallDuration = measureTimeExecutingBlock {
+            for (name, test) in tests {
```
modocache: The `test` variable here shadows the `test()` function above.

At first I wondered whether the `test()` function should be renamed to something less likely to clash, but I actually am quite fond of the name--it's simple and easy for test writers to understand.

Instead, how about renaming this variable? Perhaps `testClosure` or `wrappedTest` or something?

Sources/XCTest/XCTestCaseType.swift
```diff
@@ -0,0 +1,119 @@
...
+/// the signature required by XCTestCaseType.allTests
+public func test<T: XCTestCaseType>(testFunc: T -> () throws -> Void) -> TestMethod {
+    return { testCaseType in
+        guard let testCase: T = testCaseType as? T else {
+            fatalError("Attempt to invoke test on class \(T.self) with incompatible instance type \(testCaseType.dynamicType)")
+        }
+
+        try testFunc(testCase)()
+    }
+}
```
modocache: Kudos on the Swift-fu here! :key: 

One suggestion for this pull request: by merging this we broaden the API surface area considerably. In addition to the current API, users will have to understand `XCTestCaseType` and the `test()` function. I think it's very important that the documentation for each of these types is approachable and user-friendly.

For example, the documentation for this function is very clear and explains the function's purpose. However, a user completely new to swift-corelibs-xctest may not understand where and why to use it. I'd like to consider providing an example, like so:

```swift
/// Wraps a static reference to an instance method on a test case such that it can be run by XCTest.
/// Use this when specifying your test methods in `XCTestCaseType.allTests()`:
///
///     class MyTestCase: XCTestCase {
///         static var allTests: [(String, XCTestCaseType throws -> Void)] {
///             ("testSomething", test(testSomething)),
///         }
///         func testSomething() {
///             // ...
///         }
///     }
```

Similarly, it might be nice to provide documentation for `XCTestCaseType`, to answer the question: "where should I use this public protocol?"

briancroom: Totally agreed about documentation being critical here. I was just holding off on fleshing that out until a consensus was reached on whether this API direction seems reasonable.

mike-ferris-apple:
Summarizing the changes that implementors have to make:
* Update allTests to be static and have the new signature
* Update the tuples in allTests to wrap the method name mention with test()
* Update the XCTMain call site to add conformance to the XCTestCaseType protocol to each XCTestCase
* Update XCTMain itself to pass types (replace the instantiation "()" with ".self" after each class name.

I think I get why all this is necessary, and I agree with @modocache that this seems pretty reasonable, especially considering that the long term goal would be that all this evaporates when we get a solution to automatic discovery.

I am not sure if the observer stuff will require anything more... The observer protocol API that we're in a position to implement are the methods that take an XCTestCase instance. Since invokeTest is implemented on the protocol, this would currently require some down casting... Not sure if there's a way around that?

mike-ferris-apple:
What about this proposal wouldn't work if we removed the XCTestCaseType protocol and phrased everything in terms of XCTestCase, the class?

I know that we have the issue around allTests needing to override in corelibs but not in Xcode.  Is there anything else that could not be made to work with that approach?

briancroom:
@mike-ferris-apple As you mentioned, the primary concern with scrapping the protocol would be the 'override' keyword. If that issue were worked around by e.g. adding an allTests declaration in the Swift stdlib overlay, the only issue I can think of would be that we would lose compile-time verification that subclasses actually have overridden allTests with their own data. That is a fairly minor concern as it is simple to introduce a failure at runtime in that situation.

briancroom:
Hey @mike-ferris-apple, what do you think is the way forward here? Have you had a chance to talk to any of the folks working on Xcode XCTest about adding a placeholder allTests property on that side? Would you like me to go ahead and update this PR to remove XCTestCaseType again?

mike-ferris-apple:
I am still trying to get some consensus around adding allTests to the overlay, but I think that would help bring the two versions a bit closer together and I do worry that if we don't make XCTestCase be the main player in terms of types that we'll have further hassles as we look at things like the observation protocol and possibly other APIs that we will likely want to see implemented.

mike-ferris-apple:
OK, I think we have consensus that we'd rather go the route of adding allTests to Xcode's XCTest Swift overlay in order to be able to make XCTestCase the class be the base type that we're using throughout. Would you be willing to try that approach and revise the pull request?

briancroom:
Certainly! I've got the changes in-progress locally and just need to find a little more time to clean it up and update the documentation around how to use the allTests property and XCTMain for XCTest users.

briancroom:
Alright, I've rebased everything and updated it to do away with the `XCTestCaseType

Sources/XCTest/XCTestCase.swift
```diff
@@ -8,12 +8,58 @@
...
 //
 //  XCTestCase.swift
-//  Base protocol (and extension with default methods) for test cases
+//  Base class for test cases
 //
 
-public protocol XCTestCase : XCTestCaseProvider {
-    func setUp()
-    func tearDown()
+public typealias XCTestMethod = XCTestCase throws -> Void
```
modocache: OK, so call me crazy, but I'd love to see documentation on anything marked `public`. I'd be open to hearing dissenting opinions, or even deferring adding documentation until a later pull request.

Sources/XCTest/XCTestCase.swift
```diff
@@ -8,12 +8,58 @@
...
+    /// to tell us the list of tests by implementing this property. This should look like:
+    ///
+    ///     override class var allTests : [(String, XCTestMethod)] {
+    ///         return [
+    ///             ("test_foo", test(test_foo)),
+    ///             ("test_bar", test(test_bar)),
+    ///         ]
+    ///     }
+    public class var allTests: [(String, XCTestMethod)] {
+        return [("missingAllTestsOverride", { _ in throw XCTestError.MissingAllTestsOverride })]
```
modocache: Ingenious! I was expecting to see the base implementation call `fatalError()`, but this approach is much better. :+1:

Tests/Functional/MissingAllTestsOverride/main.swift
```diff
@@ -0,0 +1,22 @@
+// RUN: %{swiftc} %s -o %{built_tests_dir}/MissingAllTestsOverride
+// RUN: %{built_tests_dir}/MissingAllTestsOverride > %t || true
+// RUN: %{xctest_checker} %t %s
+// CHECK: Test Case 'MissingAllTestsOverride.missingAllTestsOverride' started.
+// CHECK: <EXPR>:0: unexpected error: MissingAllTestsOverride.missingAllTestsOverride : threw error "This test case class must override the allTests property" -
```
modocache: Awesome!

I don't know if this is technically possible, but is there any way we can get `<EXPR>:0` to instead point to a relevant file and line? It would be *amazing* if we could get this to point to the file and line `class MissingAllTestsOverride` is defined, for example.

build_script.py
```diff
@@ -92,7 +92,6 @@ def main():
 
     sourceFiles = [
                    "XCTAssert.swift",
-                   "XCTestCaseProvider.swift",
                    "XCTestCase.swift",
                    "XCTimeUtilities.swift",
                    "XCTestMain.swift",
```
modocache: I've been meaning to change this to a `glob.glob()` call. That way we won't have to modify this list of files every time.

modocache:
I agree, I also think this could be merged as-is. I had two suggestions on how it could potentially be improved, but those could be addressed in a separate pull request.

Excellent work, @briancroom! :100:

Sources/XCTest/XCTestCase.swift
briancroom: I've added a couple more doc entries to public symbols now, just for you :smile:

Tests/Functional/MissingAllTestsOverride/main.swift
briancroom: I don't *think* that is possible, at least not without calling out to some sort of symbolication utility.

Currently this is using the unhandled-error functionality that @eschaton added a while back, producing the same sort of output as here: https://github.com/apple/swift-corelibs-xctest/blob/master/Tests/Functional/ErrorHandling/main.swift#L13

modocache: Aha, neat. Perhaps it would be cool to explore populating those with file and line numbers somehow. A discussion for another time, to be sure!

briancroom:
Well..I've hit another snag in our plan when I was working on adding `allTests` to the XCTest overlay as discussed. Namely:

screen shot 2016-02-14 at 1 09 13 pm

As far as I can tell, this marks a dead-end for the plan of `allTests` in the XCTest overlay. I do have one more idea for an alternative approach that I'm exploring, besides the previously-proposed solution involving the `XCTestCaseType` protocol.

briancroom:
I'm close to having another attempt ready for this (which doesn't require modifications to the Swift XCTest overlay!), but in the meantime I thought I'd point out this relevant proposal from @jrose-apple: https://github.com/apple/swift-evolution/pull/155

mike-ferris-apple:
I talked to Jordan and it is unlikely that our particular aspect of this extension/overrie business will be resolved in the near future. It's possible that a solution is in sight for extensions in the same module as the base class (so far so good) but that it won't (yet) cover extensions on Obj-C classes (which is our case here). I'm interested to see your new attempt.

jrose-apple:
Yep. That's the right proposal, but it hasn't even gotten into formal review yet, and being accepted doesn't mean it'll get implemented promptly either.

briancroom:
@mike-ferris-apple @jrose-apple yup, for sure. I had just thought it was interesting timing that the proposal was posted so shortly after I encountered the issue here.

I've updated the branch with a New & Improved™ variant (past versions available [here](https://github.com/briancroom/swift-corelibs-xctest/tree/allTests_as_override) and [here](https://github.com/briancroom/swift-corelibs-xctest/tree/XCTestCaseType) for reference) which checks most of the boxes of goals for this work, while simultaneously not requiring any changes to the XCTest overlay.

To summarize:
* Removed protocol usage (no `XCTestCaseProvider` or `XCTestCaseType`)
* `XCTestCase` is now a class, allowing `setUp()` and `tearDown()` to be overridden methods,
* The `allTests` array on test cases becomes static, facilitating the instantiation of a new test case instance for each test invocation
* Changed from the previous iteration is that the type erasure function is now used during the call to `XCTMain` instead of on the individual test methods in `allTests`.

This is an updated list of the changes that implementors have to make to work with the new interface:
* Update `allTests` to be static and have the new signature (note that the signature now includes the name of the current test case subclass, so it isn't the exact same for all classes anymore)
* Update the `XCTMain` call to reference test cases like: `XCTMain([testCase(TestMe.allTests])`

The one element of this that I'm not entirely satisfied with is that the `allTests` property is no longer specified by any class or protocol in the framework, so its name is just convention. OTOH this could be reasonably regarded as a good thing, although it opens the door for inconsistency. Nonetheless, I am overall very pleased with the state of this branch now.

Also, while testing this out to make sure it would work for existing clients, I went ahead and created branches on [swift-corelibs-foundation](https://github.com/briancroom/swift-corelibs-foundation/commit/6f612afb10817d91c8626914f78e71b8bb6dd888) and [swift-package-manager](https://github.com/briancroom/swift-package-manager/commit/0ce5baa6da43cd325b214767b5e9f285eeb04c87) that make them work with these changes. Assuming folks here are comfortable with this branch, I can go ahead and open PR's on the other projects anytime.

/cc @parkera @phausler @mxcl @ddunbar

Sources/XCTest/XCTestCase.swift
```diff
@@ -26,46 +58,46 @@ extension XCTestCase {
             // TODO: When using the Objective-C runtime, XCTest is able to throw an exception from an assert and then catch it at the frame above the test method. This enables the framework to effectively stop all execution in the current test. There is no such facility in Swift. Until we figure out how to get a compatible behavior, we have decided to hard-code the value of 'true' for continue after failure.
         }
     }
-    
-    public func invokeTest() {
```
modocache: `-[XCTestCase invokeTest]` is a public method in Apple XCTest (although the Apple documentation does state that "in general this should not be called directly".

Removing this method here isn't necessarily a blocker for merging this pull request, but it is one step backwards in terms of achieving API parity. Anyway just consider this a footnote of mine, as personally I think merging this pull request in order to get a correct `override func setUp()` is more important.

modocache:
Bravo! Fantastic work as always. :bouquet: 

> - Update `allTests` to be static and have the new signature (note that the signature now includes the name of the current test case subclass, so it isn't the exact same for all classes anymore)
>
> [...]
>
> The one element of this that I'm not entirely satisfied with is that the `allTests` property is no longer specified by any class or protocol in the framework, so its name is just convention. 

These points are also my primary concerns with this pull request. I think the proposed version of `allTests` makes using swift-corelibs-xctest more difficult. But it's also my understanding that we are continuing to pursue a solution in which `allTests` is automatically determined via reflection of some kind, and I'm willing to sacrifice ease of use in the meantime for the sake of API parity.

All in all, this looks like it should be merged to me! :+1:

Sources/XCTest/XCTestCase.swift
briancroom: Thanks for pointing that out, @modocache, I hadn't considered that method being a part of the XCTestCase API. Having looked at the docs now though, it seems that  the current semantics of this method (run all the test methods in the test case) are significantly different than that of Xcode XCTest's (run a single test method). I would thus look at this as removing a potential point of confusion until such time as proper API compatibility may be attempted.

briancroom:
@mike-ferris-apple I'd love to get your feedback on the latest round of changes here when you get a chance! If we decide this is the way to go, I'll happily rebase the swiftpm and Foundation PRs to keep internal breakage at a minimum 🙂

mike-ferris-apple:
So looks like the trick now is that allTests is declared as a static and, as you say, "allTests" becomes convention. It could really be called anything as it is passed to XCTestMain as a block.

Needing to declare allTests with the repeated type in the signature is not ideal, but all things considered, this seems like an approach that actually gets us the most important stuff. We get the Xcode XCTest compatibility with respect to override keywords on things like setUp(). We get XCTestCase as a class and we get an instance per test method.

The downsides are all in the parts that are necessary only because we don't have automatic test discovery.

Did you say you already have pull requests for getting Foundation and swiftpm converted over to this?Go ahead and rebase them.

@parkera, @ddunbar, @mxcl: If you guys OK with these changes, we should coordinate a time to merge all 3 pull requests.

czechboy0:
@mike-ferris-apple I'm trying to tackle the automatic discovery here: https://github.com/apple/swift-package-manager/pull/156 I see that with these changes I'd need to slightly modify the way test manifests are generated, so please let me know when that happens.
Honza

parkera:
@czechboy0 interesting idea; I hadn't thought of using SwiftPM to find the test cases instead of building it into metadata in the runtime.

parkera:
@mike-ferris-apple This seems reasonable to me. CC @phausler

phausler:
the idea seems pretty solid as an approach. What would we need to do for swift-corelibs-foundation? How can one disable tests with this approach? or exclude them from certain platforms? So this means we can get rid of the main.swift in our test project and the `allTests` perhaps? If we can do that a big :+1:

briancroom:
@mike-ferris-apple I just opened https://github.com/apple/swift-package-manager/issues/158 and https://github.com/apple/swift-corelibs-foundation/issues/270 with the required changes for those projects. This should allow us to move forward with merging these changes in even before @czechboy0's (exciting!) work is ready to go.

czechboy0:
@briancroom :+1: Only a small change will be required in the new test manifest generator once this gets merged.

mike-ferris-apple:
OK. I checked with @mxcl and @parkera and got the go ahead.  Assuming all three of these patches can be cleanly merged in the morning, I will do it then.

mike-ferris-apple:
@swift-ci please test

briancroom:
I just rebased them all now, so hopefully they will still be good to go then.

mike-ferris-apple:
OK. In that case, let's go now.  All three have been merged.

briancroom:
Awesome! Thanks for doing the honors @mike-ferris-apple :tada: 

I appreciate everyone that's been involved in the discussions around this over the past weeks and thanks for your patience through all the iterations :smile:

PR #43, opened 2016-01-24
Asynchronous testing API [AsyncXCTest 6/6]

modocache:
# What's in this pull request?

An attempt to add asynchronous testing support, mirroring the Objective-C XCTest API. This is done primarily via two methods:

- `XCTestCase.expectationWithDescription()`
- `XCTestCase.waitForExpectationsWithTimeout()`

The tests in the pull request pass on Darwin, but not on Linux. Still, I've encountered a few issues that I'd like to get feedback on.

# What's worth discussing about this pull request?

The methods added are identical to their Objective-C SDK overlay counterparts, with the following exceptions:

1. If expectations are created but not waited for, Objective-C XCTest generates a test failure. In order to display a failure on the exact line of the last expectation that was created, Objective-C XCTest symbolicates the call stack. It does so using a private framework called `CoreSymbolication.framework`. We don't have this at our disposal for swift-corelibs-xctest, so instead this pull request has the method take `file` and `line` as parameters, defaulting to `__FILE__` and `__LINE__`. By doing so, we provide the same failure highlighting functionality while also maintaining a (nearly) identical API.
2. If `waitForExpectationsWithTimeout()` fails, Objective-C XCTest generates a test failure on the exact line that method was called, using the same technique from (1). For the same reason, this method also takes `file` and `line` parameters in this pull request.
3. `waitForExpectationsWithTimeout()` takes a handler callback, which in Objective-C XCTest is passed an instance of `NSError`. swift-corelibs-xctest doesn't depend upon swift-corelibs-foundation, and so does not have access to `NSError`. Instead, it defines a shim, `XCTError`.

I'd like to discuss the following:

**(A)** Without using something like `CoreSymbolication.framework`, we cannot emit failures for the correct file or line while maintaining an identical API. This pull request is a sort of middle ground; the *use* of the API looks identical:

```swift
// Objective-C XCTest
let expectation = self.expectationWithDescription("foo")
expectation.fulfill()
self.waitForExpectationsWithTimeout(10, handler: nil)
self.waitForExpectationsWithTimeout(10) { error in
    // ...
}
```

```swift
// swift-corelibs-xctest
let expectation = self.expectationWithDescription("bar")
expectation.fulfill()
self.waitForExpectationsWithTimeout(10, handler: nil)
self.waitForExpectationsWithTimeout(10) { error in
    // ...
}
```

However, they're not totally identical. The following is only possible in swift-corelibs-xctest:

```swift
// swift-corelibs-xctest
let expectation = self.expectationWithDescription("baz", file: "/path/to/file", line: 10)
expectation.fulfill()
self.waitForExpectationsWithTimeout(10, file: "/path/to/file", line: 20, handler: nil)
```

Is this an acceptable solution?

**(B)** The code in this pull request has a dummy implementation to wait for expectations to be fulfilled. For a real implementation, I think we'll need to add a dependency to swift-corelibs-xctest, and I think that should either be [swift-corelibs-libdispatch](https://github.com/apple/swift-corelibs-libdispatch) or [swift-corelibs-foundation](https://github.com/apple/swift-corelibs-foundation).

I think [swift-corelibs-foundation](https://github.com/apple/swift-corelibs-foundation) would be ideal: it includes both:

- `NSRunLoop`, which we can use to implement the waiting logic, and...
- ...`NSError`, which is used by the callback to `waitForExpectationsWithTimeout()`.

However, [swift-corelibs-foundation](https://github.com/apple/swift-corelibs-foundation) uses swift-corelibs-xctest for its own unit test suite. So while the dependency isn't exactly cyclical (Foundation's *tests* depend on XCTest, which depends on Foundation--no cycle), it's still a little murky as to whether it's a good idea.

Should swift-corelibs-xctest depend on [swift-corelibs-foundation](https://github.com/apple/swift-corelibs-foundation)? Should it depend on [swift-corelibs-libdispatch](https://github.com/apple/swift-corelibs-libdispatch)? Or should it depend on neither--and if so, how can we implement asynchronous testing?

/cc @briancroom @jeffh

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,13 +24,15 @@ The rest of this document will focus on how this version of XCTest differs from 
 
 ## Working on XCTest
 
+### On Linux
+
 XCTest can be built as part of the overall Swift package. When following [the instructions for building Swift](http://www.github.com/apple/swift), pass the `--xctest` option to the build script:
 
 ```sh
 swift/utils/build-script --xctest
 ```
 
-If you want to build just XCTest, use the `build_script.py` script at the root of the project. The `master` version of XCTest must be built with the `master` version of Swift.
+If you want to build just XCTest, use the `build_script.py` script at the root of the project. The `master` version of XCTest must be built with the `master` version of Swift. XCTest has a dependency upon Foundation, so you must have built the `master` version of that as well.
 
 If your install of Swift is located at `/swift` and you wish to install XCTest into that same location, here is a sample invocation of the build script:
 
@@ -38,6 +40,7 @@ If your install of Swift is located at `/swift` and you wish to install XCTest i
 ./build_script.py \
     --swiftc="/swift/usr/bin/swiftc" \
     --build-dir="/tmp/XCTest_build" \
+    --foundation-build-dir "/swift//usr/lib/swift/linux" \
     --library-install-path="/swift/usr/lib/swift/linux" \
     --module-install-path="/swift/usr/lib/swift/linux/x86_64"
 ```
@@ -45,8 +48,26 @@ If your install of Swift is located at `/swift` and you wish to install XCTest i
 To run the tests on Linux, use the `--test` option:
 
 ```sh
-./build_script.py --swiftc="/swift/usr/bin/swiftc" --test
+./build_script.py \
+    --swiftc="/swift/usr/bin/swiftc" \
+    --foundation-build-dir "/swift/usr/lib/swift/linux" \
+    --test
 ```
+
+You may add tests for XCTest by including them in the `Tests/Functional/` directory. For an example, see `Tests/Functional/SingleFailingTestCase`.
+
+### On OS X
+
+You may build XCTest via the "SwiftXCTest" scheme in `XCTest.xcworkspace`. The workspace assumes that Foundation and XCTest are checked out from GitHub in sibling directories. For example:
+
+```
+% cd Development
+% ls
+swift-corelibs-foundation swift-corelibs-xctest
+%
+```
+
+Unlike on Linux, you do not need to build Foundation prior to building XCTest. The "SwiftXCTest" Xcode scheme takes care of that for you.
 
 To run the tests on OS X, build and run the `SwiftXCTestFunctionalTests` target in the Xcode workspace. You may also run them via the command line:
 
@@ -54,7 +75,7 @@ To run the tests on OS X, build and run the `SwiftXCTestFunctionalTests` target 
 xcodebuild -workspace XCTest.xcworkspace -scheme SwiftXCTestFunctionalTests
 ```
 
-You may add tests for XCTest by including them in the `Tests/Functional/` directory. For an example, see `Tests/Functional/SingleFailingTestCase`.
+When adding tests to the `Tests/Functional` directory, make sure they can be opened in the `XCTest.xcworkspace` by adding references to them, but do not add them to any of the targets.
 
 ### Additional Considerations for Swift on Linux
 
--- a/Sources/XCTest/XCTestCase.swift
+++ b/Sources/XCTest/XCTestCase.swift
@@ -11,6 +11,12 @@
 //  Base class for test cases
 //
 
+#if os(Linux) || os(FreeBSD)
+    import Foundation
+#else
+    import SwiftFoundation
+#endif
+
 /// This is a compound type used by `XCTMain` to represent tests to run. It combines an
 /// `XCTestCase` subclass type with the list of test methods to invoke on the test case.
 /// This type is intended to be produced by the `testCase` helper function.
@@ -47,6 +53,12 @@ private func test<T: XCTestCase>(testFunc: T -> () throws -> Void) -> XCTestCase
         try testFunc(testCase)()
     }
 }
+
+// FIXME: Expectations should be stored in an instance variable defined on
+//        XCTestCase, but when so defined Linux tests fail with "hidden symbol
+//        isn't defined". Use a global for the time being, as this seems to
+//        appease the Linux compiler.
+private var XCTAllExpectations = [XCTestExpectation]()
 
 extension XCTestCase {
     
@@ -92,6 +104,8 @@ extension XCTestCase {
                 }
 
                 testCase.tearDown()
+                testCase.failIfExpectationsNotWaitedFor(XCTAllExpectations)
+                XCTAllExpectations = []
 
                 totalDuration += duration
 
@@ -121,5 +135,162 @@ extension XCTestCase {
         }
 
         XCTPrint("Executed \(tests.count) test\(testCountSuffix), with \(totalFailures) failure\(failureSuffix) (\(unexpectedFailures) unexpected) in \(printableStringForTimeInterval(totalDuration)) (\(printableStringForTimeInterval(overallDuration))) seconds")
+    }
+
+    /// It is an API violation to create expectations but not wait for them to
+    /// be completed. Notify the user of a mistake via a test failure.
+    private func failIfExpectationsNotWaitedFor(expectations: [XCTestExpectation]) {
+        if expectations.count > 0 {
+            let failure = XCTFailure(
+                message: "Failed due to unwaited expectations.",
+                failureDescription: "",
+                expected: false,
+                file: expectations.last!.file,
+                line: expectations.last!.line)
+            if let failureHandler = XCTFailureHandler {
+                failureHandler(failure)
+            }
+        }
+    }
+
+    /// Creates and returns an expectation associated with the test case.
+    ///
+    /// - Parameter description: This string will be displayed in the test log
+    ///   to help diagnose failures.
+    /// - Parameter file: The file name to use in the error message if
+    ///   this expectation is not waited for. Default is the file
+    ///   containing the call to this method. It is rare to provide this
+    ///   parameter when calling this method.
+    /// - Parameter line: The line number to use in the error message if the
+    ///   this expectation is not waited for. Default is the line
+    ///   number of the call to this method in the calling file. It is rare to
+    ///   provide this parameter when calling this method.
+    ///
+    /// - Note: Whereas Objective-C XCTest determines the file and line
+    ///   number of expectations that are created by using symbolication, this
+    ///   implementation opts to take `file` and `line` as parameters instead.
+    ///   As a result, the interface to these methods are not exactly identical
+    ///   between these environments. To ensure compatibility of tests between
+    ///   swift-corelibs-xctest and Apple XCTest, it is not recommended to pass
+    ///   explicit values for `file` and `line`.
+    public func expectationWithDescription(description: String, file: StaticString = #file, line: UInt = #line) -> XCTestExpectation {
+        let expectation = XCTestExpectation(
+            description: description,
+            file: file,
+            line: line,
+            testCase: self)
+        XCTAllExpectations.append(expectation)
+        return expectation
+    }
+
+    /// Creates a point of synchronization in the flow of a test. Only one
+    /// "wait" can be active at any given time, but multiple discrete sequences
+    /// of { expectations -> wait } can be chained together.
+    ///
+    /// - Parameter timeout: The amount of time within which all expectation
+    ///   must be fulfilled.
+    /// - Parameter file: The file name to use in the error message if
+    ///   expectations are not met before the given timeout. Default is the file
+    ///   containing the call to this method. It is rare to provide this
+    ///   parameter when calling this method.
+    /// - Parameter line: The line number to use in the error message if the
+    ///   expectations are not met before the given timeout. Default is the line
+    ///   number of the call to this method in the calling file. It is rare to
+    ///   provide this parameter when calling this method.
+    /// - Parameter handler: If provided, the handler will be invoked both on
+    ///   timeout or fulfillment of all expectations. Timeout is always treated
+    ///   as a test failure.
+    ///
+    /// - Note: Whereas Objective-C XCTest determines the file and line
+    ///   number of the "wait" call using symbolication, this implementation
+    ///   opts to take `file` and `line` as parameters instead. As a result,
+    ///   the interface to these methods are not exactly identical between
+    ///   these environments. To ensure compatibility of tests between
+    ///   swift-corelibs-xctest and Apple XCTest, it is not recommended to pass
+    ///   explicit values for `file` and `line`.
+    public func waitForExpectationsWithTimeout(timeout: NSTimeInterval, file: StaticString = #file, line: UInt = #line, handler: XCWaitCompletionHandler?) {
+        // Mirror Objective-C XCTest behavior; display an unexpected test
+        // failure when users wait without having first set expectations.
+        // FIXME: Objective-C XCTest raises an exception for most "API
+        //        violation" failures, including this one. Normally this causes
+        //        the test to stop cold. swift-corelibs-xctest does not stop,
+        //        and executes the rest of the test. This discrepancy should be
+        //        fixed.
+        if XCTAllExpectations.count == 0 {
+            let failure = XCTFailure(
+                message: "call made to wait without any expectations having been set.",
+                failureDescription: "API violation",
+                expected: false,
+                file: file,
+                line: line)
+            if let failureHandler = XCTFailureHandler {
+                failureHandler(failure)
+            }
+            return
+        }
+
+        // Objective-C XCTest outputs the descriptions of every unfulfilled
+        // expectation. We gather them into this array, which is also used
+        // to determine failure--a non-empty array meets expectations weren't
+        // met.
+        var unfulfilledDescriptions = [String]()
+
+        // We continue checking whether expectations have been fulfilled until
+        // the specified timeout has been reached.
+        // FIXME: Instead of polling the expectations to check whether they've
+        //        been fulfilled, it would be more efficient to use a runloop
+        //        source that can be signaled to wake up when an expectation is
+        //        fulfilled.
+        let runLoop = NSRunLoop.currentRunLoop()
+        let timeoutDate = NSDate(timeIntervalSinceNow: timeout)
+        repeat {
+            unfulfilledDescriptions = []
+            for expectation in XCTAllExpectations {
+                if !expectation.fulfilled {
+                    unfulfilledDescriptions.append(expectation.description)
+                }
+            }
+
+            // If we've met all expectations, then break out of the specified
+            // timeout loop early.
+            if unfulfilledDescriptions.count == 0 {
+                break
+            }
+
+            // Otherwise, wait another fraction of a second.
+            runLoop.runUntilDate(NSDate(timeIntervalSinceNow: 0.01))
+        } while NSDate().compare(timeoutDate) == NSComparisonResult.OrderedAscending
+
+        if unfulfilledDescriptions.count > 0 {
+            // Not all expectations were fulfilled. Append a failure
+            // to the array of expectation-based failures.
+            let descriptions = unfulfilledDescriptions.joinWithSeparator(", ")
+            let failure = XCTFailure(
+                message: "Exceeded timeout of \(timeout) seconds, with unfulfilled expectations: \(descriptions)",
+                failureDescription: "Asynchronous wait failed",
+                expected: true,
+                file: file,
+                line: line)
+            if let failureHandler = XCTFailureHandler {
+                failureHandler(failure)
+            }
+        }
+
+        // We've recorded all the failures; clear the expectations that
+        // were set for this test case.
+        XCTAllExpectations = []
+
+        // The handler is invoked regardless of whether the test passed.
+        if let completionHandler = handler {
+            var error: NSError? = nil
+            if unfulfilledDescriptions.count > 0 {
+                // If the test failed, send an error object.
+                error = NSError(
+                    domain: "org.swift.XCTestErrorDomain",
+                    code: 0,
+                    userInfo: [:])
+            }
+            completionHandler(error)
+        }
     }
 }
--- a/Sources/XCTest/XCTestExpectation.swift
+++ b/Sources/XCTest/XCTestExpectation.swift
@@ -1,0 +1,71 @@
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2016 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//
+//  XCTestExpectation.swift
+//  Expectations represent specific conditions in asynchronous testing.
+//
+
+/// Expectations represent specific conditions in asynchronous testing.
+public class XCTestExpectation {
+    internal let description: String
+    internal let file: StaticString
+    internal let line: UInt
+
+    internal var fulfilled = false
+    internal weak var testCase: XCTestCase?
+
+    internal init(description: String, file: StaticString, line: UInt, testCase: XCTestCase) {
+        self.description = description
+        self.file = file
+        self.line = line
+        self.testCase = testCase
+    }
+
+    /// Marks an expectation as having been met. It's an error to call this
+    /// method on an expectation that has already been fulfilled, or when the
+    /// test case that vended the expectation has already completed.
+    ///
+    /// - Parameter file: The file name to use in the error message if
+    ///   expectations are not met before the given timeout. Default is the file
+    ///   containing the call to this method. It is rare to provide this
+    ///   parameter when calling this method.
+    /// - Parameter line: The line number to use in the error message if the
+    ///   expectations are not met before the given timeout. Default is the line
+    ///   number of the call to this method in the calling file. It is rare to
+    ///   provide this parameter when calling this method.
+    ///
+    /// - Note: Whereas Objective-C XCTest determines the file and line
+    ///   number the expectation was fulfilled using symbolication, this
+    ///   implementation opts to take `file` and `line` as parameters instead.
+    ///   As a result, the interface to these methods are not exactly identical
+    ///   between these environments. To ensure compatibility of tests between
+    ///   swift-corelibs-xctest and Apple XCTest, it is not recommended to pass
+    ///   explicit values for `file` and `line`.
+    public func fulfill(file: StaticString = #file, line: UInt = #line) {
+        // FIXME: Objective-C XCTest emits failures when expectations are
+        //        fulfilled after the test cases that generated those
+        //        expectations have completed. Similarly, this should cause an
+        //        error as well.
+        if fulfilled {
+            // Mirror Objective-C XCTest behavior: treat multiple calls to
+            // fulfill() as an unexpected failure.
+            let failure = XCTFailure(
+                message: "multiple calls made to XCTestExpectation.fulfill() for \(description).",
+                failureDescription: "API violation",
+                expected: false,
+                file: file,
+                line: line)
+            if let failureHandler = XCTFailureHandler {
+                failureHandler(failure)
+            }
+        } else {
+            fulfilled = true
+        }
+    }
+}
--- a/Sources/XCTest/XCWaitCompletionHandler.swift
+++ b/Sources/XCTest/XCWaitCompletionHandler.swift
@@ -1,0 +1,27 @@
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2015 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//
+//  XCWaitCompletionHandler.swift
+//  A closure invoked by XCTestCase when a wait for expectations to be
+//  fulfilled times out.
+//
+
+#if os(Linux) || os(FreeBSD)
+    import Foundation
+#else
+    import SwiftFoundation
+#endif
+
+/// A block to be invoked when a call to wait times out or has had all
+/// associated expectations fulfilled.
+///
+/// - Parameter error: If the wait timed out or a failure was raised while
+///   waiting, the error's code will specify the type of failure. Otherwise
+///   error will be nil.
+public typealias XCWaitCompletionHandler = (NSError?) -> ()
--- a/Tests/Functional/Asynchronous/Expectations/main.swift
+++ b/Tests/Functional/Asynchronous/Expectations/main.swift
@@ -1,0 +1,94 @@
+// RUN: %{swiftc} %s -o %{built_tests_dir}/Asynchronous
+// RUN: %{built_tests_dir}/Asynchronous > %t || true
+// RUN: %{xctest_checker} %t %s
+
+#if os(Linux) || os(FreeBSD)
+    import XCTest
+    import Foundation
+#else
+    import SwiftXCTest
+    import SwiftFoundation
+#endif
+
+class ExpectationsTestCase: XCTestCase {
+// CHECK: Test Case 'ExpectationsTestCase.test_waitingForAnUnfulfilledExpectation_fails' started.
+// CHECK: .*/Tests/Functional/Asynchronous/Expectations/main.swift:19: error: ExpectationsTestCase.test_waitingForAnUnfulfilledExpectation_fails : Asynchronous wait failed - Exceeded timeout of 0.2 seconds, with unfulfilled expectations: foo
+// CHECK: Test Case 'ExpectationsTestCase.test_waitingForAnUnfulfilledExpectation_fails' failed \(\d+\.\d+ seconds\).
+    func test_waitingForAnUnfulfilledExpectation_fails() {
+        expectationWithDescription("foo")
+        waitForExpectationsWithTimeout(0.2, handler: nil)
+    }
+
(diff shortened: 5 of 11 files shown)


modocache:
> Without using something like `CoreSymbolication.framework`, we cannot emit failures for the correct file or line while maintaining an identical API.

Symbolication is a great solution for getting the file and line numbers in Objective-C, where default parameters aren't available. However in very large codebases, I've seen symbolication take over 90 seconds to complete. I believe the private API `+[XCTestCase _enableSymbolication]` was added to allow users to opt-out of symbolication--but then failure reporting doesn't work very well.

In a Swift-only API, however, I think `file` and `line` parameters are better. They work well in any sized codebase, and like `XCTAssert`, users will probably never actually use the parameters.

jeffh:
Capturing the last (or all) locations of un-awaited expectations seems like a good alternative solution without having to desymbolicate. :+1:

Several observations on this implementation vs Objective-C XCTest (for which Nimble's is a similar to):

1. This implementation has a leeway of `max(min(1/10th of the timeout interval, 60 * NSEC_PER_SEC), NSEC_PER_MSEC)`. The Objective-C XCTest sets this to `0` since it uses timer sources directly over `dispatch_after`. Along with using the global high priority queue, this decreases the likelihood of the timer not triggering at the desired timeout because of (dispatch) run loop congestion.
2. Objective-C XCTest doesn't actually poll for fulfillment of expectations. `fulfill()` triggers "poll-check" - that is the machinery to potentially stop the timer if all expectations are met.
3. `fulfill()` multiple times for the same expectation is an API Violation in Objective-C XCTest.
4. There's no attempted detection and recovery / abort when the main thread is blocked. Nimble attempts to continue unsafely by forcing the runloop to stop (but failing that test). Objective-C XCTest restarts the testing process. I'm not sure what's the recommended strategy here.

I think `NSRunLoop` is used to block instead of `dispatch_semaphore_wait` because it can still process traditional run loop operations, but I'm not familiar to how closely tied Objective-C CoreFoundation's `CFRunLoop` sources and libdispatch sources/queues are. So that's just an educated guess.

Just my 2 cents.

modocache:
Thanks for the comments, @jeffh! :clap:

> Capturing the last (or all) locations of un-awaited expectations seems like a good alternative solution without having to desymbolicate.

This is actually what Objective-C XCTest does as well! Running the following will report a single failure on the line where the last expectation is created:

```swift
func test_expectationsUnawaited() {
    let _ = self.expectationWithDescription("foo")
    let _ = self.expectationWithDescription("bar") // Failure reported here in Objective-C XCTest
}
```

> 3) `fulfill()` multiple times for the same expectation is an API Violation in Objective-C XCTest.

Nice catch--I'll update the code in this RFC. :+1: 

Thanks for bringing up (1), (2), and (4)--they'll be important to keep in mind once we've decided which dependencies to take on, and are ready to focus on implementation.

briancroom:
I agree with @jeffh that using the `file`/`line` params seems a very appropriate (and Swifty) way to capture the callsite info needed for failure reporting.

Pulling in Corelibs Foundation for `NSRunLoop` and `NSError` seems preferable to me as well for driving the asynchronous flows necessary for this. @phausler do you have any concerns about Foundation being testing by a library which itself uses Foundation?

parkera:
We have that circular dependency for testing Foundation with XCTest on Darwin as well, and it's never bothered me. I figure if we break `NSRunLoop` or `NSError`, then the tests are still likely to fail. It may be harder for us to figure out why, but that's the cost of being a low level library.

parkera:
@mike-ferris-apple should still comment on the actual substance of this RFC though.

phausler:
@briancroom I am of the opinion that testing something with itself only makes sense when the project is very mature. NSRunLoop is very new to swift-corelibs-foundation and I'm certain that we have at least one or two bugs still lurking around there. If we can help it I'd vote in favor of avoiding a circular graph if at all possible.

Perhaps a conditional inclusion for upper level frameworks/apps that depend on Foundation might be interesting.

But as @parkera mentioned @mike-ferris-apple should probably be the point of inquiry on that.

mike-ferris-apple:
Just a quick reply saying that I see this and will take a look when I get a chance. May be a couple days at this point, though.

mike-ferris-apple:
On point (A) in the original pull request, I think this seems like a reasonable solution. Writing source compatible tests is still very much possible, and mostly this will be transparent to folks.

Point (B) says that there's a dummy implementation, from what I am seeing it looks like the current code would simply not compile on Linux? @briancroom commented that using Foundation to get access to NSError and NSRunLoop might be preferable and I would tend to agree. But I also agree with @phausler  that establishing a dependency of Foundation (and in particular on NSRunLoop) is worrying given the current state of maturity of the corelibs Foundation.

I have my doubts about how practically useful this API will be as long as the wait call is always guaranteed to wait the full timeout. I know from experience that when we have had bugs in Xcode's implementation that yielded the same behavior it winds up making tests run unacceptably slow.

modocache:
> Point (B) says that there's a dummy implementation, from what I am seeing it looks like the current code would simply not compile on Linux?

Yup! To be clear: the current implementation only works on OS X, and is meant purely as a placeholder. My plan was to add a cross-platform implementation (that also satisfied the expectations before the timeout when possible), just as soon as we reached a consensus on whether to depend upon swift-corelibs-libdispatch or swift-corelibs-foundation.

It sounds like while there are worries about the maturity of swift-corelibs-foundation, no one is outright opposed to the idea of swift-corelibs-xctest depending upon swift-corelibs-foundation. I agree with this sentiment.

As a next step, I'd like to amend this pull request with an implementation that depends upon swift-corelibs-foundation. Integrating swift-corelibs-foundation into this project and the build script may end up being a good chunk of work, so if anyone's 100% opposed to this idea, let me know ASAP, and save me a weekend of free time! :wink:

mike-ferris-apple:
Let's separate the question slightly...

@parkera and @phausler, I'd love to here your thoughts.

1) Would there be any objection to generally establishing a dependency on Foundation? For example, what if it was for access to NSError?

2) is the NSRunLoop question specifically...  This is the probably more interesting one in terms of the newness of that code.

@modocache I would suggest that if we're OK with #1 in general, then this is worth pursuing, and that perhaps we can structure the waiting implementation internally to possibly allow for alternate implementations to be tried out... I am not terribly sure what a non-libdispach/non-Foundation implementation of this would look like, but we could at least keep that door open.

parkera:
No objection from me on using Foundation (with the caveat that we know a big renaming effort is under way so you may have to update soon).

The run loop implementation in corelibs Foundation is actually pretty close to mature. It's 95% the same as our other platforms and the Linux-specific stuff looks very reasonable to me.

Sources/XCTest/XCTestCase.swift
```diff
@@ -98,4 +129,137 @@ extension XCTestCase {
...
+            }
+
+            // If we've met all expectations, then break out of the specified
+            // timeout loop early.
+            if unfulfilledDescriptions.count == 0 {
+                break
+            }
+
+            // Otherwise, wait another fraction of a second.
+            runLoop.runUntilDate(NSDate(timeIntervalSinceNow: 0.01))
```
modocache: @parkera @phausler Here I've implemented asynchronous tests in XCTest using SwiftFoundation's `NSRunLoop`. Unfortunately, this doesn't exactly work--unlike when I use Objective-C Foundation, the SwiftFoundation runloop here does not run until the specified date.

In order to get this to work as it does with Objective-C Foundation, I had to make the following change to SwiftFoundation:

```diff
diff --git a/Foundation/NSRunLoop.swift b/Foundation/NSRunLoop.swift
index 4434429..6288242 100644
--- a/Foundation/NSRunLoop.swift
+++ b/Foundation/NSRunLoop.swift
@@ -95,10 +95,6 @@ extension NSRunLoop {
             return false
         }
         let modeArg = mode._cfObject
-        if _CFRunLoopFinished(_cfRunLoop, modeArg) {
-            return false
-        }
-
         let limitTime = limitDate.timeIntervalSinceReferenceDate
         let ti = limitTime - CFAbsoluteTimeGetCurrent()
         CFRunLoopRunInMode(modeArg, ti, true)
```

I examined the implementation of `_CFRunLoopFinished` to see what was amiss, but couldn't figure it out. Any ideas?

modocache:
I updated the commit to use SwiftFoundation. The project builds on OS X, and its tests pass if I make the change to `NSRunLoop` I describe in a comment above.

I use [the same technique swift-corelibs-foundation does to specify a dependency](https://github.com/apple/swift-corelibs-foundation/blob/bfa81efa1c55c79f6674a36433dd509933280168/Docs/GettingStarted.md#on-os-x). swift-corelibs-xctest now assumes `../swift-corelibs-foundation/Foundation.xcodeproj` exists. I'll add a note to the README to make sure contributors know what's necessary to build the project now.

For Linux, I plan on massaging the build script and `lit.cfg` sometime soon in order to get things working.

Sources/XCTest/XCTestCase.swift
modocache: Looks like the above works perfectly fine on Linux. Perhaps swift-corelibs-xctest should depend upon Objective-C Foundation when being built for OS X, instead of SwiftFoundation?

modocache:
OK! I pushed up a few changes to `build_script.py` and the tests' `lit.cfg`. These changes are dependent upon #46, so I'd love to see that PR merged soon.

This now works perfectly on Linux. It works fine on OS X, aside from that weird `NSRunLoop` behavior -- @parkera, @phausler, any help here would be appreciated. Should we depend on Objective-C Foundation on OS X, instead of SwiftFoundation?

@mike-ferris-apple, you'll be happy to hear that `XCTestCase.waitForExpectationsWithTimeout()` only waits as long as the expectations are unfulfilled, or when the timeout is reached. So tests for expectations that are fulfilled immediately pass immediately.

There are some changes to how this project is built: see my changes to the README for details. These changes will cause `swift/utils/build-script --xctest` to break. I'll send a corresponding pull request to apple/swift before we merge these changes.

> fulfill() multiple times for the same expectation is an API Violation in Objective-C XCTest.

@jeffh, I still haven't added addressed this comment, but I will! :stuck_out_tongue_closed_eyes:

modocache:
By the way, @briancroom: an interesting aspect of the new dependency on Foundation is that we can use `NSTimeInterval` for time measurements instead of `Double`. :hourglass: :+1:

modocache:
Here's how the build script will need to change once these changes are merged: https://github.com/modocache/swift/commit/a17c05b630b472a410a2b3a525c9878998c84ce1. That commit allows me to build and test swift-corelibs-xctest from OS X and Linux.

parkera:
@modocache I think it's fine in this case to use the Swift version of Foundation as a dependency on OS X. My thinking is this: the reason that swift-corelibs-foundation and swift-corelibs-xctest run on OS X at all is as a convenience for us to develop them in an environment that most of us are already familiar with (OS X, Xcode, etc). The deployment target of these two libraries is currently just Linux though (you'll notice that they are not present in the packages created for OS X).

Plus, swift-corelibs-foundation already uses swift-corelibs-xctest as its test harness on OS X instead of the system one. So we have a kind of precedent there.

modocache:
> Plus, swift-corelibs-foundation already uses swift-corelibs-xctest as its test harness on OS X instead of the system one.

It does?! Wow, this is news to me. I had assumed that, like swift-package-manager, it conditionally used either Objective-C XCTest or swift-corelibs-xctest, based on the host platform. It's very cool that it uses swift-corelibs-xctest on all platforms! :100: 

>  I think it's fine in this case to use the Swift version of Foundation as a dependency on OS X.

Awesome, great to hear. Any ideas on the behavior discrepancy around `NSRunLoop` I describe in https://github.com/apple/swift-corelibs-xctest/pull/43#discussion_r52113249?

parkera:
If you can, let's open a PR on swift-corelibs-foundation to discuss the required change for NSRunLoop.

modocache:
@parkera Sure thing: https://github.com/apple/swift-corelibs-foundation/pull/258 Thanks!

modocache:
Boy, I'm not sure why, but using the latest Swift snapshot this branch triggers an assertion in the Swift compiler on OS X: https://gist.github.com/modocache/2b4f9f6252010e55af1e -- `Assertion failed: ((!FileContext || issue != Status::Malformed) && "error deserializing an individual record"), function error, file /Users/buildnode/jenkins/workspace/oss-swift-package-osx/swift/include/swift/Serialization/ModuleFile.h, line 358."`

Looking into it, but if anyone has any ideas I'd love to hear them.

mike-ferris-apple:
Looking at the latest version. In general this looks good. I must admit I feel not quite up to speed on the issue discussed above about the functioning of NSRunLoop on OS X in the core libs Foundation. Has that been sufficiently resolved that this pull request can be merged?

A couple specific comments:

- In the "Note:"s for the two APIs, it would probably be worth explicitly saying something like "To ensure compatibility of tests between core libs XCTest and Xcode's XCTest, do not pass explicit values for file: and line:."
- Is there an issue in the wait implementation if the timeout is hit before the first iteration of the while loop? It looks like unfulfilledDescriptions will never get generated in that case?
- Probably more than we should look to resolve in the initial pull, but it would be good to rework the runloop stuff to avoid polling. Some kind of runloop source that can be signaled when an expectation is fulfilled to wake things up would be a better long term approach. But I'd say let's get the basics nailed down and in before embarking on that as it will pull on more of the Foundation functionality and who knows what that will turn up.

modocache:
Dang, still running into a compiler crash, even on the latest toolchain: https://gist.github.com/modocache/ad562405d19aa4ad6194. I was really hoping an update would fix all my problems! :stuck_out_tongue_closed_eyes: 

I'll keep at it. If anyone has any idea what that build log means, feedback would be much appreciated!!

drewcrawford:
Having delved into the sourcecode for the serializer recently, the compile crash is hitting some kind of mystery problem while importing a module.  

Quite possibly, "stuff happened" recently in the serializer and it can no longer read modules it wrote from an earlier version.  So it may be as simple as very carefully cleaning out old build kludge.

Alternately the serializer may be emitting some artifact that it itself can't read, in that case we need to isolate which one and why.

Stuff to try:

* Deleting DerivedData
* Deleting ModuleCache
* Cleaning out any other intermediate build files you can find on your system (may want to re-clone everything from github)
* Trying to work out which module won't import, by stubbing out each import statement and replacing the compile errors with dummy functions.  This will at least isolate which import is the villain.

modocache:
@drewcrawford Thanks for the suggestions! Very much appreciated.

I've tried deleting DerivedData entirely, but haven't gone as far as deleting and re-cloning everything. It might be worth a try!

Once @shahmishal enables OS X testing for this project, we may be able to see the crash occur (or not occur) on a CI system.

drewcrawford:
Does this error only occur on OSX?

Regardless, I can run this on my private buildfarm and report behavior, if I know the reproduction steps.  I would need to know affected platform(s), and how you are building (such as which preset for build-script, or if you are using a custom preset, its definition, or whether you build some other way, etc.)

That would at least isolate whether it's an "old build artifact" kind of issue.

modocache:
@drewcrawford Yes, only on OS X. To reproduce, using an Xcode configured to use the `swift-DEVELOPMENT-SNAPSHOT-2016-02-08` or later:

1. Clone both swift-corelibs-foundation and swift-corelibs-xctest into the same directory. (If your machine can build Swift from source, you should already have this directory structure.)
2. `xcodebuild -workspace swift-corelibs-xctest/XCTest.xcworkspace -scheme SwiftXCTest`

On Linux, you may confirm that this works as is:

1. Build Swift and swift-corelibs-foundation, using `swift/utils/build-script -R --foundation`.
2. `swift-corelibs-xctest/build_script.py --test --swiftc /path/to/swift-linux-x86_64/bin/swiftc --foundation-build-dir /path/to/foundation-linux-x86_64/Foundation`

modocache:
@drewcrawford I rebased this branch onto master, resolving the merge conflicts. The problem persists: Linux tests pass, OS X hits an assertion. Would love another set of eyes!

czechboy0:
@modocache I'm seeing the same thing, using the last dev snapshot of swift (02/25). Crashes with the same details that you have in your gist. Might be worth a bug report to the compiler ninjas? (Sorry if you've already done that, I'm being Mr. Obvious today)

drewcrawford:
My build farm confirms what you are seeing. 

Warning, deep serializer magic:  adding the value `$BUILT_PRODUCTS_DIR/usr/local/include/CoreFoundation` to `User Header Search Paths` for XCTest target resolves for me.  Perhaps it also resolves for you?

If so, I think I know what is going on here.  Sadly.  But that may work for now.

parkera:
I think we saw something similar while bringing up corelibs-foundation. There is some limitation right now that everyone who wishes to use a module must also include all of its C header dependencies.

drewcrawford:
I believe this is probably related to [SR-655](https://bugs.swift.org/browse/SR-655?) if you're looking for a bug to vote on.

modocache:
@drewcrawford Your advice was absolutely spot-on. Adding CoreFoundation headers to the include path solved all my problems on OS X.

It also seems the discrepancy in `NSRunLoop` behavior on OS X has been fixed since I first submitted this pull request. The tests pass on both OS X and Linux now! Thanks @parkera and @phausler! :heart_eyes: 

I'll address the following comments:

- [ ] In the "Note:"s for the two APIs, it would probably be worth explicitly saying something like "To ensure compatibility of tests between core libs XCTest and Xcode's XCTest, do not pass explicit values for file: and line:."
- [ ] Is there an issue in the wait implementation if the timeout is hit before the first iteration of the while loop? It looks like unfulfilledDescriptions will never get generated in that case?
- [ ] `fulfill()` multiple times for the same expectation is an API Violation in Objective-C XCTest.

In order to land this without breaking CI, I've submitted #62. That pull request, plus changes to the Swift build script, should be merged before merging this pull request.

modocache:
@mxcl I believe XCTest's new dependency on Foundation will necessitate some changes to the Swift package manager, but I'm not sure how much will have to change. The SwiftPM tests fail with this pull request:

```
swiftpm/Utilities/bootstrap \
    --swiftc=/home/modocache/GitHub/apple/build/Ninja-ReleaseAssert/swift-linux-x86_64/bin/swiftc \
    --sbt=/home/modocache/GitHub/apple/build/Ninja-ReleaseAssert/llbuild-linux-x86_64/bin/swift-build-tool \
    --build=/home/modocache/GitHub/apple/build/Ninja-ReleaseAssert/swiftpm-linux-x86_64 \
    --xctest=/home/modocache/GitHub/apple/build/Ninja-ReleaseAssert/xctest-linux-x86_64 \
    test

# ...

Compiling Swift Module 'PackageTypetest' (1 sources)
/home/modocache/GitHub/apple/swiftpm/Tests/PackageType/PackageNameTests.swift:12:8: error: missing required module 'Foundation'
import XCTest
       ^
```

I'll investigate further in the coming days, but just thought I'd ping you as a heads up.

modocache:
@mike-ferris-apple This is ready for final review! In order to keep CI passing, here's the release checklist:

- [ ] 1. Review and merge #62, #63, and https://github.com/apple/swift-package-manager/pull/176 (in any order). These contain future-proof changes to the XCTest and SwiftPM build scripts.
- [ ] 2. Review and merge https://github.com/apple/swift/pull/1556. This passes correct values for the parameters introduced in (1).
- [ ] 3. Review and merge https://github.com/apple/swift-package-manager/pull/177. This prepares SwiftPM to build XCTest with its new Foundation dependency.
- [ ] 4. Review and merge this pull request.

README.md
```diff
@@ -45,16 +48,34 @@ If your install of Swift is located at `/swift` and you wish to install XCTest i
...
+You may build XCTest via the "SwiftXCTest" scheme in `XCTest.xcworkspace`. The workspace assumes that Foundation and XCTest are checked out from GitHub in sibling directories. For example:
+
 ```
+% cd Development
+% ls
+swift-corelibs-foundation swift-corelibs-xctest
+%
+```
+
+Unlike on Linux, you do not need to build Foundation prior to building XCTest. The "SwiftXCTest" Xcode scheme takes care of that for you.
```
briancroom: :clap:

Sources/XCTest/XCTestCase.swift
```diff
@@ -48,6 +54,14 @@ private func test<T: XCTestCase>(testFunc: T -> () throws -> Void) -> XCTestCase
     }
 }
 
+// FIXME: These should be instance variables defined on XCTestCase, but when so
+//        defined Linux tests fail with "hidden symbol isn't defined". Use
+//        globals for the time being, as these seem to appease the Linux
+//        compiler.
+private var XCTLatestExpectationLocation: (file: StaticString, line: UInt) = ("", 0)
```
briancroom: I'm not sure, but I'm wondering if it may be less error-prone to not have a separate global var for this but rather to include the file+line values on `XCTestExpectation`? It doesn't feel right that this is non-optional but is set both here and below to what is effectively a placeholder value for when there is no "latest expectation".

Sources/XCTest/XCTestCase.swift
```diff
@@ -48,6 +54,14 @@ private func test<T: XCTestCase>(testFunc: T -> () throws -> Void) -> XCTestCase
...
+private var XCTAllExpectations = [XCTestExpectation]()
+private var XCTAllExpectationFailures = [XCTFailure]()
```
briancroom: I'm curious why you decided to gather expectation-related failures separately instead of calling into `XCTFailureHandler` directly when they arise?

Sources/XCTest/XCTestCase.swift
```diff
@@ -93,6 +107,24 @@ extension XCTestCase {
 
                 testCase.tearDown()
 
+                // It is an API violation to create expectations but not wait
```
briancroom: This entire block feels like something that could be extracted into a private helper. The `invokeTests` function is getting a bit unwieldy!

Sources/XCTest/XCTestCase.swift
```diff
@@ -122,4 +154,155 @@ extension XCTestCase {
...
+        let failure = XCTFailure(
+            message: "multiple calls made to XCTestExpectation.fulfill() for \(expectation.description).",
+            failureDescription: "API violation",
+            expected: false,
+            file: file,
+            line: line)
+        XCTAllExpectationFailures.append(failure)
+    }
+
+    /// Creates and returns an expectation associated with the test case.
```
briancroom: Nice documentation on these!

Sources/XCTest/XCTestCase.swift
```diff
@@ -122,4 +154,155 @@ extension XCTestCase {
...
+    ///   as a test failure.
+    ///
+    /// - Note: Whereas Objective-C XCTest determines the file and line
+    ///   number of the "wait" call using symbolication, this implementation
+    ///   opts to take `file` and `line` as parameters instead. As a result,
+    ///   the interface to these methods are not exactly identical between
+    ///   these environments. To ensure compatibility of tests between
+    ///   swift-corelibs-xctest and Apple XCTest, it is not recommended to pass
+    ///   explicit values for `file` and `line`.
+    public func waitForExpectationsWithTimeout(timeout: Double, file: StaticString = #file, line: UInt = #line, handler: XCWaitCompletionHandler?) {
```
briancroom: `NSTimeInterval`?

Sources/XCTest/XCTestCase.swift
modocache: Good call. On multiple occasions I came close to giving XCTestExpectation file and line attributes. I'll try consolidating on that approach and seeing how much cleaner this can get.

Sources/XCTest/XCTestExpectation.swift
```diff
@@ -0,0 +1,59 @@
...
+    /// - Note: Whereas Objective-C XCTest determines the file and line
+    ///   number the expectation was fulfilled using symbolication, this
+    ///   implementation opts to take `file` and `line` as parameters instead.
+    ///   As a result, the interface to these methods are not exactly identical
+    ///   between these environments. To ensure compatibility of tests between
+    ///   swift-corelibs-xctest and Apple XCTest, it is not recommended to pass
+    ///   explicit values for `file` and `line`.
+    public func fulfill(file: StaticString = #file, line: UInt = #line) {
+        if fulfilled {
+            if testCase == nil {
```
briancroom: Maybe invert and use `if let testCase = testCase` here?

Tests/Functional/Asynchronous/main.swift
```diff
@@ -0,0 +1,184 @@
...
+        ]
+    }
+
+    func test_whenExpectationsAreNotFulfilled_handlerCalled_andFails() {
+        self.expectationWithDescription("fog")
+
+        var handlerWasCalled = false
+        self.waitForExpectationsWithTimeout(0.2) { error in
+            XCTAssertNotNil(error, "Expectation handlers for unfulfilled expectations should not be nil.")
+            XCTAssertTrue(error!.domain.hasSuffix("XCTestErrorDomain"), "The last component of the error domain should match Objective-C XCTest.")
```
briancroom: Aside: it makes me sad that we have to force-unwrap here. It would be nice if `XCTAssertTrue` (and other assertions) would accept optional values and fail on nil (à la [Nimble](https://github.com/Quick/Nimble/blob/master/Tests/Nimble/Matchers/BeLogicalTest.swift#L36))

Tests/Functional/Asynchronous/main.swift
```diff
@@ -0,0 +1,184 @@
+// RUN: %{swiftc} %s -o %{built_tests_dir}/Asynchronous
```
briancroom: Just a few thoughts on this file. I don't think anything needs to be changed now for this PR, but I've been thinking a lot about these things especially in light of the discussion on https://github.com/apple/swift-corelibs-xctest/pull/61 and I want to make sure we keep iterating towards the most effective testing strategy we can manage to produce.

Overall I really like these tests! I like the way you've named the individual test cases and I like how straightforward the test method bodies are throughout. Nonetheless, when I first started reviewing this file I found it pretty opaque and difficult to navigate. I think this is because the expectations are mostly encoded in the `CHECK` lines at the top of the file, relatively far way from the code actually being tested. Additionally, the expected failures/passes are buried in a screenful of text with no syntax highlighting (at least in Xcode), so the signal-to-noise ratio is pretty poor when trying to parse the expectations.

This test file also presents several barriers to adding new tests in the future, should that prove necessary. Besides adding the new test method body and/or `XCTestCase` subclass, it is also necessary to figure out precisely where the new `CHECK` lines fit into the existing output. Once that is done, other numbers like the total number of tests executed and the line numbers of existing failure messages may need to be adjusted! :fearful: 

Some of this could be improved by reducing the specificity of the `CHECK` contents (e.g. wildcards for the total number of tests being run, wildcard line numbers for failures where we aren't specifically testing the file/line output, wildcard for a greater portion of the file path) but I don't know if there's any reasonable way to improve the proximity of the tests and their expected outcomes without a more radical change in approach.

README.md
```diff
@@ -24,13 +24,15 @@ The rest of this document will focus on how this version of XCTest differs from
...
+### On Linux
+
 XCTest can be built as part of the overall Swift package. When following [the instructions for building Swift](http://www.github.com/apple/swift), pass the `--xctest` option to the build script:
 
 ```sh
 swift/utils/build-script --xctest
 ```
 
-If you want to build just XCTest, use the `build_script.py` script at the root of the project. The `master` version of XCTest must be built with the `master` version of Swift.
+If you want to build just XCTest, use the `build_script.py` script at the root of the project. The `master` version of XCTest must be built with the `master` version of Swift. XCTest has a dependency upon Foundation, so you must have built the `master` version of that as well.
```
briancroom: I'm wondering how true it is that a `master` build of Foundation is necessary? I just hacked around a little bit and saw a successful build and test run using the Foundation+CoreFoundation shipped with the latest snapshot. Clearly there may be times where we depend on code only present on Foundation `master` but I would imagine that won't be too frequent.

Do you think it would be reasonable to treat `--foundation-build-dir` as an optional parameter to the build script, allowing the compiler to find a pre-installed Foundation library from a snapshot? This would simplify the use case where someone just wants to verify that their changes don't break `XCTest`'s Linux test suite without doing too much extra legwork getting a more complete swift.org development environment set up. The same logic applies to the `--swiftc` argument too, I think.

It'd be pretty neat if people could spin up an environment from a Docker image like [this](https://hub.docker.com/r/swiftdocker/swift/), clone just `swift-corelibs-xctest` and run `./build_script.py --test` without having to wait for `swift` and `llvm` to build which (AFAIK?) is currently required when building swift-corelibs-foundation from source.

briancroom:
Thanks for all the persistent work on this @modocache! It was really cool to see it all working on both platforms when I checked it out today. I've gone through and left a bunch of comments, some nits, some larger discussion points, but none of which I would consider to be merge blockers on this.

Tests/Functional/Asynchronous/main.swift
briancroom: Well I've just learned that the `// CHECK:` lines can be located anywhere in the file rather than just being one big block at the top. It seems like just moving them around a bit could go a long way towards making these tests more approachable.

mike-ferris-apple:
Lots of activity here, trying to make sense of it all.

@modocache , on your checklist above, 1, 2 and 3 look good to me. Since there's been a bit of discussion about what parameters should be required in the pm build script after this change, it would probably be a good idea to get buy in from @mxcl or @ddunbar for that PR.

@briancroom had a bunch of interested comments/suggestions above. I agree with him that most changes from his suggestions could be addressed as later refinements if you want (though see below for one possible exception).

On the meat of the PR, I made a couple comments on individual parts of the commit. Overall this looks good! Thanks for your persistence in putting this all together! The one thing that seems possibly worth addressing is the business about keeping the expectation failures out of band with the regular failure stream.

Brian, once any final tweaks are made and once the pm guys sign off, would you like to do the actual merging here since it will need several steps to be staged? If testing is needed in the middle of the sequence, ping me and I can kick it off.

Tests/Functional/Asynchronous/main.swift
modocache: @briancroom Thank you for the excellent feedback. You're absolutely right, and as I responded to feedback on this pull request I definitely found the tests extremely hard to update.

- First off, your suggestion to place the `// CHECK:` lines close to what they are testing is spot-on. I've done so, and the tests are much easier to read.
- Unfortunately, the line needs to begin with `// CHECK:`, meaning it can't be indented. I'll fix this in a separate pull request, since I think it's nice to have identical indentation level.
- I also split up the `Asynchronous/main.swift` test into three separate files one for each test case. This reduces the reasons for line numbers to change--those are annoying to update.

It'd also be really nice to "name" line numbers. Here's an example:

```swift
// By storing the line number in a named variable, even if the line
// upon which the failure is reported in the source file changes, the
// test won't have to be updated.
func testBar() {
    // CHECK: .*/Tests/Functional/Foo/main.swift:[[LINE_NUMBER]]: error: FooCase.testBar : .*
    XCTAssert(false) // LINE_NUMBER: #line
}
```

We could also do the following:

```swift
// The check line references the line number directly.
XCTAssert(false) // CHECK: .*/Tests/Functional/Foo/main.swift:[[#line]]: error: FooCase.testBar : .*
```

Your suggestions gave me several ideas on how we can make the functional tests more maintainable--thanks!! :heart_eyes:

README.md
modocache: I specified `master` in the README as a matter of convenience--using the Swift CI we can guarantee that XCTest and Foundation `master` build together. Of course, it might be possible for people to contribute by linking the latest XCTest `master` against an older Foundation snapshot, but it'll be hard for us to keep track of which versions of XCTest works with which versions of Foundation.

I totally agree with your suggestions to make it easier to contribute (and I think it's imperative we keep the barrier to entry as low as possible), but if possible I'd like to leave those as future improvements.

Sources/XCTest/XCTestCase.swift
modocache: Very good point! I've done away with this in favor of `XCTFailureHandler`.

Sources/XCTest/XCTestCase.swift
modocache: Excellent, yes!! Once this lands we should replace all `XCTest.TimeInterval` with `Foundation.NSTimeInterval` -- I won't do so in this pull request, it's big enough as it is! :wink:

modocache:
Thanks for the excellent reviews, @briancroom and @mike-ferris-apple! :100: I believe I've addressed all the comments.

@mike-ferris-apple, as you suggested, I'll wait for @mxcl or @ddunbar to review https://github.com/apple/swift-package-manager/pull/176 and https://github.com/apple/swift-package-manager/pull/177. If those look good, I'll merge everything [in the order I specified](https://github.com/apple/swift-corelibs-xctest/pull/43#issuecomment-192813377).

As for CI: @mike-ferris-apple, could you "@swift-ci please test" the following three pull requests?

- https://github.com/apple/swift-corelibs-xctest/pull/62
- https://github.com/apple/swift-corelibs-xctest/pull/63
- https://github.com/apple/swift-package-manager/pull/176

modocache:
@mike-ferris-apple Everything's in place for this to be merged. Could you ask @swift-ci to please test? If the tests go well on Linux I'll merge. (I assume the CI still doesn't work for OS X, I'll test that locally prior to merging.)

mike-ferris-apple:
@swift-ci please test

modocache:
I'll double check this works on OS X, then merge this within the next few hours.

modocache:
We did it!! :tada:

Thanks for all the awesome reviews and the amazing troubleshooting everyone. I'll send an email to the corelibs-dev mailing list to let people know about the new Foundation dependency.

briancroom:
Yeaaaah! So exciting to see this go in.

mike-ferris-apple:
Great! Thanks for your perseverance on this one!

PR #44, opened 2016-01-25
flush stdout after any printed test result

phausler:
When running tests from wrappers, such as continuous integration servers, the output of tests may be incorrectly buffered that will result in truncated displaying of tests when failures occur. This behavior incorrectly identifies the wrong test as the failure point in the suites, enforcing a fflush of stdout ensures the buffer is correctly flushed after each marker point for testing.

### DIFF
--- a/Sources/XCTest/XCTestCase.swift
+++ b/Sources/XCTest/XCTestCase.swift
@@ -46,7 +46,7 @@ extension XCTestCase {
                     }
                 }
 
-                print("Test Case '\(method)' started.")
+                XCTPrint("Test Case '\(method)' started.")
 
                 setUp()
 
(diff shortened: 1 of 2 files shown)


modocache:
How about an `internal

phausler:
Fair idea: updated to funnel to a single method wrapping print and fflush (kept the parameters from print just in case)

briancroom:
I wonder if inserting a call to `setlinebuf` or `setvbuf` inside of `XCTMain` would be simpler than manually calling `fflush`?

phausler:
as it stands the originating issue for this was to verify a CI failure from Foundation where we were getting a timeout in a test that was not able to get the full output; specifically this was behaving differently than the Mac OS X version. Flushing at each print not only ensures proper output regardless of what may have been potentially mucked with before it. Furthermore it could be reasonable for a test to print multiple lines in the assert, flushing at the print will end up making it flush the entire contents instead of partial contents that `setlinebuf` may do.

mike-ferris-apple:
This seems reasonable, especially when done within a single funnel point like this.

PR #45, opened 2016-02-04
[XCTestMain] Don't XCTPrint arrays of strings

modocache:
`XCTPrint()` takes a variadic argument of items to print, mirroring the API of `print()` itself. Unfortunately, passing this variadic argument directly to `print()` causes an array object to be printed. For example:

```swift
func XCTPrint(items: Any...) {
    print(items)
}

XCTPrint("Hello!") // Prints '["Hello!"]\n'
```

As a result of this behavior, swift-corelibs-xctest currently prints out test output such as the following:

```
["Test Case \'ErrorHandling.test_shouldButDoesNotThrowErrorInAssertion\' failed (0.0 seconds)."]
["Test Case \'ErrorHandling.test_shouldThrowErrorInAssertion\' started."]
```

This is different from the expected output in `Tests/Functional`, which is causing the test suite to fail.

Instead of mirroring the `print()` API, have `XCTPrint()` simply take a string object to print. This has the added benefit of being simpler--we can add the variadic arguments and the line separator parameter back if we ever need them. The test suite passes with this change.

---

On a related note:

1. Would it be possible to have Apple CI cover this repository as well? I'd love to be able to ask @swift-ci to "please test"!
2. I'm still trying to find time to add these tests to the Swift validation-test suite. I'd love it if someone beat me to it! :wink:

PR #46, opened 2016-02-04
[build_script] Use built XCTest for Linux tests

modocache:
Previously, tests on Linux could only be run using the installed swift-corelibs-xctest. Modify the script to use the built XCTest instead of the installed one.

In addition, since the script no longer relies on the installed XCTest, move the `--test` logic such that it can be used even if `--module-install-path` and `--library-install-path` are not specified.

---

Now, running tests on Linux is as easy as:

```
$ ./build_script.py \
    --swiftc ~/build/Ninja-ReleaseAssert/swift-linux-x86_64/bin/swiftc \
    --build-dir ~/build/Ninja-ReleaseAssert/xctest-linux-x86_64 \
    --swift-build-dir ~/build/Ninja-ReleaseAssert/swift-linux-x86_64 \
    --test
```

Keep in mind that the tests won't pass until after #45 is merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,16 +38,13 @@ If your install of Swift is located at `/swift` and you wish to install XCTest i
 ./build_script.py --swiftc="/swift/usr/bin/swiftc" --build-dir="/tmp/XCTest_build" --swift-build-dir="/swift/usr" --library-install-path="/swift/usr/lib/swift/linux" --module-install-path="/swift/usr/lib/swift/linux/x86_64"
 ```
 
-To run the tests on Linux, pass the `--test` option in combination with options to
-install XCTest in your active version of Swift:
+To run the tests on Linux, use the `--test` option:
 
 ```sh
 ./build_script.py \
     --swiftc="/swift/usr/bin/swiftc" \
     --build-dir="/tmp/XCTest_build" \
     --swift-build-dir="/swift/usr" \
-    --library-install-path="/swift/usr/lib/swift/linux" \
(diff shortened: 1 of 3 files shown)


briancroom:
I haven't had a chance to give it a try yet, but this seems like an important step towards making this test suite robust enough to catch the sorts of things we want it to. 👍

briancroom:
The README should also be updated with the new build script invocation here!

README.md
```diff
@@ -38,16 +38,13 @@ If your install of Swift is located at `/swift` and you wish to install XCTest i
...
-install XCTest in your active version of Swift:
+To run the tests on Linux, use the `--test` option:
 
 ```sh
 ./build_script.py \
     --swiftc="/swift/usr/bin/swiftc" \
     --build-dir="/tmp/XCTest_build" \
     --swift-build-dir="/swift/usr" \
-    --library-install-path="/swift/usr/lib/swift/linux" \
-    --module-install-path="/swift/usr/lib/swift/linux/x86_64" \
```
modocache: > The README should also be updated with the new build script invocation here!

@briancroom Is this what you meant?

briancroom: Yup!

modocache: Awesome. Thanks!

modocache:
@mike-ferris-apple Friendly ping! This is one of those rare pull requests with no tradeoffs--only benefits to be had here! :sunglasses:

modocache:
Thanks!!

PR #48, opened 2016-02-07
[XCTAssert] s/__FILE__/#file/g, s/__LINE__/#line/g

modocache:
"Screaming snake case" debugging identifiers like `__FILE__` and `__LINE__` are scheduled to be removed as per evolution proposal [SE-0028](https://github.com/apple/swift-evolution/blob/6e040b5d5d19df60329deffc125552a4cd63a21d/proposals/0028-modernizing-debug-identifiers.md). They were marked as deprecated in https://github.com/apple/swift/commit/0619e57a61f27f721e273ab6f808ac81011aeb2c#diff-b1b9777d77a07ee6e082cce7a2c52608R87.

Replace all uses of `__FILE__` with `#file`, and `__LINE__` with `#line`. This fixes compilation warnings emitted when building swift-corelibs-xctest with the latest version of Swift.

:warning: Keep in mind that this means people using swift-corelibs-xctest must build the latest version of Swift from source (which includes the commit referenced above) in order to maintain compatibility. To run the tests on OS X, users must launch Xcode pointing to a toolchain built from the latest source.

### DIFF
--- a/Sources/XCTest/XCTAssert.swift
+++ b/Sources/XCTest/XCTAssert.swift
@@ -81,7 +81,7 @@ private enum _XCTAssertionResult {
     }
 }
 
-private func _XCTEvaluateAssertion(assertion: _XCTAssertion, @autoclosure message: () -> String = "", file: StaticString = __FILE__, line: UInt = __LINE__, @noescape expression: () throws -> _XCTAssertionResult) {
+private func _XCTEvaluateAssertion(assertion: _XCTAssertion, @autoclosure message: () -> String = "", file: StaticString = #file, line: UInt = #line, @noescape expression: () throws -> _XCTAssertionResult) {
     let result: _XCTAssertionResult
     do {
         result = try expression()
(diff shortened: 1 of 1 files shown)


mike-ferris-apple:
Let me ask around here a bit about staging of the Swift change... We'll certainly take this eventually, but it may be that waiting a bit will be a bit more convenient.

modocache:
Sounds reasonable. Waiting for the next snapshot, which will include the new `#file` and `#line`, might be the best course of action.

modocache:
@mike-ferris-apple This can probably go in now that https://github.com/apple/swift/tree/swift-DEVELOPMENT-SNAPSHOT-2016-02-08 has been tagged. Users who encounter build errors should be encouraged to migrate to the latest snapshot.

briancroom:
@mike-ferris-apple @modocache agreed - this should be merged pretty soon now, as it needs to be included in the next snapshot, or else XCTest will fail to build.

parkera:
I'll go ahead and merge this.

modocache:
Great, thanks!

PR #50, opened 2016-02-18
Fixes to enable swift section objects

hpux735:
This change enables xctest to continue to function once the swift sections changes are applied.  This pull should not be applied prior to that change, and should be pulled at that time.

https://github.com/apple/swift/pull/1157

@gribozavr @tienex

### DIFF
--- a/build_script.py
+++ b/build_script.py
@@ -81,11 +81,17 @@ def main():
                         action="store_true",
                         dest="test",
                         default=False)
+    parser.add_argument("--arch",
+                        help="target architecture",
+                        action="store",
+                        dest="arch",
+                        default=None)
     args = parser.parse_args()
 
     swiftc = os.path.abspath(args.swiftc)
     build_dir = os.path.abspath(args.build_dir)
     swift_build_dir = os.path.abspath(args.swift_build_dir)
+    arch = args.arch
 
     if not os.path.exists(build_dir):
         run("mkdir -p {}".format(build_dir))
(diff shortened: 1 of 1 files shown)


build_script.py
```diff
@@ -112,7 +118,7 @@ def main():
     run("{0} -c {1} -emit-object {2} -module-name XCTest -parse-as-library -emit-module "
         "-emit-module-path {3}/XCTest.swiftmodule -o {3}/XCTest.o -force-single-frontend-invocation "
         "-module-link-name XCTest".format(swiftc, style_options, " ".join(sourcePaths), build_dir))
-    run("clang {0}/XCTest.o -shared -o {0}/libXCTest.so -Wl,--no-undefined -Wl,-soname,libXCTest.so -L{1}/lib/swift/linux/ -lswiftGlibc -lswiftCore -lm".format(build_dir, swift_build_dir))
+    run("clang {1}/lib/swift/linux/{2}/swift_begin.o {0}/XCTest.o {1}/lib/swift/linux/{2}/swift_end.o -shared -o {0}/libXCTest.so -Wl,--no-undefined -Wl,-soname,libXCTest.so -L{1}/lib/swift/linux/ -lswiftGlibc -lswiftCore -lm".format(build_dir, swift_build_dir, arch))
```
modocache: Fantastic. Thanks, @hpux735! :100: 

Personally I've been meaning to reformat this convoluted string interpolation to use identifiers--something like the following:

```python
run('clang {swift_build_dir}/lib/swift/linux/{arch}/swift_begin.o '
    '{build_dir}/XCTest.o '
    '{swift_build_dir}/lib/swift/linux/{arch}/swift_end.o '
    '-shared -o {build_dir}/libXCTest.so '
    '-Wl,--no-undefined -Wl,-soname,libXCTest.so '
    '-L{swift_build_dir}/lib/swift/linux/ '
    '-lswiftGlibc -lswiftCore -lm'.format(
        build_dir=build_dir,
        swift_build_dir=swift_build_dir,
        arch=arch))
```

But I can take care of that at a later time. I think this is fine for now!

build_script.py
```diff
@@ -81,11 +81,17 @@ def main():
                         action="store_true",
                         dest="test",
                         default=False)
+    parser.add_argument("--arch",
+                        help="target architecture",
+                        action="store",
+                        dest="arch",
+                        default=None)
```
modocache: You don't need to explicitly mark `default=None` here. `argparse` already uses `None` in the absence of the option being specified.

modocache: It looks like the build invocation below uses `arch`. Shouldn't `--arch` be marked as `required=True` here?

build_script.py
```diff
@@ -81,11 +81,17 @@ def main():
...
     args = parser.parse_args()
 
     swiftc = os.path.abspath(args.swiftc)
     build_dir = os.path.abspath(args.build_dir)
     swift_build_dir = os.path.abspath(args.swift_build_dir)
+    arch = args.arch
```
modocache: nit-pick: You could simply reference `args.arch` below, and eliminate this line. Other parameters, such as `args.build_style`, follow that pattern.

gribozavr:
@hpux735 @modocache Sorry, I need to merge this PR to keep builds correct.  Could you submit fixes for @modocache's comments as a separate PR?

modocache:
Makes sense, thanks @gribozavr! I've addressed my comments in #51. Thanks again, @hpux735! :+1:

PR #51, opened 2016-02-18
[build_script.py] Remove redundant argparse params

modocache:
- For options `like-this`, `argparse` stores the variable name `like_this`. Remove redundant `dest` parameters in the build script.
- For options marked as "store", `argparse` stores `None` when the option is not specified. Remove redundant `default=None` parameters in the build script.
- For options marked as "store_true", the default value is `False`. Remove redundant `default=True` parameters in the build script.
- The default action for options is "store". Remove redundant `action="store"` parameters in the build script.
- `argparse` raw formatters only really make sense when used with multi-line strings that have custom indentation. It's not necessary in the build script.
- `--arch` option is required.

### DIFF
--- a/build_script.py
+++ b/build_script.py
@@ -24,41 +24,31 @@ def run(command):
     subprocess.check_call(command, shell=True)
 
 def main():
-    parser = argparse.ArgumentParser(formatter_class=argparse.RawDescriptionHelpFormatter,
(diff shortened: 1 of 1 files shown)


mike-ferris-apple:
Looks like a good simplification. Thanks!

PR #54, opened 2016-02-26
[build_script.py] Deprecate obsolete parameters

modocache:
"--swift-build-dir" and "--arch" are no longer used by the build script. We could remove them outright, but that would break CI. Instead, mark them as deprecated first. We will remove them once apple/swift's build script has been updated to no longer pass "--swift-build-dir" and "--arch".

The build script also contains some instructions on how to build, but these are duplicated in the README. Remove them to save us the trouble of updating two sets of instructions.

---

Let's give the CI a spin! Could someone please ask @swift-ci to please test? :grin:

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,8 +38,6 @@ If your install of Swift is located at `/swift` and you wish to install XCTest i
 ./build_script.py \
     --swiftc="/swift/usr/bin/swiftc" \
     --build-dir="/tmp/XCTest_build" \
-    --swift-build-dir="/swift/usr" \
-    --arch="x86_64" \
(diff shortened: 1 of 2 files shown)


ddunbar:
@swift-ci please test

modocache:
Wow, it worked!! Awesome. Thanks for setting this up, @shahmishal! One additional request, though: would it be possible to enable OS X builds for swift-corelibs-xctest as well? It would help with #43, for example--that pull request would like pass Linux CI but not OS X.

shahmishal:
@modocache Yes, I will setup OS X bot for swift-corelibs-xctest. Thanks!

PR #56, opened 2016-02-27
[build_script.py] Glob source files to build

modocache:
Maintaining a list of source files to build in the build_script requires a change to that file every time a new source file is added. Instead, automatically pick up any new source files with a .swift extension.

---

One potential downside here is that we may end up including a `.swift` file in the `Sources/XCTest` which we *don't* want to compile, and in that case this would no longer work. I can't really imagine why we'd want such a thing, though.

### DIFF
--- a/build_script.py
+++ b/build_script.py
@@ -10,6 +10,7 @@
 #
 
 import argparse
+import glob
 import os
 import subprocess
 import tempfile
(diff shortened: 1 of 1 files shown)


modocache:
This makes good on a comment I made here: https://github.com/apple/swift-corelibs-xctest/pull/40/files#r52390643

mike-ferris-apple:
@swift-ci please test

PR #57, opened 2016-02-28
[1/2][build_script.py] Add "test" subcommand

modocache:

This "FIXME" describes a problem with the swift-corelibs-xctest build script when invoked in the context of the greater Swift build script: https://github.com/apple/swift/blob/6ccf5da2dfc8a74c84b2fd99c32dcdf2255699d7/utils/build-script-impl#L2070-L2072

The problem is that there is currently no way to test an *already built* XCTest.so. Instead, the XCTest build script re-builds the entire project, then tests the newly build XCTest.so.

Add a "test" subcommand to build_script.py, which allows users to test an already built XCTest.so:

```
$ ./build_script.py test --swiftc `which swiftc` /path/to/build/dir
```

Users may still choose to build *and* test, by using the old build_script.py invocation:

```
$ ./build_script.py --swiftc `which swiftc` --test
```

The Swift build script is sort of disingenuous--maybe even dangerous--at the moment. When a user runs `swift/utils/build-script --xctest --test`, I think they'd expect that XCTest is built, then tested. In reality, it is built, then built again, then the second version is tested.

This gets even worse when we're talking about *installing* XCTest. In that case, it is:

1. Built.
2. Built again.
3. The version that was just built is tested.
4. It is built again.
5. The final version to be built (which hasn't really been tested) is installed.

This pull request is a large change just to prevent step (2) above, but it puts in place the subcommand mechanism that we'll need to fix the install step (4).

## What downsides are there to merging this pull request?

1. It's a large change that touches most of the current build script.
2. The subcommand parsing is admittedly wacky. I tried commenting why things were the way they were--for example, why we had to jump through hoops to get a "default subcommand".

Still, I think this is a significant improvement!

### DIFF
--- a/build_script.py
+++ b/build_script.py
@@ -13,6 +13,7 @@ import argparse
 import glob
 import os
 import subprocess
+import sys
 import tempfile
 
 SOURCE_DIR = os.path.dirname(os.path.abspath(__file__))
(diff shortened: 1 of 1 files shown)


modocache:
@swift-ci please test

parkera:
@swift-ci please test

mike-ferris-apple:
This seems like a good thing to have.

PR #58, opened 2016-02-28
[2/2][build_script.py] Add "install" subcommand

modocache:
This builds upon the subcommand system added in #57, to add an install command. That pull request should be reviewed and merged prior to this one.

As that pull request describes, because there is no way to test or install XCTest without also building the project, the current Swift build script ends up building XCTest three times. This adds an "install" subcommand that allows an already built XCTest to be installed.

### DIFF
--- a/build_script.py
+++ b/build_script.py
@@ -26,6 +26,14 @@ def run(command):
     subprocess.check_call(command, shell=True)
 
 
+def _mkdirp(path):
+    """
+    Creates a directory at the given path if it doesn't already exist.
+    """
+    if not os.path.exists(path):
+        run("mkdir -p {}".format(path))
+
+
 def _build(args):
     """
     Build XCTest and place the built products in the given 'build_dir'.
@@ -33,9 +41,7 @@ def _build(args):
     """
     swiftc = os.path.abspath(args.swiftc)
     build_dir = os.path.abspath(args.build_dir)
-
-    if not os.path.exists(build_dir):
-        run("mkdir -p {}".format(build_dir))
+    _mkdirp(build_dir)
 
     sourcePaths = glob.glob(os.path.join(
         SOURCE_DIR, 'Sources', 'XCTest', '*.swift'))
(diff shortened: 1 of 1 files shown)


modocache:
@swift-ci please test

modocache:
@mike-ferris-apple Could you ask @swift-ci to please test? Once this is merged, I can change the Swift build script to only build XCTest as many times as necessary.

mike-ferris-apple:
@swift-ci please test

briancroom:
It looks like the CI hook here isn't going through, since the last build on this job is from [Feb. 29](https://ci.swift.org/view/Pull%20Request/job/swift-corelibs-xctest-PR-Linux/5/). Perhaps @shahmishal can help debug?

@modocache I quite like the flexibility you're introducing with this series of build script improvements. :+1:

modocache:
Thanks, @briancroom! But, yeah, I'm worried the CI must have become sentient, and is disobeying direct orders!! :rotating_light: 

Seriously, though, I think there's some work being done to enable the CI for external core members. Maybe it's a bit unresponsive in the interim, even for @mike-ferris-apple...

build_script.py
```diff
@@ -26,6 +26,14 @@ def run(command):
     subprocess.check_call(command, shell=True)
 
 
+def _mkdirp(path):
```
briancroom: I'm curious about this `_` and the ones on `_build` etc. Is this a Python convention I should be familiar with for internal (/non-exported?) functions? If that's the case, how is this function different from e.g. `note` in its role in the module?

modocache: I do think preceding with an underscore is typically done for "internal" functions, but I may be wrong. `note` was tangential to this change so I left it alone. @practicalswift, what do you think?

modocache: It is admittedly kind of silly, since this is just a single script, so really *every* function is "internal" :stuck_out_tongue:

shahmishal:
@swift-ci Please test

shahmishal:
@swift-ci Please test OS X platform

modocache:
Amazing! Thanks for the OS X support, @shahmishal! It looks like the build is failing because `xcodebuild` isn't using the latest Swift toolchain--it can't handle the new `#file` and `#line` identifiers. Is this some sort of configuration issue?

This pull request only affects Linux, and the tests are passing on that platform, so I'll go ahead and merge this.

PR #64, opened 2016-03-06
Allow selecting a particular test or test case to run from the command line

briancroom:
This adds the ability for test binaries using `XCTMain` to receive a command-line argument that specifies that only an individual test or test case should be run, as implied in [this proposal](https://github.com/apple/swift-evolution/blob/master/proposals/0019-package-manager-testing.md):

```sh
./FooTests FooTestCase/testFoo  # Run a single test method
./FooTests FooTestCase          # Run all the tests in FooTestCase
```

 See also https://github.com/apple/swift-package-manager/pull/168 which begins implementing this in `swift test`.

### DIFF
--- a/Sources/XCTest/XCTestMain.swift
+++ b/Sources/XCTest/XCTestMain.swift
@@ -67,11 +67,18 @@ internal struct XCTRun {
 ///
 ///     XCTMain([ testCase(TestFoo.allTests) ])
 ///
+/// Command line arguments can be used to select a particular test or test case to execute. For example:
+///
+///     ./FooTests FooTestCase/testFoo  # Run a single test method
+///     ./FooTests FooTestCase          # Run all the tests in FooTestCase
+///
 /// - Parameter testCases: An array of test cases run, each produced by a call to the `testCase` function
 /// - seealso: `testCase`
 @noreturn public func XCTMain(testCases: [XCTestCaseEntry]) {
+    let filter = TestFiltering()
+
     let overallDuration = measureTimeExecutingBlock {
-        for (testCase, tests) in testCases {
+        for (testCase, tests) in TestFiltering.filterTests(testCases, filter: filter.selectedTestFilter) {
             testCase.invokeTests(tests)
         }
     }
--- a/Tests/Functional/SelectedTest/main.swift
+++ b/Tests/Functional/SelectedTest/main.swift
@@ -1,0 +1,60 @@
+// RUN: %{swiftc} %s -o %{built_tests_dir}/SelectedTest
+// RUN: %{built_tests_dir}/SelectedTest SelectedTest.ExecutedTestCase/test_foo > %T/one_test_method || true
+// RUN: %{built_tests_dir}/SelectedTest SelectedTest.ExecutedTestCase > %T/one_test_case || true
+// RUN: %{built_tests_dir}/SelectedTest > %T/all || true
+// RUN: %{xctest_checker} -p "// CHECK-METHOD:   " %T/one_test_method %s
+// RUN: %{xctest_checker} -p "// CHECK-TESTCASE: " %T/one_test_case %s
+// RUN: %{xctest_checker} -p "// CHECK-ALL:      " %T/all %s
+
+#if os(Linux) || os(FreeBSD)
+    import XCTest
+#else
+    import SwiftXCTest
+#endif
+
+class ExecutedTestCase: XCTestCase {
+    static var allTests: [(String, ExecutedTestCase -> () throws -> Void)] {
+        return [
+            ("test_bar", test_bar),
+            ("test_foo", test_foo),
+        ]
+    }
+
+// CHECK-METHOD:   Test Case 'ExecutedTestCase.test_foo' started.
+// CHECK-METHOD:   Test Case 'ExecutedTestCase.test_foo' passed \(\d+\.\d+ seconds\).
+// CHECK-TESTCASE: Test Case 'ExecutedTestCase.test_bar' started.
+// CHECK-TESTCASE: Test Case 'ExecutedTestCase.test_bar' passed \(\d+\.\d+ seconds\).
+// CHECK-ALL:      Test Case 'ExecutedTestCase.test_bar' started.
+// CHECK-ALL:      Test Case 'ExecutedTestCase.test_bar' passed \(\d+\.\d+ seconds\).
+    func test_bar() {}
+
+// CHECK-TESTCASE: Test Case 'ExecutedTestCase.test_foo' started.
+// CHECK-TESTCASE: Test Case 'ExecutedTestCase.test_foo' passed \(\d+\.\d+ seconds\).
+// CHECK-ALL:      Test Case 'ExecutedTestCase.test_foo' started.
+// CHECK-ALL:      Test Case 'ExecutedTestCase.test_foo' passed \(\d+\.\d+ seconds\).
+    func test_foo() {}
+}
+// CHECK-METHOD:   Executed 1 test, with 0 failures \(0 unexpected\) in \d+\.\d+ \(\d+\.\d+\) seconds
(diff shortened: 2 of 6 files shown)


czechboy0:
:+1: I'm curious about two things
What if the passed-in TestCase/Test doesn't match an existing test? I can't find anything that would warn/error out, so users might accidentally filter out all tests (test nothing), but still get a "success" return code. Or is it ensured elsewhere that at least one test runs, otherwise the invocation fails?

briancroom:
@czechboy0 nope, you are correct that it would run 0 tests and exit with a `0` status code. I was copying the behavior of the `xctest` runner app on OS X because that's what `swift test` would call into there, although I agree that is potentially misleading and a different behavior may be desirable.

mike-ferris-apple:
Very cool. The test you added for this might be expanded to run a third time with no argument to ensure that the default behavior is intact, running all the tests.

briancroom:
Yeah I considered that, but figured that there were plenty of other funcional tests that thoroughly cover that case. Still, having the contrast present in the one test file could be nice.

briancroom:
Added it now :grinning:

Tests/Functional/SelectedTest/main.swift
```diff
@@ -0,0 +1,58 @@
...
+//// Output from the third invocation (running all tests)
+// CHECK: Test Case 'ExecutedTestCase.test_bar' started.
+// CHECK: Test Case 'ExecutedTestCase.test_bar' passed \(\d+\.\d+ seconds\).
+// CHECK: Test Case 'ExecutedTestCase.test_foo' started.
+// CHECK: Test Case 'ExecutedTestCase.test_foo' passed \(\d+\.\d+ seconds\).
+// CHECK: Executed 2 tests, with 0 failures \(0 unexpected\) in \d+\.\d+ \(\d+\.\d+\) seconds
+// CHECK: Test Case 'SkippedTestCase.test_baz' started.
+// CHECK: Test Case 'SkippedTestCase.test_baz' passed \(\d+\.\d+ seconds\).
+// CHECK: Executed 1 test, with 0 failures \(0 unexpected\) in \d+\.\d+ \(\d+\.\d+\) seconds
+// CHECK: Total executed 3 tests, with 0 failures \(0 unexpected\) in \d+\.\d+ \(\d+\.\d+\) seconds
```
modocache: I've been thoroughly convinced by your suggestions in #43, and I think you should place these assertions as close to the source code they're testing as possible (like this: https://github.com/modocache/swift-corelibs-xctest/commit/22f1d24a0a48dd278d133f18a0252ed91967b289#diff-38abae29a59be553f8b7fc357a30e6ecR14). In fact, I think we should modify all of our tests in this style!

briancroom: For this test in particular, I don't think it's going to work to reorganize the `CHECK` lines, unfortunately, because of the way that this one executes the tests repeatedly in order to compare the output when running them with different test filters active.

Or is there a neat way to structure this that I haven't thought of yet?

Tests/Functional/SelectedTest/main.swift
```diff
@@ -0,0 +1,58 @@
+// RUN: %{swiftc} %s -o %{built_tests_dir}/SelectedTest
+// RUN: %{built_tests_dir}/SelectedTest SelectedTest.ExecutedTestCase/test_foo > %t || true
+// RUN: %{built_tests_dir}/SelectedTest SelectedTest.ExecutedTestCase >> %t || true
+// RUN: %{built_tests_dir}/SelectedTest >> %t || true
+// RUN: %{xctest_checker} %t %s
+//// Output from the first invocation (running a single test method)
```
modocache: You should be able to handle these cases as `llvm-lit` would: by using a custom prefix. That is, run the checker once for each variation:

- `%{xctest_checker} --prefix "// CHECK-METHOD:" %T/one_test_method`
- `%{xctest_checker} --prefix "// CHECK-TESTCASE:" %T/one_test_case`
- `%{xctest_checker} --prefix "// CHECK-ALL:" %T/all`

(Here I'm assuming you've piped the output into files named `one_test_method`, etc., in a temporary directory `%T`. If our `lit.cfg` doesn't provide a temporary directory substitution we should add one!)

Sources/XCTest/XCTestMain.swift
```diff
@@ -67,11 +67,18 @@ internal struct XCTRun {
 ///
 ///     XCTMain([ testCase(TestFoo.allTests) ])
 ///
+/// Command line arguments can be used to select a particular test or test case to execute. For example:
+///
+///     ./FooTests FooTestCase/testFoo  # Run a single test method
+///     ./FooTests FooTestCase          # Run all the tests in FooTestCase
+///
```
modocache: Loving these documentation updates! :heart_eyes:

Tests/Functional/SelectedTest/main.swift
modocache: See my other comment: you may be able to use a distinct check prefix for each of the three runs.

modocache:
Wow, you make it look easy! Brilliant.

briancroom:
Updated and squashed!

The reordered assertions worked out really well @modocache, thanks for pointing out this feature of the test infrastructure!

briancroom:
This is ready to go as far as I am concerned. I'm not sure if there's any point in asking CI to do anything with it right now?

modocache:
I'm under the impression that CI is mostly broken right now due to the Swift 3 migration, yeah. @mike-ferris-apple, could you ask @swift-ci to please test, just to see what happens?

Also, @mike-ferris-apple, how would you feel if I tested this using the CI presets locally, then merged? Just like the good old days before @swift-ci was enabled for this repository! :joy:

modocache:
CI issues aside, I agree that this is ready to be merged anytime!

mike-ferris-apple:
@swift-ci please test

mike-ferris-apple:
I'd be OK with merging this hand-tested since the CI seems to be in a bit of a state right now.

modocache:
Great. I'll merge this and #68 sometime today.

modocache:
Sorry for the slow turnaround, @briancroom. I tried running the tests on Linux but encountered the following failure, which is probably due the Swift 3 migration:

```
/home/modocache/GitHub/apple/swift-corelibs-xctest/Sources/XCTest/TestFiltering.swift:63:54: error: 'split(_:maxSplit:allowEmptySlices:)' is unavailable: Please use split(separator:maxSplits:omittingEmptySubsequences:) instead
        let components = selectedTestName.characters.split("/").map(String.init)
                                                     ^~~~~
Swift.Collection:3:17: note: 'split(_:maxSplit:allowEmptySlices:)' has been explicitly marked unavailable here
    public func split(separator: Self.Iterator.Element, maxSplit: Int = default, allowEmptySlices: Bool = default) -> [Self.SubSequence]
                ^
```

The following change got all the tests passing for me:

```diff
diff --git a/Sources/XCTest/TestFiltering.swift b/Sources/XCTest/TestFiltering.swift
index f67d3a1..78e5191 100644
--- a/Sources/XCTest/TestFiltering.swift
+++ b/Sources/XCTest/TestFiltering.swift
@@ -60,7 +60,7 @@ private struct SelectedTest {
 
 private extension SelectedTest {
     init?(selectedTestName: String) {
-        let components = selectedTestName.characters.split("/").map(String.init)
+        let components = selectedTestName.characters.split(separator: "/").map(String.init)
         switch components.count {
         case 1:
             testCaseName = components[0]
```

On OS X, I needed to fix a build error in Foundation (https://github.com/apple/swift-corelibs-foundation/pull/284) in addition to applying the above change, but the tests all pass!

Ping me once you've made the above change and I'll merge--thanks!

briancroom:
I for one welcome our new Swift 3 overlords..and am regretting not investing the time in building an updated toolchain earlier.

Regardless, I've made the change now!

modocache:
Great! Thanks, @briancroom :100:

PR #65, opened 2016-03-07
Use a folder reference instead of groups for the functional tests

briancroom:
As suggested by @mike-ferris-apple https://github.com/modocache/swift-corelibs-xctest/commit/9466dc1990c89624e60913669a5ff225fb147213#commitcomment-16524676

This removes the need for people to explicitly add new functional tests to Xcode for them to be browsable, since Xcode doesn't care about them during its native build process anyway.

### DIFF
--- a/XCTest.xcodeproj/project.pbxproj
+++ b/XCTest.xcodeproj/project.pbxproj
@@ -24,10 +24,8 @@
 /* End PBXContainerItemProxy section */
 
 /* Begin PBXFileReference section */
-		342A33611C470D91001AA02A /* main.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = main.swift; sourceTree = "<group>"; };
-		342A33631C47383E001AA02A /* main.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = main.swift; sourceTree = "<group>"; };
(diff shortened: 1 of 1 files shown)


modocache:
Pulled it down to try it--looks good to me! One silly thing I'll miss: I used to be able to temporarily add the files to the corelibs-xctest target in order to get autocompletion and syntax highlighting from Xcode. :stuck_out_tongue_closed_eyes: 

@mike-ferris-apple could you ask @swift-ci to please test? :bow:

briancroom:
Hmm that's actually not entirely silly at all. I remember doing the same from time to time and appreciating letting the IDE help me out more.

Still, I think that if we are feeling the urge to do that frequently, it may be a sign that the functional tests are too complex. In any case, it's still *possible* to pop the file into the target, it just requires a few more clicks than before.

mike-ferris-apple:
@swift-ci please test

mike-ferris-apple:
@swift-ci Please test

modocache:
The OS X check is broken until `xcodebuild` uses the latest Swift toolchain. (Still, I'm hesitant to disable the OS X path in the Swift build script, since I believe this `xcodebuild` problem will be fixed in the CI infra.)

This seems good to merge, so I went ahead and merged. Thanks, @briancroom!

modocache:
Wow, one really nice thing about this change is that I can easily see the output of the functional tests, right in Xcode:

screen shot 2016-03-07 at 8 31 49 pm

That more than makes up for the lack of autocompletion. Thanks @briancroom! :bouquet:

briancroom:
Ohhh, nice! That's beautiful.

Thanks @mike-ferris-apple for pointing out this improvement.

PR #66, opened 2016-03-07
Reorganize test assertions

briancroom:
I've gone ahead and taken a first pass at rearranging the assertions in our existing functional tests, as discussed.

@modocache already has [more ideas](https://github.com/apple/swift-corelibs-xctest/pull/43#discussion_r55156054) of ways that this could potentially be made even better, but this seemed like nice low-hanging fruit until such time as new functionality is added to the `xctest_checker`.

I made a couple of other minor changes to the assertions here as well. In particular, I reduced the number of tests that explicitly check for correct line numbers in assertion failure, because the specific line numbers are prone to change as the project evolves, and it's wasteful to test the same behavior repeatedly. I also made a slight simplification to the file path in lines checking assertion failures. What previous was:

```
// CHECK: .*/Tests/Functional/SingleFailingTestCase/main.swift:22: error: SingleFailingTestCase.test_fails : XCTAssertTrue failed -
```

is now

```
// CHECK: .*/SingleFailingTestCase/main.swift:22: error: SingleFailingTestCase.test_fails : XCTAssertTrue failed -
```

I don't think that including the extra path components improved the quality of the assertion in any way, while it *does* increase the amount of noise, making it harder to determine what parts of the line are most relevant.

### DIFF
--- a/Tests/Functional/ErrorHandling/main.swift
+++ b/Tests/Functional/ErrorHandling/main.swift
@@ -1,24 +1,6 @@
 // RUN: %{swiftc} %s -o %{built_tests_dir}/ErrorHandling
 // RUN: %{built_tests_dir}/ErrorHandling > %t || true
 // RUN: %{xctest_checker} %t %s
-// CHECK: Test Case 'ErrorHandling.test_shouldButDoesNotThrowErrorInAssertion' started.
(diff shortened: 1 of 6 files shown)


mike-ferris-apple:
@swift-ci please test

modocache:
Awesome! Thanks for updating `__FUNCTION__`, too.

mike-ferris-apple:
@swift-ci Please test

modocache:
Again, the OS X check is broken until `xcodebuild` uses the latest Swift toolchain. (Still, I'm hesitant to disable the OS X path in the Swift build script, since I believe this `xcodebuild` problem will be fixed in the CI infra.)

PR #69, opened 2016-03-10
XCTestObservation and XCTestObservationCenter

briancroom:
As requested by @modocache, here is the start of some work on this missing feature for early review. I'm waiting for the next snapshot to rebase on top of the latest master because I don't have an environment with the new Swift 3 stdlib and corelibs-foundation stuff at this point.

For the `XCTestObservation` protocol, I've only included the methods related to `XCTestCase` at this point, not the bundle or test suite ones. If and when we decide to introduce `XCTestSuite` we can introduce the relevant events at that point.

Note that I haven't touched `XCTestRun` yet which could be added to provide a useful summary (failure count, test case success) in `testCaseDidFinish()`. I do remember that @mike-ferris-apple had [mentioned](https://github.com/apple/swift-corelibs-xctest/pull/40#issuecomment-171766079) avoiding it for some reason, and I was hoping to get an understanding of why that was before starting down that path. I did add `XCTestCase`'s `name` property though, which allows differentiating between instances in the observation callbacks.

Method documentation will also still be needed before this is mergeable! Looking for feedback first though :grinning:

### DIFF
--- a/Sources/XCTest/ObjectWrapper.swift
+++ b/Sources/XCTest/ObjectWrapper.swift
@@ -1,0 +1,25 @@
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2016 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//
+//  ObjectWrapper.swift
+//  Utility type for adapting implementors of a `class` protocol to Hashable
+//
+
+/// A `Hashable` representation of an object and its ObjectIdentifier. This is
+/// useful because Swift classes aren't implicitly hashable based on identity.
+internal struct ObjectWrapper<T>: Hashable {
+    let object: T
+    let objectIdentifier: ObjectIdentifier
+
+    var hashValue: Int { return objectIdentifier.hashValue }
+}
+
+internal func ==<T>(lhs: ObjectWrapper<T>, rhs: ObjectWrapper<T>) -> Bool {
+    return lhs.objectIdentifier == rhs.objectIdentifier
+}
--- a/Sources/XCTest/XCTestCase.swift
+++ b/Sources/XCTest/XCTestCase.swift
@@ -26,7 +26,14 @@ public typealias XCTestCaseEntry = (testCaseClass: XCTestCase.Type, allTests: [(
 
 public class XCTestCase {
 
+    /// The name of the test case, consisting of its class name and the method name it will run.
+    /// - Note: FIXME: This property should be readonly, but currently has to be publicly settable due to a
+    ///         toolchain bug on Linux. To ensure compatibility of tests between
+    ///         swift-corelibs-xctest and Apple XCTest, this property should not be modified.
+    public var name: String
+
     public required init() {
+        name = "\(self.dynamicType).<unknown>"
     }
 
     public func setUp() {
@@ -78,19 +85,19 @@ extension XCTestCase {
         let overallDuration = measureTimeExecutingBlock {
             for (name, test) in tests {
                 let testCase = self.init()
-                let fullName = "\(testCase.dynamicType).\(name)"
+                testCase.name = "\(testCase.dynamicType).\(name)"
 
                 var failures = [XCTFailure]()
                 XCTFailureHandler = { failure in
                     if !testCase.continueAfterFailure {
-                        failure.emit(fullName)
+                        failure.emit(testCase.name)
                         fatalError("Terminating execution due to test failure", file: failure.file, line: failure.line)
                     } else {
                         failures.append(failure)
                     }
                 }
 
-                XCTPrint("Test Case '\(fullName)' started.")
+                XCTPrint("Test Case '\(testCase.name)' started.")
 
                 testCase.setUp()
 
@@ -111,7 +118,7 @@ extension XCTestCase {
 
                 var result = "passed"
                 for failure in failures {
-                    failure.emit(fullName)
+                    failure.emit(testCase.name)
                     totalFailures += 1
                     if !failure.expected {
                         unexpectedFailures += 1
@@ -119,8 +126,8 @@ extension XCTestCase {
                     result = failures.count > 0 ? "failed" : "passed"
                 }
 
-                XCTPrint("Test Case '\(fullName)' \(result) (\(printableStringForTimeInterval(duration)) seconds).")
-                XCTAllRuns.append(XCTRun(duration: duration, method: fullName, passed: failures.count == 0, failures: failures))
+                XCTPrint("Test Case '\(testCase.name)' \(result) (\(printableStringForTimeInterval(duration)) seconds).")
+                XCTAllRuns.append(XCTRun(duration: duration, method: testCase.name, passed: failures.count == 0, failures: failures))
                 XCTFailureHandler = nil
             }
         }
--- a/Sources/XCTest/XCTestCase.swift
+++ b/Sources/XCTest/XCTestCase.swift
@@ -79,6 +79,8 @@ extension XCTestCase {
     }
 
     internal static func invokeTests(tests: [(String, XCTestCase throws -> Void)]) {
+        let observationCenter = XCTestObservationCenter.sharedTestObservationCenter()
+
         var totalDuration = 0.0
         var totalFailures = 0
         var unexpectedFailures = 0
@@ -89,6 +91,11 @@ extension XCTestCase {
 
                 var failures = [XCTFailure]()
                 XCTFailureHandler = { failure in
+                    observationCenter.testCase(testCase,
+                                               didFailWithDescription: failure.failureMessage,
+                                               inFile: String(failure.file),
+                                               atLine: failure.line)
+
                     if !testCase.continueAfterFailure {
                         failure.emit(testCase.name)
                         fatalError("Terminating execution due to test failure", file: failure.file, line: failure.line)
@@ -98,6 +105,8 @@ extension XCTestCase {
                 }
 
                 XCTPrint("Test Case '\(testCase.name)' started.")
+
+                observationCenter.testCaseWillStart(testCase)
 
                 testCase.setUp()
 
@@ -113,6 +122,8 @@ extension XCTestCase {
                 testCase.tearDown()
                 testCase.failIfExpectationsNotWaitedFor(XCTAllExpectations)
                 XCTAllExpectations = []
+
+                observationCenter.testCaseDidFinish(testCase)
 
                 totalDuration += duration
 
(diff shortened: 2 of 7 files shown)


Sources/XCTest/XCTestObservation.swift
```diff
@@ -0,0 +1,58 @@
...
+    func testCaseWillStart(testCase: XCTestCase)
+    func testCase(testCase: XCTestCase, didFailWithDescription description: String, inFile filePath: String?, atLine lineNumber: UInt)
+    func testCaseDidFinish(testCase: XCTestCase)
+}
+
+
+
+public class XCTestObservationCenter {
+    private static var center = XCTestObservationCenter()
+    private var observers = [XCTestObservation]()
```
czechboy0: Would it help to have `Set<XCTestObservation>` here instead, so that below you can avoid looking up the index of the observer to be removed?

briancroom: Unfortunately not, because `XCTestObservation` doesn't conform to `Hashable`. I briefly considered `NSMutableSet` but I'm not convinced that it's worth losing the strong typing.

czechboy0: Hmm since we know it's a class protocol, I'm surprised we can't force the set to use the pointer value as the hash value (basically defining `XCTestObservation`'s equality as `===`).

modocache:
Awesome, thanks @briancroom!! It's so exciting to see one of the final hurdles to API parity with Apple XCTest getting tackled. :+1: 

> If and when we decide to introduce XCTestSuite we can introduce the relevant events at that point.

This sounds like a reasonable approach. Also, thanks for linking to @mike-ferris-apple's comment about avoiding `XCTestRun`. That sounds strange to me, though--to write a cross-platform test observer, I'd expect XCTest users would want to write the exact same code--which means accessing the test results via `testCase.testRun`. I'll let him comment further, but I think this is something we should resolve sooner rather than later.

Sources/XCTest/XCTestCase.swift
```diff
@@ -92,6 +99,8 @@ extension XCTestCase {
 
                 XCTPrint("Test Case '\(fullName)' started.")
 
+                observationCenter.testCaseWillStart(testCase)
```
modocache: Here's a crazy idea: could we define an `XCTestObserver` that we use internally to call `XCTPrint()`? That is, could we use our own observation infrastructure to print the default test output?

I'm not sure off the top of my head whether Apple XCTest does this, but I think it'd be wild! :sunglasses:

Sources/XCTest/XCTestObservation.swift
briancroom: Yeah, that would make a lot of sense. AFAICT there's no way to do that without making `XCTestObservation` itself extend `Hashable` though.

Sources/XCTest/XCTestCase.swift
briancroom: As it turns out, Apple XCTest does exactly that (well, not *exactly*, because the logging uses the legacy `XCTestObserver` protocol, but close enough. I learned all this while working through https://github.com/Quick/Quick/pull/486)

I would like to make that refactor, yes. In fact, one reason why I wrote the observer in the functional test to keep track of state in properties instead of just `print`ing was to facilitate this, since `XCTestObservationCenter` doesn't provide any guarantees of the order in which observers are called.

mike-ferris-apple:
I don't recall exactly the thinking behind my comment about avoiding XCTestRun... It may have been motivated primarily by wanting to keep things simple at first. I think @modocache makes a good point that for some of what folks would be doing in an observer, they may need access and if that looks different, it's not portable.

briancroom:
I've rebased this and done some cleanups now:

* Switch to using `Set` to track observers by introducing a wrapper type (thanks @czechboy0 for getting me thinking about how I could make that work!)
* Add documentation to public symbols
* Add empty default implementations to `XCTestObservation` methods to line up with their `@optional` status in Objective-C

Note: Unfortunately I still haven't managed to get a functional Swift 3 dev environment up and running on my machine, so there could be build errors related to that.

In the interest of incremental development and getting more focused code review, I'm thinking it makes the most sense to defer `XCTestRun` and refactoring the test logging code to separate pull requests.

Sources/XCTest/ObjectWrapper.swift
```diff
@@ -0,0 +1,25 @@
...
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//
+//  ObjectWrapper.swift
+//  Utility type for adapting implementors of a `class` protocol to Hashable
+//
+
+/// A `Hashable` representation of an object and its ObjectIdentifier. This is
+/// useful because Swift classes aren't implicitly hashable based on identity.
```
czechboy0: Great way of putting it - I wonder whether there'd be value in opening a thread about changing that on Swift evolution. What do you think?

Sources/XCTest/XCTestObservationCenter.swift
```diff
@@ -0,0 +1,64 @@
...
+    internal func testCase(testCase: XCTestCase, didFailWithDescription description: String, inFile filePath: String?, atLine lineNumber: UInt) {
+        forEachObserver { $0.testCase(testCase, didFailWithDescription: description, inFile: filePath, atLine: lineNumber) }
+    }
+
+    internal func testCaseDidFinish(testCase: XCTestCase) {
+        forEachObserver { $0.testCaseDidFinish(testCase) }
+    }
+
+    private func forEachObserver(@noescape body: XCTestObservation -> Void) {
+        for observer in observers {
```
czechboy0: You could also follow the function's name :blush: 
`observers.forEach { body($0.object) }`

modocache:
>  I'm thinking it makes the most sense to defer `XCTestRun` and refactoring the test logging code to separate pull requests.

Makes sense to me! `XCTestRun` without `XCTestObservation` would add an API without any real consumers, after all.

Please remove the "[WIP]", rebase, and ping this thread once you're ready. I also emailed you a Swift 3 toolchain I built a few days ago--maybe that'll work...?

briancroom:
Done!

Thanks for the toolchain help @modocache, it worked great and allowed me to make a minor adjustment that's needed for this to build in the Swift 3 world.

Sources/XCTest/XCTestObservationCenter.swift
briancroom: Yeah that's true! I remember considering that but decided to do the traditional `for` loop instead. I seem to remember reading a comment from Chris Lattner at some point about `forEach` having a small performance overhead and `for` being perfectly idiomatic Swift, though either way would be perfectly valid here I think.

Sources/XCTest/ObjectWrapper.swift
briancroom: It does seem worth discussing. I was a bit disappointed that I had to jump through such hoops to make `Set` work out in this situation. I'm not sure how soon I'd have time to do a proper write-up though, so please feel free to open a thread if you feel so inclined!

Sources/XCTest/XCTestCase.swift
```diff
@@ -26,7 +26,11 @@ public typealias XCTestCaseEntry = (testCaseClass: XCTestCase.Type, allTests: [(
 
 public class XCTestCase {
 
-    public required init() {
+    /// The name of the test case, consisting of its class name and the method name it will run
+    public let name: String
+
+    public required init(methodName: String) {
```
modocache: Apple XCTest's `XCTestCase` defines two initializers: `-[XCTestCase initWithInvocation:]` and `-[XCTestCase testCaseWithSelector]`. I suppose neither of those could be made available in an environment without the Objective-C runtime. But I feel like we shouldn't provide `public` API beyond what Apple XCTest defines--that's the stuff evolution proposals are for. (I feel the empty initializer is fine because Apple XCTest also technically allows you to call `-[[XCTestCase alloc] init]`).

Is there a reason this API needs to be `public`? It seems we could have `methodName` be a `private` attribute that can be set after the `XCTestCase` instance is initialized. That would avoid the issue of providing an additional `public` API.

Sources/XCTest/XCTestCase.swift
```diff
@@ -72,25 +76,33 @@ extension XCTestCase {
...
-                let testCase = self.init()
-                let fullName = "\(testCase.dynamicType).\(name)"
+                let testCase = self.init(methodName: name)
 
                 var failures = [XCTFailure]()
                 XCTFailureHandler = { failure in
+                    observationCenter.testCase(testCase,
+                                               didFailWithDescription: "\(failure.failureDescription) - \(failure.message)",
+                                               inFile: String(failure.file),
+                                               atLine: failure.line)
```
modocache: The description we pass to the observers is different from the description that's printed by `XCTFailure.emit()`. Is this intentional? It seems like it opens us up to some maintenance dilemmas down the line... for example, @mike-ferris-apple [mentioned looking into removing trailing whitespace from failure messages](https://github.com/apple/swift-corelibs-xctest/pull/71#issuecomment-196318176). This code would necessitate us to modify two different call paths.

Could we have `XCTFailure.emit()` return a string, which we then pass to observers? Or is the description passed to observers supposed to be different here? I'll try checking to see what Apple XCTest does myself to make sure.

Of course, this could be something we take care of later. Let me know if you'd rather we address this in later commits.

modocache:
I had two comments. But boy, I love your tests for this feature! :100:

Sources/XCTest/XCTestCase.swift
briancroom: I've cleaned this up now :+1:

Sources/XCTest/XCTestCase.swift
briancroom: Thanks for pointing this out. When I first saw your comment I was pretty sure I had simply made that constructor `public` by mistake. Attempting to change it landed me in the middle of an intense battle with the compiler, which ended with me conceding we will have to live with `name` having a public setter for now (I left comments in the new code pointing this out.) Sadly non-public fields seem to be off-limits for now :disappointed:

I'm reasonably content with the latest version of this, though. What do you think?

(As an aside, while trying to understand the bigger picture here, I learned that it's effectively impossible to provide an overridden constructor on Swift `XCTestCase` subclasses under Apple XCTest right now, because the framework calls `init(invocation:)` when building the default test suite, which can't be overridden because of `NSInvocation`'s unavailability.)

Sources/XCTest/XCTestCase.swift
```diff
@@ -26,7 +26,14 @@ public typealias XCTestCaseEntry = (testCaseClass: XCTestCase.Type, allTests: [(
 
 public class XCTestCase {
 
+    /// The name of the test case, consisting of its class name and the method name it will run.
+    /// - Note: FIXME: This property should be readonly, but currently has to be publicly settable due to a
+    ///         toolchain bug on Linux. To ensure compatibility of tests between
+    ///         swift-corelibs-xctest and Apple XCTest, this property should not be modified.
```
modocache: Oh wow, interesting... It would be great to reference a JIRA issue here. :point_up:

modocache:
Awesome work! I'd be happy to merge this if @mike-ferris-apple could ask @swift-ci to please test.

parkera:
@swift-ci please test

modocache:
Fantastic! API parity with Apple XCTest is so close I can taste it. Thanks, @briancroom and @parkera! :+1:

PR #70, opened 2016-03-10
Add a README for the xctest_checker tool in the functional test suite

briancroom:
I was trying to make some minor changes to `xctest_checker` but didn't know how to get started with the tests, so I documented what I learned.

It's quite possible @modocache will come along and tell me my example invocation is all wrong. That'd be fine by me! :laughing:

### DIFF
--- a/Tests/Functional/xctest_checker/README.md
+++ b/Tests/Functional/xctest_checker/README.md
@@ -1,0 +1,13 @@
+# xctest_checker
+
+This tool is invoked by the swift-corelibs-xctest functional tests when they use the `{xctest_checker}` lit substitution. For more usage details, execute it from this directory using:
+
+```sh
+./xctest_checker.py -h
+```
+
+It includes unit tests which can be run with the following command in this directory:
+
+```sh
+python -m unittest discover
+```


Tests/Functional/xctest_checker/README.md
```diff
@@ -0,0 +1,13 @@
...
+This tool is invoked by the swift-corelibs-xctest functional tests when they use the `{xctest_checker}` lit substitution. For more usage details, execute it from this directory using:
+
+```sh
+./xctest_checker.py -h
+```
+
+It includes unit tests which can be run with the following command in this directory:
+
+```sh
+PYTHONPATH=. python tests/test_compare.py
```
modocache: There's a better way:

```
$ python -m unittest discover -s Tests/Functional/xctest_checker
```

This works with both relative and absolute paths.

modocache:
Looks great, just the one comment. Thanks, @briancroom!

Tests/Functional/xctest_checker/README.md
briancroom: Thanks for the tip!

I've updated the file, though I excluded the `-s` option because the context states that you'd be invoking the command from the correct directory already.

modocache:
Thanks! :wave:

PR #71, opened 2016-03-11
Allow differing leading/trailing whitespace when checking test output

briancroom:
There are two situations lately where I've been finding myself fighting the testing framework's current strictness with respect to whitespace at the beginning and end of lines:

1. [Here](https://github.com/apple/swift-corelibs-xctest/pull/64/files#diff-ed8e21c5d2f214bfdc63dde56bcf7930R47), I had to include some awkward whitespace after the `// CHECK-ALL:` prefix, in order to also be able to keep things nicely lined up on lines like [this](https://github.com/briancroom/swift-corelibs-xctest/blob/select-test/Tests/Functional/SelectedTest/main.swift#L27).
2. Whenever dealing with lines like [this one](https://github.com/apple/swift-corelibs-xctest/blob/master/Tests/Functional/SingleFailingTestCase/main.swift#L19), Xcode likes to trim the trailing space, which results in a test failure that is very difficult to understand because the difference between the lines is effectively invisible.

As I see it, we don't have any tests right now for which precise matching of leading/trailing whitespace is important, and I don't foresee any such tests coming up either, so it seems that this change will make the test checker a little easier to work with. On the other hand if there are concerns about this, we could also add a flag for turning this behavior off and on.

### DIFF
--- a/Tests/Functional/xctest_checker/xctest_checker/compare.py
+++ b/Tests/Functional/xctest_checker/xctest_checker/compare.py
@@ -28,8 +28,10 @@ def _expected_lines_and_line_numbers(path, check_prefix):
     with open(path) as f:
         for line_number, line in enumerate(f):
             if line.startswith(check_prefix):
-                yield line[len(check_prefix):], line_number+1
+                yield line[len(check_prefix):].strip(), line_number+1
 
+def _add_whitespace_leniency(original_regex):
+    return "^ *" + original_regex + " *$"
 
 def compare(actual, expected, check_prefix):
     """
@@ -52,7 +54,7 @@ def compare(actual, expected, check_prefix):
 
         (expected_line, expectation_source_line_number) = expected_line_and_line_number
 
-        if not re.match(expected_line, actual_line):
+        if not re.match(_add_whitespace_leniency(expected_line), actual_line):
             raise AssertionError('Actual line did not match the expected '
                                  'regular expression.\n'
                                  '{}:{}: Actual: {}\n'
--- a/Tests/Functional/ErrorHandling/main.swift
+++ b/Tests/Functional/ErrorHandling/main.swift
@@ -37,7 +37,7 @@ class ErrorHandling: XCTestCase {
     }
 
 // CHECK: Test Case 'ErrorHandling.test_shouldButDoesNotThrowErrorInAssertion' started.
-// CHECK: .*/ErrorHandling/main.swift:\d+: error: ErrorHandling.test_shouldButDoesNotThrowErrorInAssertion : XCTAssertThrowsError failed: did not throw error - 
+// CHECK: .*/ErrorHandling/main.swift:\d+: error: ErrorHandling.test_shouldButDoesNotThrowErrorInAssertion : XCTAssertThrowsError failed: did not throw error -
 // CHECK: Test Case 'ErrorHandling.test_shouldButDoesNotThrowErrorInAssertion' failed \(\d+\.\d+ seconds\).
     func test_shouldButDoesNotThrowErrorInAssertion() {
         XCTAssertThrowsError(try functionThatDoesNotThrowError())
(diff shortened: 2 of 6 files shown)


modocache:
I agree with your first point! Although I also *think* that `FileCheck` does the same thing, but I'm not sure.

My two cents on (2): personally I think the better solution is to put a `$` at the end of the line. I also think printing a superfluous space at the end of assertion failures is something we could think about addressing--and it's something I noticed due to the fact that we match trailing whitespace. I'll defer to @mike-ferris-apple on this point, though.

mike-ferris-apple:
If Xcode's XCTest is outputting extra space at the end of some lines, I wouldn't mind seeing that addressed. Does it? Any specifics as to when it does this would be appreciated, and I can file a radar for the behavior.

I think this particular change would be fine. Especially if it has precedent in FileCheck.

briancroom:
> Any specifics as to when it does this would be appreciated, and I can file a radar for the behavior.

The following test case:

```swift
import XCTest
class TestBar: XCTestCase {
    func testBar() {
        XCTAssert(false)
    }
}
```

produces the following (partial) output:

```
Test Case '-[XCTestSimpleTests.TestBar testBar]' started.
/Users/pivotal/workspace/Experiments/XCTestSimple/XCTestSimpleTests/XCTestSimpleTests.swift:15: error: -[XCTestSimpleTests.TestBar testBar] : XCTAssertTrue failed - 
Test Case '-[XCTestSimpleTests.TestBar testBar]' failed (0.005 seconds).
```

Note that the line indicating the failure *does* contain a trailing space after the hyphen. `swift-corelibs-xctest` also reproduces this behavior currently.

briancroom:
Regarding FileCheck's behavior, [this](http://llvm.org/docs/CommandGuide/FileCheck.html#cmdoption--match-full-lines) bit of documentation seems to confirm what @modocache indicated. (Note that our default matching style appears to be basically equivalent to FileCheck with the `--match-full-lines` flag)

briancroom:
Besides rebasing, I've also taken advantage of a comment on the aforementioned FileCheck docs and updated this behavior such that it now fully respects `^` or `$` characters if they are included in the check line, allowing explicit checking of leading/trailing whitespace if desired, but ignores such whitespace by default.

I've also updated a few tests to take advantage of the enhanced checker. Note that there are still a couple of tests like [this one](https://github.com/briancroom/swift-corelibs-xctest/blob/allow-leading-trailing-whitespace/Tests/Functional/FailingTestSuite/main.swift#L43) which explicit check for a trailing space. These will serve us well when we want to get rid of that trailing space (in a separate PR!)

Tests/Functional/xctest_checker/xctest_checker/compare.py
```diff
@@ -28,8 +28,10 @@ def _expected_lines_and_line_numbers(path, check_prefix):
     with open(path) as f:
         for line_number, line in enumerate(f):
             if line.startswith(check_prefix):
-                yield line[len(check_prefix):], line_number+1
+                yield line[len(check_prefix):].strip(), line_number+1
 
+def _add_whitespace_leniency(original_regex):
+    return "^ *" + original_regex + " *$"
```
modocache: Will the spaces here only match spaces, or will they match tabs as well? Perhaps use `\s`?

modocache:
Wow!! Can't argue with that. Sorry to make you jump through so many hoops. I think this is a great solution. Thanks, @briancroom! :100: 

Personally I'd prefer to keep a trailing `$` in a few places, until we file a radar for the Apple XCTest extra space and get that fixed. But certainly not all tests need them.

@mike-ferris-apple could you kick off the tests so we can merge this?

mike-ferris-apple:
@swift-ci please test

Tests/Functional/xctest_checker/xctest_checker/compare.py
briancroom: Yeah, that's a good point. `\s` is probably more appropriate here considering the way that I've been referencing `whitespace` generally in the docs for this PR.

OTOH, considering that it's not actually necessary at this moment, and CI has gone green already with the existing code (on Linux at least), how about we make that adjustment separately?

modocache: Works for me! :+1:

briancroom:
> Sorry to make you jump through so many hoops.

Not at all! This is exactly the sort of discussion I was hoping would come out of this PR. We all learned a thing or two, got some clarity on a (minor!) bug in `XCTest.framework`, and ended up with a much more robust patch than my initial naive attempt. Seems like a job well done all around. :bow:

PR #72, opened 2016-03-11
Allow Xcode to highlight failing functional test expectations

briancroom:
As a part of our continuing attempts to improve the quality of the functional test suite, I was hoping to improve the way that the test failure output integrates with Xcode.

Here, I have amended failure output to include the file name and line number where the `// CHECK:` line came from. Xcode is able to use this information to mark the line in the editor as containing an error, like so:

screen shot 2016-03-11 at 11 13 37 pm

This makes it easier to diagnose failures in the editor itself, without needing to jump over to the build log to examine the raw output from `xctest_checker`


mike-ferris-apple:
Very cool.

modocache:
I tried this out locally, and it's really amazing!! :heart_eyes: 

I wonder if we can improve `xctest_checker` output such that all assertions have a corresponding file and line number. But that's a discussion for another pull request. Thanks, @briancroom!!

PR #74, opened 2016-03-13
[Python] Add flake8 config

modocache:
In https://github.com/apple/swift/pull/1153, @practicalswift added a flake8 config to the Swift project, in order to lint the Python scripts used by that project. This does the same for XCTest.

This includes all the linting rules active on the main Swift repository. It also includes tweaks to `build_script.py`, which bring this project in full compliance with those linting rules.

To lint the Python code in the project:

    $ flake8

To install flake8:

    $ pip install flake8

See https://flake8.readthedocs.org/en/latest/ for details.

### DIFF
--- a/.pep8
+++ b/.pep8
@@ -1,0 +1,3 @@
+[flake8]
+filename = *.py
+max-line-length = 80
--- a/build_script.py
+++ b/build_script.py
@@ -18,8 +18,10 @@ import tempfile
 
 SOURCE_DIR = os.path.dirname(os.path.abspath(__file__))
 
+
 def note(msg):
     print("xctest-build: "+msg)
+
 
 def run(command):
     note(command)


briancroom:
Neat!

Do you think it would make sense to have these checks included in CI. Since the `build_script.py` is only really designed to work on Linux at this point, I guess that would mean adding the invocation to the overall Swift build script.

mike-ferris-apple:
This just adds a config file to allow those use of this listing tools for those who may want to, right?

Seems reasonable.

modocache:
> Do you think it would make sense to have these checks included in CI.

I absolutely do! I wonder if @practicalswift has already put any thought into this. Personally I think it's a great idea, but I'm not sure how I would feel if Python linter errors caused CI to fail... thoughts? :raised_hands:

practicalswift:
@modocache Perhaps the first step could be to run `flake8` (if available in `PATH`) as part of the build process, but without failing the build on linter errors.

The `--exit-zero` option could be used for that:

```
--exit-zero           exit with code 0 even if there are errors
```

Simply printing the linting errors/warnings would be a great improvement from the current state of things.

What do you think? :-)

modocache:
I like it! Would we be able to `pip install flake8` as part of the build script, or would we need to ask one of the Apple CI gurus to install them on the build machines?

Either way, you should send a pull request! :rocket:

modocache:
@practicalswift You know what would be *really* cool?

@swift-ci Please lint Python

:grin:

practicalswift:
@modocache That would be really nice! Who is in charge of the CI buildbot? Is it you @shahmishal? :-)

shahmishal:
Yes, I will set this up on CI can you please create a Jira ticket. Thanks!

PR #75, opened 2016-03-14
TimeInterval -> NSTimeInterval

briancroom:
`NSTimeInterval` is already in use in some places from @modocache's async testing work. Now that we have the Foundation dependency, we should take advantage of it and use `NSTimeInterval` consistently instead of re-declaring our own.

### DIFF
--- a/Sources/XCTest/XCTimeUtilities.swift
+++ b/Sources/XCTest/XCTimeUtilities.swift
@@ -13,24 +13,24 @@
 
 #if os(Linux) || os(FreeBSD)
     import Glibc
+    import Foundation
 #else
     import Darwin
+    import SwiftFoundation
 #endif
 
-internal typealias TimeInterval = Double
-
 /// Returns the number of seconds since the reference time as a Double.
-private func currentTimeIntervalSinceReferenceTime() -> TimeInterval {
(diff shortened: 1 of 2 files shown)


mike-ferris-apple:
@swift-ci please test

mike-ferris-apple:
Cool.

Sources/XCTest/XCTimeUtilities.swift
```diff
@@ -13,24 +13,24 @@
 
 #if os(Linux) || os(FreeBSD)
     import Glibc
+    import Foundation
 #else
     import Darwin
+    import SwiftFoundation
```
modocache: Perhaps we should borrow a page from swift-package-manager and [extract these checks into a single location](https://github.com/apple/swift-package-manager/blob/eb8c15ce30cccd5c53a5d4d7367a94bc3df0d084/Sources/libc/libc.swift#L11-L15).

Something to consider for another pull request! :point_up:

modocache:
The OS X check failed due to something in Foundation. I'm planning on taking a hard look at getting OS X CI working once and for all... :fire: 

Thanks, @briancroom! :+1:

PR #79, opened 2016-03-22
Added a line break.

marcusrossel:
Moved the `return nil` statement of a `default` switch-case to a new line, as this fits the pattern of the other cases.


modocache:
To be fair, this repository does contain some return statements on the same line: https://github.com/apple/swift-corelibs-xctest/blob/fd67e2fda318012bb9d80c8ec48923865afe83b4/Sources/XCTest/XCTAssert.swift#L31

Still, this change looks good to me. Thanks!

marcusrossel:
Yes. As I mentioned, I changed it, as it does not fit the pattern of the other cases.

PR #80, opened 2016-03-23
Various minor changes to XCTest-files.

marcusrossel:
This pull request contains multiple minor changes:
- Most of them are to the style of comments
- Some are to the formatting of code
- Few change actual code

This pull request mainly aims at making the files more consistent, and readable.

### DIFF
--- a/Sources/XCTest/TestFiltering.swift
+++ b/Sources/XCTest/TestFiltering.swift
@@ -21,15 +21,10 @@ internal struct TestFiltering {
     }
 
     var selectedTestFilter: TestFilter {
-        if let selectedTestName = selectedTestName {
-            if let selectedTest = SelectedTest(selectedTestName: selectedTestName) {
-                return selectedTest.matches
-            } else {
-                return excludeAllFilter()
-            }
-        } else {
-            return includeAllFilter()
-        }
+        guard let selectedTestName = selectedTestName else { return includeAllFilter() }
+        guard let selectedTest = SelectedTest(selectedTestName: selectedTestName) else { return excludeAllFilter() }
+        
+        return selectedTest.matches
     }
 
     private func excludeAllFilter() -> TestFilter {
--- a/Sources/XCTest/TestFiltering.swift
+++ b/Sources/XCTest/TestFiltering.swift
@@ -8,7 +8,7 @@
 //
 //
 //  XCTestFiltering.swift
-//  This provides utilities for executing only a subset of the tests provided to XCTMain
+//  This provides utilities for executing only a subset of the tests provided to `XCTMain`
 //
 
 internal typealias TestFilter = (XCTestCase.Type, String) -> Bool
--- a/Sources/XCTest/TestFiltering.swift
+++ b/Sources/XCTest/TestFiltering.swift
@@ -23,7 +23,7 @@ internal struct TestFiltering {
     var selectedTestFilter: TestFilter {
         guard let selectedTestName = selectedTestName else { return includeAllFilter() }
         guard let selectedTest = SelectedTest(selectedTestName: selectedTestName) else { return excludeAllFilter() }
-        
+
         return selectedTest.matches
     }
 
@@ -37,12 +37,12 @@ internal struct TestFiltering {
 
     static func filterTests(entries: [XCTestCaseEntry], filter: TestFilter) -> [XCTestCaseEntry] {
         return entries
-            .map({ testCase, tests in
-                return (testCase, tests.filter({ filter(testCase, $0.0) }))
-            })
-            .filter({ testCase, tests in
+            .map { testCase, tests in
+                return (testCase, tests.filter { filter(testCase, $0.0) } )
+            }
+            .filter { testCase, tests in
                 return !tests.isEmpty
-            })
+            }
     }
 }
 
--- a/Sources/XCTest/XCTAssert.swift
+++ b/Sources/XCTest/XCTAssert.swift
@@ -52,25 +52,19 @@ private enum _XCTAssertionResult {
     case UnexpectedFailure(ErrorProtocol)
 
     var expected: Bool {
-        switch (self) {
-        case .UnexpectedFailure(_):
-            return false
-        default:
-            return true
+        switch self {
+        case .UnexpectedFailure(_): return false
+        default: return true
         }
     }
 
     func failureDescription(assertion: _XCTAssertion) -> String {
         let explanation: String
-        switch (self) {
-        case .Success:
-            explanation = "passed"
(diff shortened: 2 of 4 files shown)


modocache:
Thanks, @marcusrossel! I'm supportive of many of these changes. :100:

Sources/XCTest/TestFiltering.swift
```diff
@@ -21,15 +21,10 @@ internal struct TestFiltering {
...
-            } else {
-                return excludeAllFilter()
-            }
-        } else {
-            return includeAllFilter()
-        }
+        guard let selectedTestName = selectedTestName else { return includeAllFilter() }
+        guard let selectedTest = SelectedTest(selectedTestName: selectedTestName) else { return excludeAllFilter() }
+
+        return selectedTest.matches
```
modocache: @briancroom Any opinions here? I find both equally readable.

Sources/XCTest/TestFiltering.swift
```diff
@@ -42,12 +37,12 @@ internal struct TestFiltering {
...
-                return (testCase, tests.filter({ filter(testCase, $0.0) }))
-            })
-            .filter({ testCase, tests in
+            .map { testCase, tests in
+                return (testCase, tests.filter { filter(testCase, $0.0) } )
+            }
+            .filter { testCase, tests in
                 return !tests.isEmpty
-            })
+            }
```
modocache: @briancroom Same here. I don't have a strong opinion, I think either way of writing it is fine.

Sources/XCTest/XCTAssert.swift
```diff
@@ -52,25 +52,19 @@ private enum _XCTAssertionResult {
     case UnexpectedFailure(ErrorProtocol)
 
     var expected: Bool {
-        switch (self) {
-        case .UnexpectedFailure(_):
-            return false
-        default:
-            return true
+        switch self {
+        case .UnexpectedFailure(_): return false
```
modocache: In #79, you made changes that placed a line break after a `case` statement. Here you've removed line breaks. Do you have a clear goal in mind here?

Sources/XCTest/TestFiltering.swift
briancroom: I like this change! The flatter structure is nice.

Sources/XCTest/TestFiltering.swift
briancroom: I'm pretty ambivalent about this one. I've been pretty happy following the [Rule of Kevin](http://ericasadun.com/2015/11/17/a-handful-of-swift-style-rules-swiftlang/) in my Swift code, but I don't have a strong opinion.

briancroom:
Cool! Thanks for taking the time to look through the code so thoroughly @marcusrossel. I've followed up on a couple points in the comments.

Sources/XCTest/XCTAssert.swift
```diff
@@ -100,15 +100,15 @@ private func _XCTEvaluateAssertion(assertion: _XCTAssertion, @autoclosure messag
     }
 }
 
-/// This function emits a test failure if the general Bool expression passed
-/// to it evaluates to false.
+/// This function emits a test failure if the general `Boolean` expression passed
+/// to it evaluates to `false`.
```
mike-ferris-apple: I would suggest either keeping it "Bool" and adding the backticks since that is a type name or changing to Boolean and not adding the backpacks since then it's just prose...

Sources/XCTest/TestFiltering.swift
mike-ferris-apple: Me too. Shorter and less nested.

Sources/XCTest/TestFiltering.swift
marcusrossel: So, I've searched some files in *swift/stdlib/public/core* to try to find occurrences of the string `"})"` (which would indicate the use of non-trailing syntax). I found some cases:

    return split(
      maxSplits: maxSplits,
      omittingEmptySubsequences: omittingEmptySubsequences,
      isSeparator: { $0 == separator })

---

    else if let offset = (e as? Int).map({ IntMax($0) }) ?? (e as? IntMax) { //... }

---

...

I'm not quite sure in terms of consistency, when it comes to use of trailing closures. 
But the general use of non-trailing syntax would indicate to me, that there are appropriate situations for its use. I'm therefore leaning towards the Rule Of Kevin.

Sources/XCTest/XCTAssert.swift
marcusrossel: I tried to explain my intent in the commit.
When the statements inside of a case fit on one line (for all the cases in the switch), I moved them next to the case-label.
If any case required multiple lines, I added a line break.
This results in consistency within a switch-statement, but not necessarily across different switches.

I use this style in my own projects, as it's consistent, but also reduces the size of switches, with short expressions in each case.

Sources/XCTest/XCTAssert.swift
marcusrossel: I understand your point. What I don't understand though is: if `Boolean` is not a defined type, how can it be used as one?:
`@autoclosure expression: () throws -> Boolean`

This code was taken from XCTAssert.swift

mike-ferris-apple: You're absolutely right! I forgot about the Boolean protocol as distinct from the Bool concrete type.

Never mind!

briancroom: Possibly contributing to the confusion is the fact that this was previously named BooleanType before the recent Swift 3 stdlib naming changes.

marcusrossel: Ah, I see. I haven't been aware of that name change.
So, I'll keep the changes, as they were (if that's the consensus).

Sources/XCTest/XCTAssert.swift
modocache: Makes sense to me. Thanks, and sorry I didn't realize you had already explained in the commit messages!

modocache:
I think this is good to merge. @mike-ferris-apple, could you ask @swift-ci to please test?

mike-ferris-apple:
@swift-ci please test

Sources/XCTest/XCTAssert.swift
marcusrossel: Cool.
Just as a side-question: Is there way to make a commit that affects multiple files? I'm editing the code online, so when I try to switch to another file, it would not save the changes from the current one.

modocache:
Thanks, @marcusrossel! And thanks for the reviews everyone!

marcusrossel:
Thanks for helping me out guys/gals.

PR #81, opened 2016-03-25
[XCTestObservation] Add test bundle observation

modocache:
Apple XCTest defines more methods on its XCTestObservation than swift-corelibs-xctest does. Here we add two of these missing methods:

- `testBundleWillStart()`
- `testBundleDidFinish()`

The documentation for the two methods is taken from Apple XCTest's headers.

### DIFF
--- a/Sources/XCTest/XCTestMain.swift
+++ b/Sources/XCTest/XCTestMain.swift
@@ -79,6 +79,10 @@ internal struct XCTRun {
 /// - Parameter testCases: An array of test cases run, each produced by a call to the `testCase` function
 /// - seealso: `testCase`
 @noreturn public func XCTMain(testCases: [XCTestCaseEntry]) {
+    let observationCenter = XCTestObservationCenter.sharedTestObservationCenter()
+    let testBundle = NSBundle.mainBundle()
+    observationCenter.testBundleWillStart(testBundle)
+
     let filter = TestFiltering()
 
     let overallDuration = measureTimeExecutingBlock {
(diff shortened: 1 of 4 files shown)


mike-ferris-apple:
@swift-ci please test

mike-ferris-apple:
I guess this maps reasonably.  What does mainBundle return for a linux command line tool (or an OS X one, for that matter)?

modocache:
Hmm, the Linux CI failure is an unrelated failure in the Swift test suite--strange, since [we `--skip-test-swift` for the build preset](https://github.com/apple/swift/blob/3048629bb5830296a9d7ace48e006b7b5f1c3c4a/utils/build-presets.ini#L272). Maybe that flag doesn't skip validation tests... I'll modify our preset such that we don't get blocked by intermittent failures in Swift proper.

As for command-line tool bundles:
- On OS X, using Xcode 7.2's command-line tool project template, printing `[NSBundle mainBundle]` outputs: `NSBundle </Users/bgesiak/Library/Developer/Xcode/DerivedData/BundleBundleOnTheWall-ewdkqrksmgkyedgcxaivcypxfoaq/Build/Products/Debug> (loaded)`
- On Linux, in a simple Swift executable linked with Foundation, `NSBundle.mainBundle()` returns a temporary bundle at a path like `/tmp/tmpZtm9Of`.

I'm not sure how similar these two are, but I also think improving the bundle behavior could be something we improve upon later. I consider the fact that the announcements are made with the same signature as Apple XCTest to be the main improvement here.

mike-ferris-apple:
Agreed.

modocache:
@mike-ferris-apple Great! Normally I'd ask you to kick off CI again, but I'm not entirely sure it's been fixed. I've tested this on Linux and OS X and it definitely works for me, so... merging! :grin:

PR #83, opened 2016-03-26
General Code Overhaul

marcusrossel:
In this pull request I went through every .swift file in XCTest, and updated the code if appropriate.
This includes reformatting, switching from `if` to `guard`, utilizing the ternary operator, switching from `var` to `let`, and fixing problems with array literals, which should have caused compilation errors.

For the details on each file, check the commit's description.

### DIFF
--- a/Sources/XCTest/XCTestCase.swift
+++ b/Sources/XCTest/XCTestCase.swift
@@ -92,7 +92,7 @@ extension XCTestCase {
                 let testCase = self.init()
                 testCase.name = "\(testCase.dynamicType).\(name)"
 
-                var failures = [XCTFailure]()
+                var failures: [XCTFailure] = []
                 XCTFailureHandler = { failure in
                     observationCenter.testCase(testCase,
                                                didFailWithDescription: failure.failureMessage,
@@ -146,8 +146,8 @@ extension XCTestCase {
             }
         }
 
-        let testCountSuffix = (tests.count == 1) ? "" : "s"
-        let failureSuffix = (totalFailures == 1) ? "" : "s"
+        let testCountSuffix = tests.count == 1 ? "" : "s"
+        let failureSuffix = totalFailures == 1 ? "" : "s"
 
         XCTPrint("Executed \(tests.count) test\(testCountSuffix), with \(totalFailures) failure\(failureSuffix) (\(unexpectedFailures) unexpected) in \(printableStringForTimeInterval(totalDuration)) (\(printableStringForTimeInterval(overallDuration))) seconds")
     }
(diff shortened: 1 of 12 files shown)


Sources/XCTest/XCTestCase.swift
```diff
@@ -92,7 +92,7 @@ extension XCTestCase {
                 let testCase = self.init()
                 testCase.name = "\(testCase.dynamicType).\(name)"
 
-                var failures = [XCTFailure]()
+                var failures: [XCTFailure] = []
```
modocache: I've seen [style guides that argue the former syntax is preferable](https://github.com/raywenderlich/swift-style-guide#type-inference). The convention in swift-corelibs-xctest also appears to be the former syntax. I don't have a strong preference either way, but I do think it's better to keep whatever convention is already being used, rather than make a large change such as this one.

Tests/Functional/FailingTestSuite/main.swift
```diff
@@ -10,9 +10,7 @@
 
 class PassingTestCase: XCTestCase {
     static var allTests: [(String, PassingTestCase -> () throws -> Void)] {
-        return [
-            ("test_passes", test_passes),
-        ]
+        return [("test_passes", test_passes)]
```
modocache: I actually prefer the former syntax. Listing each element on a separate line and including a trailing comma for each element ensures the diff is small when an element is added or removed.

But my preference notwithstanding, once again the convention for listing test cases appears to be one per line. I think convention trumps preference in this case.

Tests/Functional/SelectedTest/main.swift
```diff
@@ -26,19 +26,20 @@ class ExecutedTestCase: XCTestCase {
 // CHECK-TESTCASE: Test Case 'ExecutedTestCase.test_bar' passed \(\d+\.\d+ seconds\).
 // CHECK-ALL:      Test Case 'ExecutedTestCase.test_bar' started.
 // CHECK-ALL:      Test Case 'ExecutedTestCase.test_bar' passed \(\d+\.\d+ seconds\).
-    func test_bar() {}
+    func test_bar() {
+    }
```
modocache: Here again I personally prefer the former syntax. Why use two lines to represent an empty function? Is the value added here worth making this change?

modocache:
Thanks, @marcusrossel! I agree with some of the changes around ternary statements, `guard` statements, using `_` for unused variables, and writing `swift self` instead of `switch(self)`. I'd be happy to merge those as separate pull requests!

parkera:
I think we shouldn't necessarily require terser ternary statements in all cases. Sometimes it's clearly to just do an `if foo != nil`.

Also, we should rebase all of this into one commit before merging.

Sources/XCTest/XCTestCase.swift
marcusrossel: Yes, that makes sense. I'll change it back to how it was.

Tests/Functional/FailingTestSuite/main.swift
marcusrossel: I can agree on the 'one element per line', but I disagree on the trailing comma. Before this comment, I was not aware that a trailing comma was legal syntax. It could therefore be confusing to newcomers (or even people who were unaware of this).

Tests/Functional/SelectedTest/main.swift
marcusrossel: I also prefer the former syntax. I have made this change, as other methods in XCTest use this syntax. I could change those 'other' methods to this syntax though (which I would also prefer).

marcusrossel:
@parkera I agree that ternary statements should not always be preferred to if-statements. I find them useful though when there are two simple return values. They reduce the mental load of reading through an `if { return/set } else { return/set }`, when all you want to do is set a value based on a condition.

marcusrossel:
@modocache I'll setup a new pull request, once more details have been discussed.
Also... How would I go about putting all of the changes in one Commit, when editing the files online?

briancroom:
@marcusrossel As far as I know, you will have to use `git` from the command line to squash the commits into one. There are some pointers [here](https://help.github.com/articles/about-git-rebase/) if you haven't tried doing that before!

Tests/Functional/FailingTestSuite/main.swift
briancroom: I would also be in favor of keeping the [trailing comma](https://twitter.com/jaybuff/status/710918302185291776) to reduce the number of lines touched when the list of tests is changed.

marcusrossel:
Thank you, I'll go have a look.

PR #84, opened 2016-03-27
[XCTestObservation] Add XCTestSuite announcements

modocache:

Apple XCTest's `XCTestObservation` protocol includes methods that announce when an `XCTestSuite` will begin executing, or has finished executing. Add these same announcements to swift-corelibs-xctest.

swift-corelibs-xctest did not already defined `XCTestSuite`, so add it.

Apple XCTest allows `XCTestObservation` listeners to determine the number of passing and failing tests by accessing an `XCTestSuite`'s `testRun` property, but this is not yet supported in swift-corelibs-xctest.

- This makes the swift-corelibs-xctest and Apple XCTest versions of `XCTestObservation` identical, bringing us closer to our Swift 3 goal of API parity.

## What are the downsides of merging this pull request?

- This adds `XCTestSuite` to the codebase, but not `XCTestRun`. As a result, the most interesting aspect of test announcements (how many tests passed/failed) is still unimplemented. Still, I feel like this is a positive step forward, and one that keeps this pull request small enough to actually review.
- swift-corelibs-xctest doesn't use `XCTestSuite` to represent the collection of test cases it executes. As a result, the quality of implementation here leaves much to be desired. On the other hand, refactoring the codebase to seamless integrate `XCTestSuite` would involve more widespread changes, which would make this pull request more difficult to review. If you ask me, thanks to the functional test cases, we could merge this, then easily refactor in a subsequent pull request.

### DIFF
--- a/Sources/XCTest/XCAbstractTest.swift
+++ b/Sources/XCTest/XCAbstractTest.swift
@@ -1,0 +1,36 @@
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2016 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//
+//  XCAbstractTest.swift
+//  An abstract base class that XCTestCase and XCTestSuite inherit from.
+//  The purpose of this class is to mirror the design of Apple XCTest.
+//
+
+/// An abstract base class for testing. `XCTestCase` and `XCTestSuite` extend
+/// `XCTest` to provide for creating, managing, and executing tests. Most
+/// developers will not need to subclass `XCTest` directly.
+public class XCTest {
+    /// Test's name. Must be overridden by subclasses.
+    public var name: String {
+        fatalError("Must be overridden by subclasses.")
+    }
+
+    /// Number of test cases. Must be overridden by subclasses.
+    public var testCaseCount: UInt {
+        fatalError("Must be overridden by subclasses.")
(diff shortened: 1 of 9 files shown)


mike-ferris-apple:
This seems like a reasonable first step. I agree generally with your assessment in the second "downside" you mention. The weirdness around having the "All" suite and the bundle suite be peers in a list of root suites instead of having a containment relationship arises from this, I guess. I would definitely like to see continued movement towards having test suites constructed and then used to control the running of the tests instead of just being constructed in order to send the observations. But I am OK with taking this in incremental stages.

mike-ferris-apple:
@swift-ci please test

modocache:
Here's the error from Linux CI:

```
/usr/bin/ld: /home/buildnode/jenkins/workspace/swift-corelibs-xctest-PR-Linux/Ninja-ReleaseAssert/swiftpm-linux-x86_64/debug/test-Package: hidden symbol `_TFC6XCTest10XCTestCasemP33_3BE257A46ADB477C7BF2D39968B39F9D5_nameSS' isn't defined
/usr/bin/ld: final link failed: Bad value
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

I *think* this is the error @briancroom mentioned in #69 (is there a JIRA issue for this problem?). I'll try to address this in the coming days.

modocache:
@mike-ferris-apple I've resolved the Linux build issue. Could you ask @swift-ci to please test? If the tests pass, I'll merge this to achieve the "incremental stages" you describe (or you can request "@swift-ci please test and merge").

#86 is the realization of the "continued movement towards having test suites constructed and then used to control the running of the tests" you describe.

parkera:
@swift-ci please test

PR #87, opened 2016-04-03
[XCTestCase] Workaround SR-1129 using public var

modocache:
`XCTestCase.name` works around a bug in the Linux Swift compiler ([SR-1129](https://bugs.swift.org/browse/SR-1129)) by declaring what should be a `private` variable as `public`. `XCTAllExpectations` works around the same bug by using a global variable. Unify the two by using the same workaround method: declare the expectations variable as `public`.

Besides using the same workaround, this has the added benefit of reducing the amount of global state in XCTest.

### DIFF
--- a/Sources/XCTest/XCTestCase.swift
+++ b/Sources/XCTest/XCTestCase.swift
@@ -39,6 +39,14 @@ public class XCTestCase: XCTest {
     ///   https://bugs.swift.org/browse/SR-1129 for details.
     public var _name: String
 
+    /// The set of expectations made upon this test case.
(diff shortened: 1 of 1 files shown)


mike-ferris-apple:
@swift-ci please test

mike-ferris-apple:
This looks like a good change and worth the "public" workaround for avoiding global state.

briancroom:
How about prefixing the property name with an underscore for now to help reinforce its intended private-ness and reduce the chance of a naming collision, since the class is intended to be subclassed?

modocache:
@briancroom Done! Good suggestion, thanks!

PR #82, opened 2016-03-26
[XCTestObservation] Replace sharedTestObservationCenter with shared

shingt:
Replaced `sharedTestObservationCenter` with `shared` as described in https://bugs.swift.org/browse/SR-1063.

PR #88, opened 2016-04-04
Additional Swift 3 API naming updates

briancroom:
#82 got us started with updating our own APIs to match the Swift 3 guidelines, but there are a number of other changes required for us to be fully compatible. Here, I've adjusted a few public methods to match how they are imported into Swift from Apple XCTest with the Swift 3 toolchain. I have also made a handful of changes to internal APIs to bring them in line with the Swift 3 naming [guidelines](https://github.com/apple/swift-evolution/blob/master/proposals/0006-apply-api-guidelines-to-the-standard-library.md). See the commit messages for more specifics!

### DIFF
--- a/Sources/XCTest/XCTestCase.swift
+++ b/Sources/XCTest/XCTestCase.swift
@@ -190,7 +190,7 @@ extension XCTestCase {
     ///   between these environments. To ensure compatibility of tests between
     ///   swift-corelibs-xctest and Apple XCTest, it is not recommended to pass
     ///   explicit values for `file` and `line`.
-    public func expectationWithDescription(description: String, file: StaticString = #file, line: UInt = #line) -> XCTestExpectation {
+    public func expectation(withDescription description: String, file: StaticString = #file, line: UInt = #line) -> XCTestExpectation {
         let expectation = XCTestExpectation(
             description: description,
             file: file,
@@ -225,7 +225,7 @@ extension XCTestCase {
     ///   these environments. To ensure compatibility of tests between
     ///   swift-corelibs-xctest and Apple XCTest, it is not recommended to pass
     ///   explicit values for `file` and `line`.
-    public func waitForExpectationsWithTimeout(timeout: NSTimeInterval, file: StaticString = #file, line: UInt = #line, handler: XCWaitCompletionHandler?) {
+    public func waitForExpectations(withTimeout timeout: NSTimeInterval, file: StaticString = #file, line: UInt = #line, handler: XCWaitCompletionHandler?) {
         // Mirror Objective-C XCTest behavior; display an unexpected test
         // failure when users wait without having first set expectations.
         // FIXME: Objective-C XCTest raises an exception for most "API
@@ -323,9 +323,9 @@ extension XCTestCase {
     ///   notification is observed. It will not be invoked on timeout. Use the
     ///   handler to further investigate if the notification fulfills the 
     ///   expectation.
-    public func expectationForNotification(notificationName: String, object objectToObserve: AnyObject?, handler: XCNotificationExpectationHandler?) -> XCTestExpectation {
+    public func expectation(forNotification notificationName: String, object objectToObserve: AnyObject?, handler: XCNotificationExpectationHandler?) -> XCTestExpectation {
         let objectDescription = objectToObserve == nil ? "any object" : "\(objectToObserve!)"
-        let expectation = self.expectationWithDescription("Expect notification '\(notificationName)' from " + objectDescription)
+        let expectation = self.expectation(withDescription: "Expect notification '\(notificationName)' from " + objectDescription)
         // Start observing the notification with specified name and object.
         var observer: NSObjectProtocol? = nil
         func removeObserver() {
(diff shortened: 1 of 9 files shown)


Sources/XCTest/XCTestCase.swift
```diff
@@ -189,7 +189,7 @@ extension XCTestCase {
     ///   between these environments. To ensure compatibility of tests between
     ///   swift-corelibs-xctest and Apple XCTest, it is not recommended to pass
     ///   explicit values for `file` and `line`.
-    public func expectationWithDescription(description: String, file: StaticString = #file, line: UInt = #line) -> XCTestExpectation {
+    public func expectation(withDescription description: String, file: StaticString = #file, line: UInt = #line) -> XCTestExpectation {
```
gribozavr: "With" is usually a vacuous preposition that does not carry any semantics.  I think this is one of those cases.  (Other cases in XCTest may also be like that.)

briancroom: Agreed. However this is the method name as generated by the importer with the March 24 toolchain from `XCTest.framework`. Do you expect additional changes to the importer which would automatically remove "with"? Or is this a situation where the Objective-C header should be annotated with the `swift_name` attribute?

(Interestingly, it appears that https://github.com/apple/swift-3-api-guidelines-review doesn't include `XCTest.framework`)

parkera: This should match exactly what the importer does, which is leaving the with in place (and this is very much an intentional choice).

modocache:
Looks good to me! @parkera or @mike-ferris-apple, could you kick off the tests?

Also, I'm curious: what's the best way to see the Swift headers generated for Apple XCTest? I use "Jump to Definition" in Xcode (with my desired toolchain selected) to see the Swift headers. Is there a better way, perhaps something I can invoke from the command line?

mike-ferris-apple:
@swift-ci please test

modocache:
Ruh-roh, build failure. @briancroom, looks like you missed a spot:

```
Sources/XCTest/XCTestCase.swift:265:21: error: value of type 'XCTestExpectation' has no member 'fulfilled'
                if !expectation.fulfilled {
                    ^~~~~~~~~~~ ~~~~~~~~~
```

parkera:
One trick I've been using is to start up the `swift` REPL after building the whole stack from scratch. You can tab-complete in there to get a list of API on a type.

modocache:
`utils/swift-api-dump.py` seems like what I need, although it doesn't cover XCTest--that script exports headers from `/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/`, but XCTest exists at `/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/Library/Frameworks/. Maybe I'll send a pull request... :wink:

parkera:
Nice, I didn't know about that script. Thanks @modocache.

briancroom:
Oh man, thanks CI for pointing out my incomplete commit :flushed: I've fixed that now, rebased, and added one additional Swift 3 import change I had previously missed - default `nil` values for optional `handler` params.

Also thanks for the `swift-api-dump.py` tip - so far I've been viewing the Generated Interface from within Xcode to see these things.

Sources/XCTest/XCTestCase.swift
briancroom: @parkera is there any additional discussion around this particular choice you could point me to? I'm curious about the context on this.

parkera: It was a discussion we had in our Swift 3 naming 'working group'.

The summary is that we felt that the importer could not automatically make the right decision about if a 'with' was important or not in an API. It is used to mean different things in different APIs. Therefore we decided to stick with a default of leaving it in place for imported API.

We can of course change our minds about the names, but then we're talking about changing it for all clients on all platforms. Since this version of XCTest is attempting to keep the same API, and there isn't a plan to change the Darwin version of this method, we should keep the name the same for now.

briancroom: Thanks, that's exactly the info I was looking for.

modocache:
Ping @mike-ferris-apple for another test kickoff -- how about "@swift-ci please test and merge"? 😁

modocache:
It turns out `swift-api-dump.py` can generate XCTest headers as well, you just need to know which arguments to pass. I uploaded the current set of headers here, along with the command: https://github.com/modocache/XCTest.swift

mike-ferris-apple:
@swift-ci please test

briancroom:
Ah, looks like CI failed because https://github.com/apple/swift-corelibs-xctest/pull/90 was merged which contains code that needs updating. Hang tight...

briancroom:
Rebased and fixed up now. @mike-ferris-apple could I bother you for another CI test request?

harlanhaskins:
@swift-ci please test

mike-ferris-apple:
@swift-ci please test

mike-ferris-apple:
Ah, oops. Didn't see Harlan's request there.

harlanhaskins:
Nah, I'm not on the CI list quite yet 👍

modocache:
One day https://github.com/apple/swift/pull/2036 will be merged and the Swift Test OS X Platform check will succeed. Today is not that day 😂 

In the meantime, looks like Linux passed, so I'm merging this. Thanks as always, @briancroom! 💯

PR #89, opened 2016-04-04
Translate Objective-C code in `XCTestSuite` doc comments into Swift

briancroom:
The doc comments on `XCTestSuite` contain some Objective-C example code. This updates the example code to be Swift instead.

### DIFF
--- a/Sources/XCTest/XCTestSuite.swift
+++ b/Sources/XCTest/XCTestSuite.swift
@@ -15,20 +15,20 @@
 /// Suites are usually managed by the IDE, but XCTestSuite also provides API
 /// for dynamic test and suite management:
 ///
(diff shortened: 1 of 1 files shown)


briancroom:
After having blindly opened this, I realize that the entire content of this comment should be revisited, since the current text is pretty heavily based upon the Apple XCTest implementation which relies on the Obj-C runtime. I'll open something new with a more intelligent diff later.

modocache:
Yes, thanks for noticing!! Open a JIRA issue (or I will): so far I've been copying Apple XCTest documentation, but at some point we should totally revamp it. I think we can do even better than Apple here: I'd love it if we could clearly explain the relationship between test cases, suites, and runs.

briancroom:
Created: https://bugs.swift.org/browse/SR-1145

PR #90, opened 2016-04-04
Don't call a notification expectation handler after the expectation has expired

briancroom:
Previously, if a notification expectation fails, either due to a `waitForExpectationsWithTimeout` call or because the test ends, the internal notification observer could still end up calling the expectation handler if the expected notification is fired at a later point. With this change, the notification observer weakly captures the expectation, and won't do any work if the notification is observed but the expectation isn't around anymore.


mike-ferris-apple:
@swift-ci please test

modocache:
Good catch, thanks @briancroom!

PR #92, opened 2016-04-06
[SE-0046] Implemented consistent function labels

harlanhaskins:
Implemented changes specified in [SE-0046 Establish consistent label behavior across all parameters including first labels](https://github.com/apple/swift-evolution/blob/master/proposals/0046-first-label.md)

This change is pending [apple/swift#2047](https://github.com/apple/swift/pull/2047)

### DIFF
--- a/Sources/XCTest/TestFiltering.swift
+++ b/Sources/XCTest/TestFiltering.swift
@@ -35,7 +35,7 @@ internal struct TestFiltering {
         return { _ in true }
     }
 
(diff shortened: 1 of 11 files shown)


harlanhaskins:
@mike-ferris-apple We're going to merge this as soon as we merge the compiler work. Is that alright?

mike-ferris-apple:
Yes. It seems like this needs to go at the same time as the compiler stuff and I don't think this particular tail should argue in favor of wagging that dog.

harlanhaskins:
@mike-ferris-apple Can I get a merge?

harlanhaskins:
Or @modocache

PR #96, opened 2016-04-14
Use Linux Glibc and Foundation on Android

modocache:
Expand the Linux/FreeBSD check to include Android. The "Android" OS will be supported by the next snapshot (which this commit will also be a part of).

---

corelibs-xctest won't work on Android yet, even after this change. For it to work, we'll need to port corelibs-foundation as well. Still, consider this a bit of preliminary work.

### DIFF
--- a/Sources/XCTest/PrintObserver.swift
+++ b/Sources/XCTest/PrintObserver.swift
@@ -11,7 +11,7 @@
 //  Prints test progress to stdout.
 //
 
-#if os(Linux) || os(FreeBSD)
+#if os(Linux) || os(FreeBSD) || os(Android)
     import Foundation
 #else
     import SwiftFoundation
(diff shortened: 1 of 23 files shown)


briancroom:
👍 cool! Looking forward to further progress being made on this in the coming weeks.

mike-ferris-apple:
Would it be better to write this as special for OS X? I guess arguably we have the same problem there if we consider all the Apple platforms, but, on the other hand, the ability to build this on OS X at all is mostly a convenience for those developing it, so I'm not sure trying to support the other Apple platforms makes much sense anyway...

briancroom:
I agree that OS X is the only Darwin platform that we should be supporting here, so reversing the conditional does seem like a sensible move.

modocache:
OK, will do soon! 👍

modocache:
This is still something I'm interested in doing someday, but I'll close this for now, in order to clean up outstanding pull requests against this repository.

PR #97, opened 2016-04-16
Enhance the build script to support building and testing on Darwin

briancroom:
As discussed in https://github.com/apple/swift/pull/2137, we are getting the feeling that we should move build script logic for Darwin out of Swift's build-script-impl and into XCTest's own build script to reduce churn and enable better CI feedback. This PR is a first step towards making that happen. 

* Build and test are performed using `xcodebuild` actions, as the Swift
  build-script-impl script has been doing in the past
* Installation is not supported on Darwin
* Uses the strategy pattern to isolate platform branching logic

### DIFF
--- a/build_script.py
+++ b/build_script.py
@@ -16,6 +16,7 @@ import subprocess
 import sys
 import tempfile
 import textwrap
+import platform
 
 SOURCE_DIR = os.path.dirname(os.path.abspath(__file__))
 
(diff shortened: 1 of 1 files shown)


parkera:
This seems reasonable to me. I had always intended this script to be a stop gap until swift-pm was read enough to build all the things on all the platforms. Hopefully the script is a step toward unifying build logic and will help when that day comes.

build_script.py
```diff
@@ -184,29 +257,34 @@ def main(args=sys.argv[1:]):
     delegates building or testing XCTest to a sub-parser and its corresponding
     function.
     """
+    strategy = DarwinStrategy if platform.system() == 'Darwin' else LinuxStrategy
```
modocache: I've actually never tried building corelibs-xctest on FreeBSD, but I imagine it should succeed using the `LinuxStrategy`. However, I've been told that it's something of a faux pas to refer to FreeBSD as Linux. Could we rename this to `UnixStrategy` or `GenericUnixStrategy`? This would match the naming scheme of [Swift's toolchains](https://github.com/apple/swift/blob/2af7041af32c0fbda8d1a77a605905c0a99c8d02/lib/Driver/ToolChains.h#L39).

build_script.py
```diff
@@ -36,146 +37,218 @@ def _mkdirp(path):
...
-        os.path.join(build_dir, xctest_swiftdoc),
-        os.path.join(module_install_path, xctest_swiftdoc)))
+class DarwinStrategy:
+    @staticmethod
+    def requires_foundation_build_dir():
+        """
+        The Foundation build directory is not required on Darwin because the
+        Xcode workspace implicitly builds Foundation when building the XCTest
+        schemes.
+        """
```
modocache: nit-pick: This is a great explanation, but it seems like more a `# source comment` than `"""documentation"""`.

build_script.py
```diff
@@ -36,146 +37,218 @@ def _mkdirp(path):
...
+        swiftc = os.path.abspath(args.swiftc)
+        build_dir = os.path.abspath(args.build_dir)
+
+        run("xcodebuild -workspace {source_dir}/XCTest.xcworkspace "
+            "-scheme SwiftXCTestFunctionalTests "
+            "SWIFT_EXEC=\"{swiftc}\" "
+            "SWIFT_LINK_OBJC_RUNTIME=YES "
+            "SYMROOT=\"{build_dir}\" OBJROOT=\"{build_dir}\" "
+            "| grep -v \"    export\"".format(
```
modocache: Is this `grep` supposed to be here? Maybe an artifact from some debugging you were doing?

modocache:
Big +1 from me! This looks fantastic. 💯 

I imagine this will continue working with the current Swift `utils/build-script`, correct? Let's kick off the test to make sure CI isn't going to break!

briancroom:
Thanks for the feedback, guys! I'll make the suggested adjustments the next time I'm at a dev workstation, which won't be until mid-week I expect.

build_script.py
briancroom: Oh, good eye @modocache! I meant to call that out in the PR description for discussion, but forgot.

The purpose of that grep was to filter out several dozen lines of output that Xcode emits, showing the complete environment that is being exported when it invokes lit as an External Build Tool. I was finding that this output was more distracting than useful when I was using the script to repeatedly run tests while working on XCTest. They said, I'm not particularly attached to this behavior and would have no issue with leaving it out. What do you think?

briancroom:
Alright, I've made the updates now. @parkera @mike-ferris-apple could you request a test run when you have a second?

ddunbar:
@swift-ci please test

briancroom:
Green! Merging now.

briancroom:
Follow-up: https://github.com/apple/swift/pull/2259

PR #100, opened 2016-04-22
Update XCTestSuite comment to Swift

jonallured:
This update also drops the parts describing features of Apple's implementation that are currently unavailable. Those parts are:

* automatic test case extraction (`testSuiteForTestCaseClass`)
* automatic creation of all test cases found in runtime (`defaultTestSuite`)

This PR is for https://bugs.swift.org/browse/SR-1145, hopefully I did this right! :D

### DIFF
--- a/Sources/XCTest/XCTestSuite.swift
+++ b/Sources/XCTest/XCTestSuite.swift
@@ -10,14 +10,14 @@
 //  XCTestSuite.swift
 //  A collection of test cases.
 //
-// A concrete subclass of XCTest, XCTestSuite is a collection of test cases.
-// Suites are usually managed by the IDE, but XCTestSuite can also be
-// instantiated and manipulated directly:
+// A subclass of XCTest, XCTestSuite is a collection of test cases. Based on
+// what's passed into XCTMain(), a hierarchy of suites is built up, but
+// XCTestSuite can also be instantiated and manipulated directly:
 //
 //     let suite = XCTestSuite(name: "My Tests")
 //     suite.addTest(myTest)
 //     suite.testCaseCount // 1
-//     suite.perform(myRunner)
+//     suite.run()
 
 public class XCTestSuite: XCTest {
     public private(set) var tests = [XCTest]()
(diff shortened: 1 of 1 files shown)


modocache:
Excellent, thanks for taking this on, @jonallured! I have a few comments that I'll leave inline...

Sources/XCTest/XCTestSuite.swift
```diff
@@ -10,28 +10,15 @@
 //  XCTestSuite.swift
 //  A collection of test cases.
 //
+// A concrete subclass of XCTest, XCTestSuite is a collection of test cases.
+// Suites are usually managed by the IDE, but XCTestSuite can also be
```
modocache: This is not true in the case of swift-corelibs-xctest. We build up a hierarchy of suites based on the test cases the user passes into `XCTMain()`. No IDE (Xcode, in the case of Apple XCTest) is involved.

Sources/XCTest/XCTestSuite.swift
```diff
@@ -10,28 +10,15 @@
...
+// instantiated and manipulated directly:
+//
+//     let suite = XCTestSuite(name: "My Tests")
+//     suite.addTest(myTest)
+//     suite.testCaseCount // 1
+//     suite.perform(myRunner)
```
modocache: I'd write this as `suite.run()` instead. Alternatively, this could be `suite.perform(myTestRun)`.

jonallured:
Thanks for the feedback @modocache - I just pushed a commit that I think reads better. Any further thoughts?

modocache:
Excellent, thanks! 💯

modocache:
Ack, sorry @jonallured -- I noticed immediately after merging that you used source-level comments (`//`), not documentation comments (`///`). I fixed this in https://github.com/apple/swift-corelibs-xctest/commit/244507b687560170c118c003eee15658f65707c3. Thanks again for the documentation improvements!!

jonallured:
Hey @modocache, I noticed the two vs three slashes, but wasn't sure what their significance was - now I do! :D

PR #101, opened 2016-04-27
Standardize usage of the term "test case" in the code and documentation

briancroom:
While thinking about the terminology we've been using to talk about tests for this framework, I realized that the term "test case" is being used to mean two distinct things:

1) A single test to be executed by the framework (i.e. an `XCTestCase` instance)
2) An `XCTestCase` subclass, including all of its test methods

After thinking about it more carefully and looking through the Apple XCTest header comments<sup>[*](#note1)</sup>, I've come to the conclusion that #1 is the correct usage. This PR is an attempt to improve our consistency around this.

Looking for feedback on both the underlying assumptions and the specific changes I'm proposing here!

<a name="note1">* Note</a>: Assuming folks agree with this usage suggestion, I'll be opening a Radar also, as there are a couple of points where the same confusion can be detected in the Apple XCTest headers.

### DIFF
--- a/Documentation/Linux.md
+++ b/Documentation/Linux.md
@@ -23,7 +23,7 @@ class TestNSURL : XCTestCase {
 }
 ```
 
-Also, this version of XCTest does not use the external test runner binary. Instead, create your own executable which links `libXCTest.so`. In your `main.swift`, invoke the `XCTMain` function with an array of the test cases classes that you wish to run, wrapped by the `testCase` helper function. For example:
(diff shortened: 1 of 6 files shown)


modocache:
Good idea! The changes also look good to me. @mike-ferris-apple, could you ask @swift-ci to please test?

Just spitballing, but maybe we could also one day add a "glossary" or something. It took me a while to understand what the following referred to:

- Test case
- Test run
- Test suite
- Test expectation (as opposed to assertion)

ddunbar:
@swift-ci please test

mike-ferris-apple:
Thanks, Daniel.  I would definitely be in support of pushing for clarity in the Xcode docs as well, so definitely don't shy away from filing radars on that!

briancroom:
One followup Radar posted: [rdar://26002223](http://openradar.appspot.com/26002223)

PR #102, opened 2016-04-29
Add class-level `setUp` and `tearDown` methods on `XCTestCase`

briancroom:
This adds support for `setUp` and `tearDown` class methods on `XCTestCase` which Apple XCTest has, and we were still missing. These are called by a private `XCTestSuite` subclass which is configured explicitly with the test case class whose tests are in the suite.

### DIFF
--- a/Tests/Functional/TestCaseLifecycle/main.swift
+++ b/Tests/Functional/TestCaseLifecycle/main.swift
@@ -21,6 +21,12 @@ class SetUpTearDownTestCase: XCTestCase {
 
     var value = 0
 
+    // CHECK: In class setUp\(\)
+    override class func setUp() {
+        super.setUp()
+        print("In class \(#function)")
+    }
+
     override func setUp() {
         super.setUp()
         print("In \(#function)")
@@ -40,6 +46,12 @@ class SetUpTearDownTestCase: XCTestCase {
     func test_hasValueFromSetUp() {
         print("In \(#function)")
         XCTAssertEqual(value, 42)
+    }
+
+    // CHECK: In class tearDown\(\)
+    override class func tearDown() {
+        super.tearDown()
+        print("In class \(#function)")
     }
 }
 // CHECK: Test Suite 'SetUpTearDownTestCase' passed at \d+:\d+:\d+\.\d+
--- a/Sources/XCTest/XCTestCase.swift
+++ b/Sources/XCTest/XCTestCase.swift
@@ -129,6 +129,14 @@ public class XCTestCase: XCTest {
             fatalError("Terminating execution due to test failure")
         }
     }
+
+    /// Setup method called before the invocation of any test method in the
+    /// class.
+    public class func setUp() {}
+
+    /// Teardown method called after the invocation of every test method in the
+    /// class.
+    public class func tearDown() {}
 }
 
 /// Wrapper function allowing an array of static test case methods to fit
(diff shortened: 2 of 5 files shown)


Tests/Functional/TestCaseLifecycle/main.swift
```diff
@@ -21,6 +21,12 @@ class SetUpTearDownTestCase: XCTestCase {
 
     var value = 0
 
+// CHECK: In class setUp\(\)
```
modocache: With #94 merged you may now indent the `// CHECK:`, like so:

```swift
func myFunc() {
    // CHECK: foo
    print("foo")
    if (bar) {
        // CHECK: baz
        print("baz")
    }
}
```

I think indenting is more readable, so I'd encourage you to do that! (I had been meaning to start sending out pull requests to indent the checks in the tests, but forgot 😇 )

Sources/XCTest/XCTestSuite.swift
```diff
@@ -68,3 +68,33 @@ public class XCTestSuite: XCTest {
...
+            testCaseClass.setUp()
+        }
+    }
+
+    override func tearDown() {
+        if let testCaseClass = testCaseClass {
+            testCaseClass.tearDown()
+        }
+    }
+}
```
modocache: My preference would be to put this in its own file, `XCTestCaseSuite.swift`.

I'd also like to at some point separate the source files in this repository, into directories that make it clear what is public API and what is internal.

Sources/XCTest/XCTestSuite.swift
```diff
@@ -68,3 +68,33 @@ public class XCTestSuite: XCTest {
...
 }
+
+/// A test suite which is associated with a particular test case class. It will
+/// call `setUp` and `tearDown` on the class itself before and after invoking
+/// all of the test cases making up the class.
+internal class XCTestCaseSuite: XCTestSuite {
+    private let testCaseClass: XCTestCase.Type?
+
+    init(testCaseEntry: XCTestCaseEntry) {
+        let testCaseClass = testCaseEntry.0
```
modocache: nit-pick: Perhaps named attributes for this tuple, or give it a proper `struct` definition?

modocache:
Excellent! 💯

Tests/Functional/TestCaseLifecycle/main.swift
briancroom: Ooh right, thanks for the reminder!

Sources/XCTest/XCTestSuite.swift
briancroom: 👍 yeah we should totally do that.

Sources/XCTest/XCTestSuite.swift
briancroom: Well have a look at that..the tuple actually already had named attributes and I just neglected to take advantage of them here. Oops!

briancroom:
Thanks for the feedback @modocache. I pushed some tweaks now.

Would someone mind kicking off CI? (I'll also note that this will be the first build since https://github.com/apple/swift/pull/2259 was merged.) Thanks!
@mike-ferris-apple @parkera @ddunbar

mike-ferris-apple:
@swift-ci please test

briancroom:
Success 😄

PR #103, opened 2016-05-02
[XCTestExpectation] Adding expectationForPredicate constructor

yoonapps:

This pull request adds the XCTestCase convenience constructor for XCTestExpectation `expectation(forPredicate:evaluatedWithObject:handler:)`.

I added `XCPredicateExpectation`, an internal subclass of `XCTestExpectation`, to keep the evaluation logic separated (and to match the obj-c implementation). The re-evaluation interval has been shorted to `0.01` whereas the obj-c interval is about once per second.

This also includes a separate file for the `typealias` `XCPredicateExpectationHandler`.

There are tests for the expectation and the handler. Let me know if you think there should be more tests, I'd be happy to add other ideas.

https://bugs.swift.org/browse/SR-1110

This will provide a full Swift implementation of `expectationForPredicate:evaluatedWithObject:handler:` that allows asynchronous testing for `NSPredicate`.

### DIFF
--- a/Sources/XCTest/XCPredicateExpectation.swift
+++ b/Sources/XCTest/XCPredicateExpectation.swift
@@ -1,0 +1,58 @@
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2016 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//
+//  XCPredicateExpectation.swift
+//  Expectations with a specified predicate and object to evaluate.
+//
+
+#if os(Linux) || os(FreeBSD)
+    import Foundation
+#else
+    import SwiftFoundation
+#endif
+
+internal class XCPredicateExpectation: XCTestExpectation {
+    internal let predicate: NSPredicate
+    internal let object: AnyObject
+    internal var timer: NSTimer?
+    internal let handler: XCPredicateExpectationHandler?
+    private let evaluationInterval = 0.01
+    
+    internal init(predicate: NSPredicate, object: AnyObject, description: String, file: StaticString, line: UInt, testCase: XCTestCase, handler: XCPredicateExpectationHandler? = nil) {
+        self.predicate = predicate
+        self.object = object
+        self.handler = handler
+        self.timer = nil
+        super.init(description: description, file: file, line: line, testCase: testCase)
+    }
+    
+    internal func considerFulfilling() {
+        self.timer = NSTimer.scheduledTimer(self.evaluationInterval, repeats: true, fire: { [weak self] timer in
+            guard let strongSelf = self else {
+                timer.invalidate()
+                return
+            }
+            
+            if strongSelf.predicate.evaluateWithObject(strongSelf.object) {
+                if let handler = strongSelf.handler {
+                    if handler() {
+                        strongSelf.fulfill()
+                        timer.invalidate()
+                    }
+                    // The timer does not invalidate even if the handler returns
+                    // false. The object is still re-evaluated until timeout.
+                } else {
+                    strongSelf.fulfill()
+                    timer.invalidate()
+                }
+            }
+        })
+        self.timer?.fire()
+    }
+}
--- a/Sources/XCTest/XCTestCase.swift
+++ b/Sources/XCTest/XCTestCase.swift
@@ -367,5 +367,41 @@ extension XCTestCase {
         
         return expectation
     }
+    
+    /// Creates and returns an expectation that is fulfilled if the predicate 
+    /// returns true when evaluated with the given object. The expectation
+    /// periodically evaluates the predicate and also may use notifications or 
+    /// other events to optimistically re-evaluate.
+    ///
+    /// - Parameter predicate: The predicate that will be used to evaluate the 
+    ///   object.
+    /// - Parameter object: The object that is evaluated against the conditions
+    ///   specified by the predicate.
+    /// - Parameter file: The file name to use in the error message if
+    ///   this expectation is not waited for. Default is the file
+    ///   containing the call to this method. It is rare to provide this
+    ///   parameter when calling this method.
+    /// - Parameter line: The line number to use in the error message if the
+    ///   this expectation is not waited for. Default is the line
+    ///   number of the call to this method in the calling file. It is rare to
+    ///   provide this parameter when calling this method.
+    /// - Parameter handler: A block to be invoked when evaluating the predicate
+    ///   against the object returns true. If the block is not provided the 
+    ///   first successful evaluation will fulfill the expectation. If provided,
+    ///   the handler can override that behavior which leaves the caller
+    ///   responsible for fulfilling the expectation.
+    public func expectation(for predicate: NSPredicate, evaluatedWith object: AnyObject, file: StaticString = #file, line: UInt = #line, handler: XCPredicateExpectationHandler? = nil) -> XCTestExpectation {
+        let expectation = XCPredicateExpectation(
+            predicate: predicate,
+            object: object,
+            description: "Expect `\(predicate)` for object \(object)",
+            file: file,
+            line: line,
+            testCase: self,
+            handler: handler)
+        _allExpectations.append(expectation)
+        expectation.considerFulfilling()
+        return expectation
+    }
 
 }
--- a/Tests/Functional/Asynchronous/Predicates/Expectations/main.swift
+++ b/Tests/Functional/Asynchronous/Predicates/Expectations/main.swift
@@ -1,0 +1,83 @@
+// RUN: %{swiftc} %s -o %{built_tests_dir}/Asynchronous-Predicates
+// RUN: %{built_tests_dir}/Asynchronous-Predicates > %t || true
+// RUN: %{xctest_checker} %t %s
+
+#if os(Linux) || os(FreeBSD)
+    import XCTest
+    import Foundation
+#else
+    import SwiftXCTest
+    import SwiftFoundation
+#endif
+
+// CHECK: Test Suite 'All tests' started at \d+:\d+:\d+\.\d+
+// CHECK: Test Suite '.*\.xctest' started at \d+:\d+:\d+\.\d+
+
+// CHECK: Test Suite 'PredicateExpectationsTestCase' started at \d+:\d+:\d+\.\d+
+class PredicateExpectationsTestCase: XCTestCase {
+    // CHECK: Test Case 'PredicateExpectationsTestCase.test_immediatelyTruePredicateAndObject_passes' started at \d+:\d+:\d+\.\d+
+    // CHECK: Test Case 'PredicateExpectationsTestCase.test_immediatelyTruePredicateAndObject_passes' passed \(\d+\.\d+ seconds\).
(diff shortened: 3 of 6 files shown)


Tests/Functional/Asynchronous/Predicates/Expectations/main.swift
```diff
@@ -0,0 +1,83 @@
...
+    import SwiftFoundation
+#endif
+
+// CHECK: Test Suite 'All tests' started at \d+:\d+:\d+\.\d+
+// CHECK: Test Suite '.*\.xctest' started at \d+:\d+:\d+\.\d+
+
+// CHECK: Test Suite 'PredicateExpectationsTestCase' started at \d+:\d+:\d+\.\d+
+class PredicateExpectationsTestCase: XCTestCase {
+// CHECK: Test Case 'PredicateExpectationsTestCase.test_immediatelyTruePredicateAndObject_passes' started at \d+:\d+:\d+\.\d+
+// CHECK: Test Case 'PredicateExpectationsTestCase.test_immediatelyTruePredicateAndObject_passes' passed \(\d+\.\d+ seconds\).
```
modocache: With #94 merged you may now indent the `// CHECK:`, like so:

```swift
func myFunc() {
    // CHECK: foo
    print("foo")
    if (bar) {
        // CHECK: baz
        print("baz")
    }
}
```

I think indenting is more readable, so I'd encourage you to do that! (I had been meaning to start sending out pull requests to indent the checks in the tests, but forgot 😇 )

Tests/Functional/Asynchronous/Predicates/Expectations/main.swift
```diff
@@ -0,0 +1,83 @@
...
+    func test_immediatelyTruePredicateAndObject_passes() {
+        let predicate = NSPredicate(value: true)
+        let object = NSObject()
+        expectation(forPredicate:predicate, evaluatedWithObject: object)
+        waitForExpectations(withTimeout: 0.1)
+    }
+
+// CHECK: Test Case 'PredicateExpectationsTestCase.test_immediatelyFalsePredicateAndObject_fails' started at \d+:\d+:\d+\.\d+
+// CHECK: .*/Tests/Functional/Asynchronous/Predicates/Expectations/main.swift:34: error: PredicateExpectationsTestCase.test_immediatelyFalsePredicateAndObject_fails : Asynchronous wait failed - Exceeded timeout of 0.1 seconds, with unfulfilled expectations: Expect `<NSPredicate: 0x[0-9A-Fa-f]{1,16}>` for object <NSObject: 0x[0-9A-Fa-f]{1,16}>
```
modocache: Thanks for using the correct line number, `34`, here. I think verifying these appear on the correct line is a good thing, but I know it's painful to keep these expected line numbers in sync with the actual ones that are output. I'm planning on adding a `[[@LINE+<offset>]]` feature soon (originally implemented in #68)--hang tight! 😄

Tests/Functional/Asynchronous/Predicates/Expectations/main.swift
```diff
@@ -0,0 +1,83 @@
...
+        let halfSecLaterDate = NSDate(timeIntervalSinceNow: 2.0)
+        let predicate = NSPredicate(block: {
+            evaluatedObject, bindings in
+            if let evaluatedDate = evaluatedObject as? NSDate {
+                return evaluatedDate.compare(NSDate()) == NSComparisonResult.OrderedAscending
+            }
+            return false
+        })
+        expectation(forPredicate:predicate, evaluatedWithObject: halfSecLaterDate)
+        waitForExpectations(withTimeout: 3.0)
```
modocache: A test that passes after two seconds is a little too slow for my tastes. Can we change these to `0.1` and `0.2` seconds instead?

Sources/XCTest/XCPredicateExpectation.swift
```diff
@@ -0,0 +1,53 @@
...
+#else
+    import SwiftFoundation
+#endif
+
+public class XCPredicateExpectation: XCTestExpectation {
+    internal let predicate: NSPredicate
+    internal let object: AnyObject
+    internal var timer: NSTimer?
+    internal let handler: XCPredicateExpectationHandler?
+    private let evaluationInterval = 1.0
```
modocache: An evaluation interval of one second seems super slow to me... is similar to the evaluation interval for Apple XCTest? I would mimic Apple XCTest, but if I had to choose an interval I would choose something closer to `0.1` second, or maybe even a shorter interval than that!

modocache:
This is really excellent work, @yoonapps, thank you!!

@mike-ferris-apple Could you ask @swift-ci to please test?

Tests/Functional/Asynchronous/Predicates/Expectations/main.swift
yoonapps: The timer in the expectation re-evaluates every second which is why I gave it some time to succeed.
But I would like to reduce the time as well. I could lower the re-evaluation interval when testing, maybe based on a flag? Or just lower the interval in general?

modocache: See my other comment: https://github.com/apple/swift-corelibs-xctest/pull/103#discussion_r61815739 -- I think you should lower the re-evaluation interval, not just when testing but all the time. After all, `XCTestCase` itself currently [re-evaluates whether expectations are met every `0.01` seconds](https://github.com/apple/swift-corelibs-xctest/blob/e3b502b78e1c31f44ed2f9542703648838d9ff99/Sources/XCTest/XCTestCase.swift#L287).

yoonapps:
@modocache Addressed reviews! I lowered the interval to `0.01` and updated the tests to wait for only `0.1`.

ddunbar:
@swift-ci please test

Sources/XCTest/XCTestCase.swift
```diff
@@ -367,5 +367,43 @@ extension XCTestCase {
...
+    /// - Parameter line: The line number to use in the error message if the
+    ///   this expectation is not waited for. Default is the line
+    ///   number of the call to this method in the calling file. It is rare to
+    ///   provide this parameter when calling this method.
+    /// - Parameter handler: A block to be invoked when evaluating the predicate
+    ///   against the object returns true. If the block is not provided the 
+    ///   first successful evaluation will fulfill the expectation. If provided,
+    ///   the handler can override that behavior which leaves the caller
+    ///   responsible for fulfilling the expectation.
+    public func expectation(forPredicate predicate: NSPredicate, evaluatedWithObject object: AnyObject, file: StaticString = #file, line: UInt = #line, handler: XCPredicateExpectationHandler? = nil) -> XCPredicateExpectation {
```
briancroom: I think the precise signature of this method needs to be adjusted. I'm seeing the following imported signature in Xcode using the latest Swift toolchain:
```swift
public func expectation(for predicate: NSPredicate, evaluatedWith object: AnyObject, handler: XCPredicateExpectationHandler? = nil) -> XCTestExpectation
```
Does this look right? @mike-ferris-apple @modocache

Sources/XCTest/XCPredicateExpectation.swift
```diff
@@ -0,0 +1,53 @@
...
+    internal init(predicate: NSPredicate, objectToEvaluate object: AnyObject, description: String, file: StaticString, line: UInt, testCase: XCTestCase, handler: XCPredicateExpectationHandler? = nil) {
+        self.predicate = predicate
+        self.object = object
+        self.handler = handler
+        self.timer = nil
+        super.init(description: description, file: file, line: line, testCase: testCase)
+    }
+    
+    internal func considerFulfilling() {
+        self.timer = NSTimer.scheduledTimer(self.evaluationInterval, repeats: true, fire: { timer in
```
briancroom: Maybe I'm missing something, but I don't see how this polling timer would get invalidated when the expectation times out. This looks like a similar issue to the one addressed in https://github.com/apple/swift-corelibs-xctest/pull/90

briancroom:
Excited to see this come in @yoonapps, thanks! I just left a couple questions.

Sources/XCTest/XCTestCase.swift
modocache: Agreed. [This](https://github.com/modocache/XCTest.swift/blob/5b5a152150bacea9c373dcfb16a8384563cd4419/XCTest.swift#L85) was generated with the Swift 3 importer from a month ago, but it matches @briancroom's signature.

yoonapps:
I'm learning so much every PR! It's awesome. 
@briancroom Thank you for fixing strong reference issue for notification! I applied a similar fix here. I use `weakSelf` for networking a whole bunch at work, such a "duh" moment. :laughing:
One question, I was going to fix the signature for `init` of `XCPredicateExpectation` as well but it looked like `init`s don't follow the same rules. Is that right?

Sources/XCTest/XCPredicateExpectation.swift
```diff
@@ -0,0 +1,58 @@
...
+                        timer.invalidate()
+                    }
+                    // The timer does not invalidate even if the handler returns
+                    // false. The object is still re-evaluated until timeout.
+                } else {
+                    strongSelf.fulfill()
+                    timer.invalidate()
+                }
+            }
+            })
```
yoonapps: I've tried CTRL + I a bunch here but this is where the indentation goes. Does this look funky or is it just me?

modocache:
> One question, I was going to fix the signature for init of XCPredicateExpectation as well but it looked like inits don't follow the same rules. 

Good observation! Apple XCTest doesn't provide a public initializer for `XCTestExpectation`, nor does it provide a public definition of any of its private subclasses (such as `_XCPredicateExpectation`). So we don't need to worry about providing an identical signature: no users of Apple XCTest should ever initialize an `XCTestExpectation` directly.

Sources/XCTest/XCPredicateExpectation.swift
modocache: Yes, it does look strange. Personally, I would remove four spaces, in order to align it with the open brace on the `self.timer = NSTimer.scheduledTimer(self.evaluationInterval, repeats: true, fire: {` line. It's up to you, though! :)

yoonapps:
Gotcha! Is there anything else that needs to be addresed?

modocache:
@briancroom does the timer invalidation look good?

@mike-ferris-apple Could you ask @swift-ci to please test?

ddunbar:
@swift-ci please test

briancroom:
Looks great! Nice work @yoonapps

Sources/XCTest/XCPredicateExpectation.swift
```diff
@@ -0,0 +1,58 @@
...
+//  Expectations with a specified predicate and object to evaluate.
+//
+
+#if os(Linux) || os(FreeBSD)
+    import Foundation
+#else
+    import SwiftFoundation
+#endif
+
+public class XCPredicateExpectation: XCTestExpectation {
```
briancroom: Actually, just one more thing - can we make this subclass have internal visibility, and return the base class type from the factory method on XCTestCase?

yoonapps:
@briancroom Ah!! I had it `internal` initially but changed to `public` because of the factory method return type. I didn't realize that I could just return the base type! :scream::boom: Mind blown.

modocache:
Excellent, thanks @yoonapps!! That's two tasks under your belt! 💯

yoonapps:
Not a problem. The experience I'm getting from these tasks are invaluable. Thanks for sending this task my way, @modocache! I'm always ready for more so let me know if you have any suggestions.

PR #105, opened 2016-05-09
Fix XCTestCase typos

ultramiraculous:
Fixing some misc typos I came across looking through this code.


briancroom:
Thanks @ultramiraculous!

PR #106, opened 2016-05-10
Add an option to specify the install prefix for CoreFoundation depend…

bryanpkc:
…encies.

Currently, swift-corelibs-xctest will fail to build if build-script option `--install-prefix` is set to use a location other than `/usr`. This is because build_script.py assumes that the CoreFoundation libraries can always be found under `$foundation_build_dir/usr/lib/swift`. This patch adds a new option to the Python script, to allow the top-level build-script to pass the correct prefix to the swift-corelibs-xctest build.

### DIFF
--- a/build_script.py
+++ b/build_script.py
@@ -114,7 +114,7 @@ class GenericUnixStrategy:
         build_dir = os.path.abspath(args.build_dir)
         foundation_build_dir = os.path.abspath(args.foundation_build_dir)
         core_foundation_build_dir = GenericUnixStrategy.core_foundation_build_dir(
-            foundation_build_dir)
+            foundation_build_dir, args.foundation_install_prefix)
         if args.libdispatch_build_dir:
             libdispatch_build_dir = os.path.abspath(args.libdispatch_build_dir)
         if args.libdispatch_src_dir:
@@ -201,7 +201,7 @@ class GenericUnixStrategy:
         tests_path = os.path.join(SOURCE_DIR, "Tests", "Functional")
         foundation_build_dir = os.path.abspath(args.foundation_build_dir)
         core_foundation_build_dir = GenericUnixStrategy.core_foundation_build_dir(
-            foundation_build_dir)
+            foundation_build_dir, args.foundation_install_prefix)
 
         run('SWIFT_EXEC={swiftc} '
             'BUILT_PRODUCTS_DIR={built_products_dir} '
@@ -246,7 +246,7 @@ class GenericUnixStrategy:
             os.path.join(module_install_path, xctest_swiftdoc)))
 
     @staticmethod
-    def core_foundation_build_dir(foundation_build_dir):
+    def core_foundation_build_dir(foundation_build_dir, foundation_install_prefix):
         """
         Given the path to a swift-corelibs-foundation built product directory,
         return the path to CoreFoundation built products.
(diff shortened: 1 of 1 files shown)


briancroom:
Cool, this seems to make sense. I find it a little strange that the *installation* prefix affects the layout of Foundation's *build products* directory, but I see that it does indeed work that way. 

Could you rebase this please? Sorry, I merged another PR which also had build script changes.

build_script.py
```diff
@@ -307,13 +308,19 @@ def main(args=sys.argv[1:]):
...
     build_parser.add_argument(
         "--foundation-build-dir",
         help="Path to swift-corelibs-foundation build products, which "
              "the built XCTest.so will be linked against.",
         required=strategy.requires_foundation_build_dir())
+    build_parser.add_argument(
+        "--foundation-install-prefix",
+        help="Path to the installation location for swift-corelibs-foundation "
+             "build products; CoreFoundation dependencies are expected to be "
+             "found under $foundation_build_dir/$foundation_install_prefix.",
```
modocache: I think it'd be nice to include in the description what the default value is. For example, the `--lit` argument does this: `"'%(default)s' by default."`.

build_script.py
```diff
@@ -373,6 +380,12 @@ def main(args=sys.argv[1:]):
         help="Path to swift-corelibs-foundation build products, which the "
              "tests will be linked against.",
         required=strategy.requires_foundation_build_dir())
+    test_parser.add_argument(
+        "--foundation-install-prefix",
+        help="Path to the installation location for swift-corelibs-foundation "
+             "build products; CoreFoundation dependencies are expected to be "
+             "found under $foundation_build_dir/$foundation_install_prefix.",
```
modocache: Same as above, I think it'd be nice to include what the `%(default)s` value is in this description.

bryanpkc:
@briancroom Thanks, I have rebased the patch.

@modocache Thank you for the suggestion. I have updated the help message to make it indicate the default value and look more consistent with the other messages.

briancroom:
@swift-ci please test

briancroom:
@swift-ci please test

briancroom:
@swift-ci please test

briancroom:
@swift-ci please test

briancroom:
CI seems unwilling to accept a request from me for this PR for whatever reason. The changes LGTM, but I'm unable to try it out myself right now and don't want to hit merge without seeing it actually execute. I'll try to give it a go tomorrow unless @modocache or someone else do so first.

modocache:
@swift-ci please test

modocache:
Swift CI appears to be a very fickle creature. It was listening to me just [two hours ago](https://github.com/apple/swift-corelibs-xctest/pull/112#issuecomment-219334610)...

modocache:
@swift-ci please test

modocache:
So strange! It listened to me [just a minute ago](https://github.com/apple/swift-corelibs-xctest/pull/113#issuecomment-219344857). @shahmishal, do you have any idea why @swift-ci is choosing when to obey orders?

inline_space_odyssey

Seriously, though, perhaps it has something to do with *who* submitted the pull request? I've noticed "@swift-ci please test" works for my own pull requests, but not this one.

bryanpkc:
Maybe it has to do with the repo being owned by an organization rather than a personal account? If that's the case, I can move the branch to a personal repo. But that would be sad because I have a bunch of other pull requests in the same boat.

briancroom:
Ok this checks out locally for me. I'm going to pull the trigger on it now.

modocache:
Great, thanks!

I only just noticed this is related to the https://github.com/apple/swift/pull/2541. Great work!! 💯

bryanpkc:
Thanks @briancroom!

PR #107, opened 2016-05-10
Configure use of Dispatch on Linux if included in Foundation

seabaylea:
Its currently possible to optionally build Foundation with access to Dispatch for NSRunLoop, NSURLSession, NSOperation etc.

Once there's a dependency between Foundation and Dispatch, there's build failures for XCTest as it doesn't have access to Dispatch build artefacts.

This updates the XCTest build script to add that configuration if Dispatch has been added to the Foundation builds. 

Note that there's an equivalent PR against Swift to pass in the libdispatch source and build locations as arguments, if Dispatch is included in Foundation.


briancroom:
@swift-ci please test

briancroom:
This LGTM. Thanks @seabaylea!

PR #108, opened 2016-05-12
Adjusted usage of NSComparisonResult cases to Darwin API

tomquist:
This is a follow-up PR for [#356 Changed API of comparison/sort/enumeration/qos options to match Darwin version](https://github.com/apple/swift-corelibs-foundation/pull/356) in swift-corelibs-foundation


briancroom:
Thanks @tomquist! This should be good to go once the Foundation PR goes in 👍

parkera:
I'll merge both.

PR #109, opened 2016-05-12
[SR-1355] Add Performance Measurement APIs

briancroom:
# What's in this pull request
This is an initial implementation of the XCTest performance measurement APIs. All public facing APIs from Apple XCTest are present, but not all behaviors are implemented at this point. It is currently possible to measure the wall clock time that a given block of code takes to execute, and see the measured times, the mean value, and the relative standard deviation in the printed test output.

The core behaviors mimic those described in the Apple XCTest framework headers, supplemented by information presented in the [Testing in Xcode 6 WWDC session](https://developer.apple.com/videos/play/wwdc2014/414/).

* The code being measured is always executed exactly 10 times (assuming no test failures occur)
* A test failure is triggered if the relative standard deviation of the times measured is greater than 10%, and the standard deviation value is greater than 0.1

This does not include any support for recording baseline values or triggering test failures due to a regression relative to a baseline.

Resolves [SR-1355](https://bugs.swift.org/browse/SR-1355).

# Code structure
Although the build system does not currently facilitate the use of multiple Swift modules, I have structured the code with modularity in mind. The `PerformanceMeter.swift` and `WallClockTimeMetric.swift` contain the heart of the implementation and have (almost) no dependence on other parts of the XCTest library. `XCTestCase+Performance.swift` contains the public APIs which delegate to the `PerformanceMeter` class.

# Test coverage
There are extensive functional tests covering of API misuse scenarios and basic measurement output, however many finer points (e.g. standard deviation calculation and situations triggering failure) have only cursory tests due to the lack of a usable unit testing infrastructure (see [this PR](https://github.com/apple/swift-corelibs-xctest/pull/61#issuecomment-207822336) for previous discussion about this.)

I attempted to make the tests robust, however due to the time-sensitive nature of the APIs under test, there is an increased possibility of flakiness that could appear when running in a CI environment. I would greatly appreciate feedback on this aspect of the test cases and what approaches we could take to minimize the risk of false negative results. I would also like to hear feedback on the `test_measuresWallClockTimeInBlock` test case in `Performance/main.swift`, which currently includes a 1-second sleep in order to 
test that time passing is being measured.

# Error reporting
I took pains to implement failure reporting due to API misuse, just as Apple XCTest does. One element of this is reporting the file and line number where the misuse occurred. I have taken cues from @modocache's `XCTestExpectation` implementation and added `file` and `line` params where necessary to capture that information. The current implementation records test failures for API misuse, allowing test suite execution to continue at the cost of some additional bookkeeping code. An alternative would be to simply `fatalError` in these scenarios, since they are clearly programmer error. I would appreciate others' input on these options.

# Time measurement
I am no expert in benchmarking, so I am not especially confident in the optimal mechanism for measuring the passage of time for this purpose, however after some cursory research, I settled on using `NSProcessInfo`'s `systemUptime` property, which ultimately calls down to `mach_absolute_time` on Darwin, and `clock_gettime` with `CLOCK_MONOTONIC` on Linux. As far as I can tell, these are the appropriate primitives to use for this purpose in these environments. One potential source of error is introduced in that the time values are being converted to `Double`, but I'm unsure of the practical impact of that.

--
While not complete, this is still a significant step towards feature parity with Apple XCTest, and should be sufficient to start gathering feedback on future implementation directions. In particular, it should put us in a good position to productively carry on discussions like [this one](https://lists.swift.org/pipermail/swift-corelibs-dev/Week-of-Mon-20151207/000115.html) from @drewcrawford which have great potential to improve the value that these tools provide.

cc @ddunbar @paulofaria

### DIFF
--- a/Sources/XCTest/PerformanceMeter.swift
+++ b/Sources/XCTest/PerformanceMeter.swift
@@ -1,0 +1,190 @@
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2016 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//
+//  PerformanceMeter.swift
+//  Measures the performance of a block of code and reports the results.
+//
+
+#if os(Linux) || os(FreeBSD)
+    import Foundation
+#else
+    import SwiftFoundation
+#endif
+
+/// Describes a type that is capable of measuring some aspect of code performance
+/// over time.
+internal protocol PerformanceMetric {
+    /// Called once per iteration immediately before the tested code is executed. 
+    /// The metric should do whatever work is required to begin a new measurement.
+    func startMeasuring()
+
+    /// Called once per iteration immediately after the tested code is executed.
+    /// The metric should do whatever work is required to finalize measurement.
+    func stopMeasuring()
+
+    /// Called once, after all measurements have been taken, to provide feedback
+    /// about the collected measurements.
+    /// - Returns: Measurement results to present to the user.
+    func calculateResults() -> String
+
+    /// Called once, after all measurements have been taken, to determine whether
+    /// the measurements should be treated as a test failure or not.
+    /// - Returns: A diagnostic message if the results indicate failure, else nil.
+    func failureMessage() -> String?
+}
+
+internal protocol PerformanceMeterDelegate {
+    func recordMeasurements(results: String, file: StaticString, line: UInt)
+    func recordFailure(description: String, file: StaticString, line: UInt)
+    func recordAPIViolation(description: String, file: StaticString, line: UInt)
+}
+
+/// - Bug: This class is intended to be `internal` but is public to work around
+/// a toolchain bug on Linux. See `XCTestCase._performanceMeter` for more info.
+public final class PerformanceMeter {
+    enum Error: ErrorProtocol, CustomStringConvertible {
+        case noMetrics
+        case unknownMetric(metricName: String)
+        case startMeasuringAlreadyCalled
+        case stopMeasuringAlreadyCalled
+        case startMeasuringNotCalled
+        case stopBeforeStarting
+
+        var description: String {
+            switch self {
+            case .noMetrics: return "At least one metric must be provided to measure."
+            case .unknownMetric(let name): return "Unknown metric: \(name)"
+            case .startMeasuringAlreadyCalled: return "Already called startMeasuring() once this iteration."
+            case .stopMeasuringAlreadyCalled: return "Already called stopMeasuring() once this iteration."
+            case .startMeasuringNotCalled: return "startMeasuring() must be called during the block."
+            case .stopBeforeStarting: return "Cannot stop measuring before starting measuring."
+            }
+        }
+    }
+
+    internal var didFinishMeasuring: Bool {
+        return state == .measurementFinished || state == .measurementAborted
+    }
+
+    private enum State {
+        case iterationUnstarted
+        case iterationStarted
+        case iterationFinished
+        case measurementFinished
+        case measurementAborted
+    }
+    private var state: State = .iterationUnstarted
+
+    private let metrics: [PerformanceMetric]
+    private let delegate: PerformanceMeterDelegate
+    private let invocationFile: StaticString
+    private let invocationLine: UInt
+
+    private init(metrics: [PerformanceMetric], delegate: PerformanceMeterDelegate, file: StaticString, line: UInt) {
+        self.metrics = metrics
+        self.delegate = delegate
+        self.invocationFile = file
+        self.invocationLine = line
+    }
+
+    static func measureMetrics(_ metricNames: [String], delegate: PerformanceMeterDelegate, file: StaticString = #file, line: UInt = #line, for block: @noescape (PerformanceMeter) -> Void) {
+        do {
+            let metrics = try self.metrics(forNames: metricNames)
+            let meter = PerformanceMeter(metrics: metrics, delegate: delegate, file: file, line: line)
+            meter.measure(block)
+        } catch let e {
+            delegate.recordAPIViolation(description: String(e), file: file, line: line)
+        }
+    }
+
+    func startMeasuring(file: StaticString = #file, line: UInt = #line) {
+        guard state == .iterationUnstarted else {
+            state = .measurementAborted
+            return recordAPIViolation(.startMeasuringAlreadyCalled, file: file, line: line)
+        }
+        state = .iterationStarted
+        metrics.forEach { $0.startMeasuring() }
+    }
+
+    func stopMeasuring(file: StaticString = #file, line: UInt = #line) {
+        guard state != .iterationUnstarted else {
+            return recordAPIViolation(.stopBeforeStarting, file: file, line: line)
+        }
+
+        guard state != .iterationFinished else {
+            return recordAPIViolation(.stopMeasuringAlreadyCalled, file: file, line: line)
+        }
+
+        state = .iterationFinished
+        metrics.forEach { $0.stopMeasuring() }
+    }
+
+    func abortMeasuring() {
+        state = .measurementAborted
+    }
+
+
+    private static func metrics(forNames names: [String]) throws -> [PerformanceMetric] {
+        guard !names.isEmpty else { throw Error.noMetrics }
+
+        let metricsMapping = [WallClockTimeMetric.name : WallClockTimeMetric.self]
+
+        return try names.map({
+            guard let metricType = metricsMapping[$0] else { throw Error.unknownMetric(metricName: $0) }
+            return metricType.init()
+        })
+    }
+
+    private var numberOfIterations: Int {
+        return 10
+    }
+
+    private func measure(_ block: @noescape (PerformanceMeter) -> Void) {
+        for _ in (0..<numberOfIterations) {
+            state = .iterationUnstarted
+
+            block(self)
+            stopMeasuringIfNeeded()
+
+            if state == .measurementAborted { return }
+
+            if state == .iterationUnstarted {
+                recordAPIViolation(.startMeasuringNotCalled, file: invocationFile, line: invocationLine)
+                return
+            }
+        }
+        state = .measurementFinished
+
+        recordResults()
+        recordFailures()
+    }
+
+    private func stopMeasuringIfNeeded() {
+        if state == .iterationStarted {
+            stopMeasuring()
+        }
+    }
+
+    private func recordResults() {
+        for metric in metrics {
+            delegate.recordMeasurements(results: metric.calculateResults(), file: invocationFile, line: invocationLine)
+        }
+    }
+
+    private func recordFailures() {
+        metrics.flatMap({ $0.failureMessage() }).forEach { message in
+            delegate.recordFailure(description: message, file: invocationFile, line: invocationLine)
+        }
+    }
+
+    private func recordAPIViolation(_ error: Error, file: StaticString, line: UInt) {
+        state = .measurementAborted
+        delegate.recordAPIViolation(description: String(error), file: file, line: line)
+    }
+}
--- a/Sources/XCTest/WallClockTimeMetric.swift
+++ b/Sources/XCTest/WallClockTimeMetric.swift
@@ -1,0 +1,86 @@
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2016 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//
+//  WallClockTimeMetric.swift
+//  Performance metric measuring how long it takes code to execute
+//
+
+#if os(Linux) || os(FreeBSD)
+    import Foundation
+#else
+    import SwiftFoundation
+#endif
+
+/// This metric uses the system uptime to keep track of how much time passes
+/// between starting and stopping measuring.
+internal final class WallClockTimeMetric: PerformanceMetric {
+    static let name = "org.swift.XCTPerformanceMetric_WallClockTime"
+
+    typealias Measurement = NSTimeInterval
+    private var startTime: NSTimeInterval?
+    var measurements: [Measurement] = []
+
+    func startMeasuring() {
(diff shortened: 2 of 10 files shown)


briancroom:
@swift-ci please test

modocache:
Amazing!! Thanks for this. Swift 3.0, here we come!!

> Although the build system does not currently facilitate the use of multiple Swift modules, I have structured the code with modularity in mind.

Is this something you think we should address? Are there consequences for end-users if we split it up, such as requiring them to link multiple modules, besides XCTest?

> `XCTestCase+Performance.swift` contains the public APIs which delegate to the `PerformanceMeter` class.

This has been in the back of my mind for a while now: Apple places its asynchronous testing APIs in a category on `XCTestCase`, placed in a file called `XCTestCase+AsynchronousTesting.h`. Do you think we should split the asynchronous testing portion of `SwiftXCTest.XCTestCase` into an extension in its own file? Following your naming convention, perhaps `XCTestCase+Asynchronous.swift`? Or is this not worth the effort? One tricky part is that `XCTestCase` uses an internal variable to hold onto the expectations it's vended, whereas we can't place additional variables on extensions...

More granular comments to follow.

Sources/XCTest/PerformanceMeter.swift
```diff
@@ -0,0 +1,190 @@
...
+    /// the measurements should be treated as a test failure or not.
+    /// - Returns: A diagnostic message if the results indicate failure, else nil.
+    func failureMessage() -> String?
+}
+
+internal protocol PerformanceMeterDelegate {
+    func recordMeasurements(results: String, file: StaticString, line: UInt)
+    func recordFailure(description: String, file: StaticString, line: UInt)
+    func recordAPIViolation(description: String, file: StaticString, line: UInt)
+}
```
modocache: The awesome documentation for `PerformanceMetric` spoiled me, now I wish there was documentation here, too 😝

Sources/XCTest/PerformanceMeter.swift
```diff
@@ -0,0 +1,190 @@
...
+
+/// - Bug: This class is intended to be `internal` but is public to work around
+/// a toolchain bug on Linux. See `XCTestCase._performanceMeter` for more info.
```
modocache: Dang, this bug still hasn't been fixed, huh? :( I commented on [SR-272](https://bugs.swift.org/browse/SR-272), that report appears to have more subscribers than [SR-1129](https://bugs.swift.org/browse/SR-1129).

Sources/XCTest/XCTestObservation.swift
```diff
@@ -77,3 +77,13 @@ public extension XCTestObservation {
     func testSuiteDidFinish(_ testSuite: XCTestSuite) {}
     func testBundleDidFinish(_ testBundle: NSBundle) {}
 }
+
+/// Expanded version of `XCTestObservation` used internally to respond to
+/// additional events not publicly exposed.
+internal protocol _XCTestObservation: XCTestObservation {
```
modocache: Cool! Does this deserve its own file? 🤔

Tests/Functional/Performance/Misuse/main.swift
```diff
@@ -0,0 +1,135 @@
...
+
+// CHECK: Test Suite 'All tests' started at \d+:\d+:\d+\.\d+
+// CHECK: Test Suite '.*\.xctest' started at \d+:\d+:\d+\.\d+
+
+// CHECK: Test Suite 'PerformanceMisuseTestCase' started at \d+:\d+:\d+\.\d+
+class PerformanceMisuseTestCase: XCTestCase {
+
+    // CHECK: Test Case 'PerformanceMisuseTestCase.test_whenMeasuringMultipleInOneTest_fails' started at \d+:\d+:\d+\.\d+
+    func test_whenMeasuringMultipleInOneTest_fails() {
+        // CHECK: .*/Misuse/main.swift:20: Test Case 'PerformanceMisuseTestCase.test_whenMeasuringMultipleInOneTest_fails' measured.*
```
modocache: Gah, I'm so, so sorry you had to hardcode these without line offsets... 😭  I really need to revive that pull request.

Anyway, thanks for checking the exact lines! I like the fact that these ensure we emit failures on the correct lines.

Tests/Functional/Performance/Misuse/main.swift
```diff
@@ -0,0 +1,135 @@
...
+        }
+    }
+    // CHECK: Test Case 'PerformanceMisuseTestCase.test_whenMeasuringMetricsAndStartingTwice_fails' failed \(\d+\.\d+ seconds\).
+
+    // CHECK: Test Case 'PerformanceMisuseTestCase.test_whenMeasuringMetricsAndStoppingTwice_fails' started at \d+:\d+:\d+\.\d+
+    func test_whenMeasuringMetricsAndStoppingTwice_fails() {
+        measureMetrics([XCTPerformanceMetric_WallClockTime], automaticallyStartMeasuring: false) {
+            self.startMeasuring()
+            self.stopMeasuring()
+            // CHECK: .*/Misuse/main.swift:60: error: PerformanceMisuseTestCase.test_whenMeasuringMetricsAndStoppingTwice_fails : API violation - Already called stopMeasuring\(\) once this iteration.
```
modocache: I'm loving these comprehensive tests! 😍

Tests/Functional/Performance/Misuse/main.swift
```diff
@@ -0,0 +1,135 @@
...
+            self.stopMeasuring()
+        }
+    }
+    // CHECK: Test Case 'PerformanceMisuseTestCase.test_whenMeasuringMetricsAndStoppingTwice_fails' failed \(\d+\.\d+ seconds\).
+
+    // CHECK: Test Case 'PerformanceMisuseTestCase.test_startMeasuringOutsideOfBlock_fails' started at \d+:\d+:\d+\.\d+
+    func test_startMeasuringOutsideOfBlock_fails() {
+        // CHECK: .*/Misuse/main.swift:68: error: PerformanceMisuseTestCase.test_startMeasuringOutsideOfBlock_fails : API violation - Cannot start measuring. startMeasuring\(\) is only supported from a block passed to measureMetrics\(...\).
```
modocache: I'm assuming these error messages are copied verbatim from the Apple XCTest output, is that correct?

Tests/Functional/Performance/main.swift
```diff
@@ -0,0 +1,120 @@
...
+    // CHECK: Test Case 'PerformanceTestCase.test_abortsMeasurementsAfterTestFailure' failed \(\d+\.\d+ seconds\).
+
+    // CHECK: Test Case 'PerformanceTestCase.test_measuresWallClockTimeInBlock' started at \d+:\d+:\d+\.\d+
+    func test_measuresWallClockTimeInBlock() {
+        var hasWaited = false
+        // CHECK: .*/Performance/main.swift:90: Test Case 'PerformanceTestCase.test_measuresWallClockTimeInBlock' measured \[Time, seconds\] average: \d+.\d{3}, relative standard deviation: \d+.\d{3}%, values: \[1.\d{6}, 0.\d{6}, 0.\d{6}, 0.\d{6}, 0.\d{6}, 0.\d{6}, 0.\d{6}, 0.\d{6}, 0.\d{6}, 0.\d{6}\], performanceMetricID:org.swift.XCTPerformanceMetric_WallClockTime, maxPercentRelativeStandardDeviation: \d+.\d{3}%, maxStandardDeviation: \d.\d{3}
+        // CHECK: .*/Performance/main.swift:90: error: PerformanceTestCase.test_measuresWallClockTimeInBlock : failed: The relative standard deviation of the measurements is \d+.\d{3}% which is higher than the max allowed of \d+.\d{3}%.
+        measure {
+            if !hasWaited {
+                NSThread.sleepForTimeInterval(1)
```
modocache: While they don't exactly sleep, several of the tests for the asynchronous testing feature set end up spinning the CPU for anywhere between `0.1` and `0.2` seconds. So I'm not opposed to this sleep being here for the purposes of inducing a standard deviation threshold failure. Could it be shorter, though? Or do we need a full second?

Sources/XCTest/WallClockTimeMetric.swift
```diff
@@ -0,0 +1,86 @@
...
+                          "values: [\(measurements.map({ String(format: "%.6f", $0) }).joined(separator: ", "))]",
+                          "performanceMetricID:\(self.dynamicType.name)",
+                          String(format: "maxPercentRelativeStandardDeviation: %.3f%%", maxRelativeStandardDeviation),
+                          String(format: "maxStandardDeviation: %.3f", standardDeviationNegligibilityThreshold),
+                          ]
+        return "[Time, seconds] \(results.joined(separator: ", "))"
+    }
+
+    func failureMessage() -> String? {
+        // The relative standard deviation of the measurements is \d+.\d{3}% which is higher than the max allowed of \d+.\d{3}%.
```
modocache: nit-pick: I feel like this comment doesn't add much--my eyes only need to scan a few lines down to see this method either returns a standard deviation failure, or no failure at all. Just my two cents.

Sources/XCTest/WallClockTimeMetric.swift
```diff
@@ -0,0 +1,86 @@
...
+    var average: WallClockTimeMetric.Measurement {
+        return self.reduce(0, combine: +) / Double(count.toIntMax())
+    }
+
+    var standardDeviation: WallClockTimeMetric.Measurement {
+        let average = self.average
+        let squaredDifferences = self.map({ pow($0 - average, 2.0) })
+        let variance = squaredDifferences.reduce(0, combine: +) / Double(count.toIntMax()-1)
+        return sqrt(variance)
+    }
```
modocache: Calculating the standard deviation almost seems like something that could be in the stdlib. Or, once we get unit tests set up thanks to #61, the logic here could be extracted into its own function and unit tested.

modocache: Sorry, that wasn't super clear: this is a nit-pick. Prior to actually having unit tests, I don't feel strongly about whether this should be more generic than it already is.

Sources/XCTest/XCTestCase+Performance.swift
```diff
@@ -0,0 +1,166 @@
...
+    ///   must be called once and only once before the end of the block or the test will fail.
+    /// * If `stopMeasuring()` is called multiple times during the block the test will fail.
+    ///
+    /// - Parameter metrics: An array of Strings (XCTPerformanceMetrics) to measure. 
+    ///     Providing an unrecognized string is a test failure.
+    /// - Parameter automaticallyStartMeasuring: If `false`, `XCTestCase` will 
+    ///     not take any measurements until -startMeasuring is called.
+    /// - Parameter block: A block whose performance to measure.
+    /// - ToDo: The `block` param should be marked @noescape once Apple XCTest
+    ///     has been updated to do so. rdar://26224596
```
modocache: Thanks for [the radar](https://openradar.appspot.com/26224596)!

modocache:
This looks fantastic! I only had minor nit-picks, which I feel could adequately be addressed after merging. Considering:

1. The API matches Apple XCTest
2. We can make improvements over time (the implementation doesn't have to be perfect right now)
3. CI is passing

I'd encourage you to hit the big green merge button. 😄 

I don't know much about time measurement, but as for the other things you mention in your pull request:

- Code structure: 💯 
- Test coverage: I agree that unit tests would be nice, but I think we can wait for #61 to add those. A one second sleep doesn't seem like a big deal to me at this point in this codebase's life, but it'd be nice to make it shorter if possible.
- Error reporting: If Apple XCTest reports failures and keeps executing, I'd prefer to do that. `fatalError` seems harsh -- but again, I think we should always do whatever Apple XCTest does. Asynchronous tests report failures for this reason.

ddunbar:
The time measurement parts sound reasonable to me for an initial implementation:
1. Imprecision w.r.t. conversion to Double is a non-issue for this level of API.
2. Using `mach_absolute_time` is the right answer on OS X, via a wrapper in Foundation seems reasonable to me.

Sources/XCTest/PerformanceMeter.swift
briancroom: Good call!

Sources/XCTest/XCTestObservation.swift
briancroom: Yeah, that probably makes sense, in particular considering that we'd like to make the public/internal separation of files more explicit in the file system structure. Thoughts on the name of the file? I feel like having a prefixed underscore on a source filename wouldn't be so great.

Tests/Functional/Performance/Misuse/main.swift
briancroom: Including the line numbers already caught regressions on my part a couple of times while I was working on this 👍

Tests/Functional/Performance/Misuse/main.swift
briancroom: Indeed.

Tests/Functional/Performance/main.swift
briancroom: I opted for a full second because it (1) allowed for a simpler regex when contrasting the values on line 88, and (2) seemed like it would decrease the risk of system load on the CI system causing the failure to not occur because of the other iterations taking longer to execute.

I'm thinking we stick with it for now?

Sources/XCTest/WallClockTimeMetric.swift
briancroom: Ooooops, that line wan't supposed to be included here, good catch. 😆 I had copied it in from the test file for reference when I was writing the method.

Sources/XCTest/WallClockTimeMetric.swift
briancroom: Yes, I'm looking forward to getting unit tests on this code at some point.

briancroom:
Thanks for the positive feedback guys! 😎 I'm going to go ahead and merge this. Exciting that it appears to be making it in before the creation of the `swift-3.0-preview-1-branch`.

briancroom:
Oh, regarding modules - I do believe it should be possible to split the library internals into multiple modules without require consumers to modify their imports (which would make this a non-starter), but I'm not certain yet. @ddunbar or @mxcl could likely confirm the feasibility of this given their experience with SwiftPM?

Tests/Functional/Performance/main.swift
modocache: Fine by me! 😁

Sources/XCTest/XCTestObservation.swift
modocache: At the risk of posting another nit-pick, I had to re-read this two or three times before I realized `_XCTestObservation` was not the same as `XCTestObservation`. How about `XCTestInternalObservation`, to emphasize the difference and alleviate your concerns of an underscore prefix?

PR #112, opened 2016-05-16
[Tests] Automatically substitute [[@LINE]] numbers

modocache:
Enhance xctest_checker such that it substitutes the string `"[[@LINE]]"` and `"[[@LINE+<offset>]]"` with the current line number of the source file when performing checks.

This allows us to add and remove arbitary lines in the functional test suite without manually updating line numbers in most cases. Line offsets will still need to be updated when the distance between the CHECK and the test failure changes.

To test the offsets out, use them in one of the asynchronous tests.

PR #113, opened 2016-05-16
[Tests] Use line number offsets for assertions

modocache:
Use the line number offsets introduces in https://github.com/apple/swift-corelibs-xctest/pull/112 to make the functional test suite assertions less brittle. Now contributors may add and remove arbitary lines in the functional test suite without manually updating line numbers in most cases. Line offsets will still need to be updated when the distance between the CHECK and the test failure changes.


modocache:
@swift-ci please test

modocache:
Tests pass, I think this is a significant improvement, and very few currently open pull requests will be affected by merge conflicts. I'll go ahead and merge!

briancroom:
Great! 👏

PR #115, opened 2016-05-23
[SR-1589] Add non-throwing testCase() wrapper function.

ddunbar:
- This is useful in clients which would like to use Swift type inference to
   define the `allTests` array. If none of the tests in the class throw, Swift
   will infer the type of that array as `(String, (T) -> () -> ())` which is not
   convertible to the type expected by the `testCase` method.

 - Resolves: https://bugs.swift.org/browse/SR-1589


ddunbar:
CC @modocache @briancroom

ddunbar:
I didn't intentionally close this, part of the GitHub server issues this morning I think. Reopened as #116

PR #116, opened 2016-05-24
[SR-1589] Add non-throwing testCase() wrapper function.

ddunbar:
- This is useful in clients which would like to use Swift type inference to
   define the `allTests` array. If none of the tests in the class throw, Swift
   will infer the type of that array as `(String, (T) -> () -> ())` which is not
   convertible to the type expected by the `testCase` method.

 - Resolves: https://bugs.swift.org/browse/SR-1589

### DIFF
--- a/Sources/XCTest/XCTestCase.swift
+++ b/Sources/XCTest/XCTestCase.swift
@@ -169,6 +169,13 @@ public func testCase<T: XCTestCase>(_ allTests: [(String, (T) -> () throws -> Vo
     return (T.self, tests)
 }
 
+/// Wrapper function for the non-throwing variant of tests.
+/// - seealso: `XCTMain`
+public func testCase<T: XCTestCase>(_ allTests: [(String, (T) -> () -> Void)]) -> XCTestCaseEntry {
+    let tests: [(String, (XCTestCase) throws -> Void)] = allTests.map { ($0.0, test($0.1)) }
+    return (T.self, tests)
+}
+
 private func test<T: XCTestCase>(_ testFunc: (T) -> () throws -> Void) -> (XCTestCase) throws -> Void {
     return { testCaseType in
         guard let testCase = testCaseType as? T else {


Sources/XCTest/XCTestCase.swift
```diff
@@ -178,3 +178,20 @@ private func test<T: XCTestCase>(_ testFunc: (T) -> () throws -> Void) -> (XCTes
...
 }
+
+/// Wrapper function for the non-throwing variant of tests.
+/// - seealso: `XCTMain`
+public func testCase<T: XCTestCase>(_ allTests: [(String, (T) -> () -> Void)]) -> XCTestCaseEntry {
+    let tests: [(String, (XCTestCase) throws -> Void)] = allTests.map { ($0.0, test($0.1)) }
+    return (T.self, tests)
+}
+
+private func test<T: XCTestCase>(_ testFunc: (T) -> () -> Void) -> (XCTestCase) throws -> Void {
```
briancroom: I don't think it's necessary to add this overload, because the type conversion should allow the existing `test(_:)` function to be used (no tuples involved in the argument type).

briancroom:
Thanks for the PR @ddunbar! This is a nice improvement. I left one minor comment on the code.

A little testing seems to indicate that the alternative `allTests` form can be used in all cases, including when mixing throwing and non-throwing tests. So I wonder if we should actually update our tests and change the recommended form to use this now?

```swift
class FooTests: XCTestCase {
  func testFoo() {}
  // Notice use of type inference to avoid boilerplate.
  static var allTests = {
    ("testFoo", testFoo)
  }()
}
```

briancroom:
@swift-ci please test

ddunbar:
Yes, the alternative form will work in all cases now, because either of the types Swift infers will work. I think it would be a good idea to update to this as the standard recommendation as it reduces some of the redundant boilerplate that currently needs to be copied and then modified (to change the test case name).

I can follow up on that.

Sources/XCTest/XCTestCase.swift
ddunbar: Ah yes, good point. Fixed and rebased (c65b335).

ddunbar:
@swift-ci please test

briancroom:
The Linux test failure is caused by the ongoing issues with Foundation's test suite. I tested it locally and it looks good to go.

@ddunbar If you don't have time, I have no problem updating the other tests and docs here myself. Thanks for bringing this up!

PR #118, opened 2016-05-26
[Tests] Set LD_LIBRARY_PATH when running tests

modocache:
https://github.com/apple/swift/pull/2609 introduced a change to the default behavior of the Swift driver, which now uses the gold linker unless specified otherwise. This caused a linker error in swift-corelibs-xctest's test suite, which this commit fixes.

/cc @compnerd -- does this seem like a reasonable fix? To be honest, I'm not sure why such a change would be necessary for gold but not bfd.

### DIFF
--- a/Tests/Functional/lit.cfg
+++ b/Tests/Functional/lit.cfg
@@ -76,6 +76,7 @@ else:
         '-I', foundation_dir,
         '-I', core_foundation_dir,
     ])
+    config.environment['LD_LIBRARY_PATH'] = foundation_dir
 
 # Having prepared the swiftc command, we set the substitution.
 config.substitutions.append(('%{swiftc}', ' '.join(swift_exec)))


modocache:
@swift-ci please test

ddunbar:
This doesn't _feel_ right to me, I'm not sure why clients should be expected to change as a result of that PR.

modocache:
I wonder if the real issue has more to do with swift-corelibs-foundation and how it packages `libFoundation.so`, as opposed to the library search paths our tests use at runtime. I'm still looking into it.

How would you feel about merging this as a short-term fix? I feel bad having broken @rintaro and @RLovelett's builds. 😞

compnerd:
This is okay to do.  However, I think it would be better to just embed an DT_RUNPATH in the test binaries as they are tests.  (--enable-new-dtags --rpath /path/to/libraries).

modocache:
The Foundation failure is preventing the Linux tests from running on CI, so we don't have a ton of visibility into how this affects the Linux build. That being said, I can assure you that this "works for me" 😁

modocache:
Created a task to track uncovering why this change was needed: [SR-1625](https://bugs.swift.org/browse/SR-1625). For now, I'll merge this and unbreak @rintaro and @RLovelett's builds.

PR #120, opened 2016-05-31
[build] Add an rpath of $ORIGIN.

ddunbar:
- This allows XCTest to find Foundation adjacent to itself, and fixes runtime
   linker problems when using Gold, which automatically uses DT_RUNPATH instead
   of DT_RPATH and so no longer inherits the search path into the Swift runtime
   libraries location used at build time on the client (which `swiftc` currently
   automatically injects).

 - Fixes SR-1631.

### DIFF
--- a/build_script.py
+++ b/build_script.py
@@ -155,6 +155,9 @@ class GenericUnixStrategy:
                 source_paths=" ".join(sourcePaths)))
         run("{swiftc} -emit-library {build_dir}/XCTest.o "
             "-L {foundation_build_dir} -lswiftGlibc -lswiftCore -lFoundation -lm "
+            # We embed an rpath of `$ORIGIN` to ensure other referenced
+            # libraries (like `Foundation`) can be found solely via XCTest.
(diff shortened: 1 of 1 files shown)


ddunbar:

gribozavr:
LGTM

gribozavr:
@swift-ci Please test

modocache:
Should this also remove the code added in https://github.com/apple/swift-corelibs-xctest/pull/118, which was a workaround for this issue?

ddunbar:
I don't _think_ so, because at test time in the build root, `libXCTest.so` is not adjacent to `libFoundation.so`, so `$ORIGIN` won't work.

If, however, we move to a toolchain based development model where XCTest is built against a synthesized toolchain, *and* it also installs `libXCTest.so` into that so that the built toolchain looks like an installed one, *then* I think we can get rid of it.

ddunbar:

ddunbar:
@swift-ci please test

ddunbar:
Linux passed, which is what matters; the OS X failure appears unrelated.

PR #123, opened 2016-06-08
[SR-1417] Add non-optional overloads of XCTAssertEqual and XCTAssertN…

nsalmoria:
…otEqual

Previously, the only version of the functions that accepted values was the one that implicitly wraps them into Optionals. This generated a confusing error message when the assert failed. Having a separate overload that accepts non-optional types ensures that the correct description is printed when the assert fails.

### DIFF
--- a/Tests/Functional/FailureMessagesTestCase/main.swift
+++ b/Tests/Functional/FailureMessagesTestCase/main.swift
@@ -18,6 +18,7 @@ class FailureMessagesTestCase: XCTestCase {
     static var allTests = {
         return [
             ("testAssert", testAssert),
+            ("testAssertEqualValues", testAssertEqualValues),
             ("testAssertEqualOptionals", testAssertEqualOptionals),
             ("testAssertEqualArraySlices", testAssertEqualArraySlices),
             ("testAssertEqualContiguousArrays", testAssertEqualContiguousArrays),
@@ -30,6 +31,7 @@ class FailureMessagesTestCase: XCTestCase {
             ("testAssertLessThan", testAssertLessThan),
             ("testAssertLessThanOrEqual", testAssertLessThanOrEqual),
             ("testAssertNil", testAssertNil),
+            ("testAssertNotEqualValues", testAssertNotEqualValues),
             ("testAssertNotEqualOptionals", testAssertNotEqualOptionals),
             ("testAssertNotEqualArraySlices", testAssertNotEqualArraySlices),
             ("testAssertNotEqualContiguousArrays", testAssertNotEqualContiguousArrays),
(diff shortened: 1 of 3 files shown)


Tests/Functional/FailureMessagesTestCase/main.swift
```diff
@@ -49,13 +51,20 @@ class FailureMessagesTestCase: XCTestCase {
...
+
 // CHECK: Test Case 'FailureMessagesTestCase.testAssertEqualOptionals' started at \d+:\d+:\d+\.\d+
 // CHECK: test.swift:[[@LINE+3]]: error: FailureMessagesTestCase.testAssertEqualOptionals : XCTAssertEqual failed: \("Optional\(1\)"\) is not equal to \("Optional\(2\)"\) - message
 // CHECK: Test Case 'FailureMessagesTestCase.testAssertEqualOptionals' failed \(\d+\.\d+ seconds\).
     func testAssertEqualOptionals() {
-        XCTAssertEqual(1, 2, "message", file: "test.swift")
+        XCTAssertEqual(Optional(1), Optional(2), "message", file: "test.swift")
     }
-
+
```
modocache: nit-pick: Looks like you added some extra whitespace here. Can you remove it?

Tests/Functional/FailureMessagesTestCase/main.swift
```diff
@@ -133,13 +142,20 @@ class FailureMessagesTestCase: XCTestCase {
...
 // CHECK: Test Case 'FailureMessagesTestCase.testAssertNotEqualOptionals' started at \d+:\d+:\d+\.\d+
 // CHECK: test.swift:[[@LINE+3]]: error: FailureMessagesTestCase.testAssertNotEqualOptionals : XCTAssertNotEqual failed: \("Optional\(1\)"\) is equal to \("Optional\(1\)"\) - message
 // CHECK: Test Case 'FailureMessagesTestCase.testAssertNotEqualOptionals' failed \(\d+\.\d+ seconds\).
     func testAssertNotEqualOptionals() {
-        XCTAssertNotEqual(1, 1, "message", file: "test.swift")
+        XCTAssertNotEqual(Optional(1), Optional(1), "message", file: "test.swift")
     }
-
+
```
modocache: nit-pick: Same as above, looks like you added some extra whitespace here. Can you remove it?

nsalmoria:
@modocache done!

modocache:
Very excited to merge this change!! Thanks, @nsalmoria!

@swift-ci please test

modocache:
Thanks, @nsalmoria! 💯

ddunbar:
I haven't looked closely, but is it possible that this means there is code Linux XCTest will accept that Darwin XCTest will reject?

modocache:
@ddunbar No, I don't think you have to worry about that here, for two reasons:

1. This same change has been applied to the Swift overlay for Darwin XCTest: https://github.com/apple/swift/pull/2788. So, assuming one is using versions of Swift and swift-corelibs-xctest that were both built after two days ago, they should be identical.
2. The optional overloads simply improve the failure message for `XCTAssertEqual`. There's no change to how that function is invoked.

But let me know if I'm missing something here!

PR #127, opened 2016-06-24
Make libdispatch available to Tests/Functional if built

seabaylea:
When libdispatch is build into Foundation, XCTest needs access to it. The XCTest build has already been updated to handle that, and runtime linkage works for the full toolchain because Foundation has an rpath of ORIGIN, and libdispatch and libFoundation are in the same directory.

These changes are for the XCTest tests, which can't use the toolchain because its not built yet. They add the build time flags to the tests, and creates a symlink from the Foundation build directory to the libdispatch library so that its picked up on the rpath of ORIGIN. This is similar to the approach being taken in Swift PM already.

### DIFF
--- a/Tests/Functional/lit.cfg
+++ b/Tests/Functional/lit.cfg
@@ -77,6 +77,15 @@ else:
         '-I', core_foundation_dir,
     ])
     config.environment['LD_LIBRARY_PATH'] = foundation_dir
+    # as well as libdispatch if Foundation is using it
+    libdispatch_src_dir = os.getenv('LIBDISPATCH_SRC_DIR')
+    libdispatch_build_dir = os.getenv('LIBDISPATCH_BUILD_DIR')
+    if (libdispatch_src_dir is not None) and (libdispatch_build_dir is not None):
+        swift_exec.extend([
+            '-Xcc', '-fblocks',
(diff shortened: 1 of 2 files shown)


modocache:
This all looks great, thank you! I'm excited to update the asynchronous tests to use a `dispatch_after`, which will probably be less verbose than using `NSRunLoop`.

@swift-ci please test

modocache:
Oh, although I guess the tests should continue to use Foundation's `NSRunLoop`, since the libdispatch build is conditional -- is that right?

seabaylea:
Its currently conditional but we're trying to get to it being a formal part of the toolchain as soon as possible.

We've now got libdispatch in the CI, and this is one of 3 PRs which will allow us to then enable libdispatch to be used by Foundation. Once that's working and stable, the final step will be to add libdispatch into the Linux buildbot for the full toolchain.

briancroom:
This is great @seabaylea! Thanks for your hard work on getting libdispatch into the standard toolchain.

modocache:
It'll be great to have libdispatch integrated into the toolchain, both for corelibs-xctest and for @briancroom's work in [SR-710](https://bugs.swift.org/browse/SR-710). Thanks!

PR #131, opened 2016-06-29
[Tests] Allow xctest_checker to read from stdin

modocache:
It's sometimes useful to run xctest_checker directly when debugging the test suite. To make it easier to compare the actual output of an executable against the expected output, allow xctest_checker to read from stdin. This makes the following possible:

```
Tests/Functional/MyTestCase/Output/MyTestCase | xctest_checker.py - Tests/Functional/MyTestCase/main.swift
```

This pipes the output from the "MyTestCase" executable into xctest_checker, which compares that output to the expected output from "MyTestCase/main.swift".


modocache:
@swift-ci please test

modocache:
Tests pass! I think the added flexibility here is a big improvement. @mike-ferris-apple, @briancroom, please merge if you agree.

briancroom:
Love it! Thanks for this enhancement.

PR #124, opened 2016-06-13
[SE-0025] Adapt accessibility for `fileprivate`

modocache:
The standard migration path for SE-0025 is to find-and-replace all uses of `private` with `fileprivate`. However, this expands accessibility beyond the necessary minimum. This commit uses compiler errors to determine the minimum set of changes.

This should only be tested/merged once https://github.com/apple/swift/pull/3000 has landed in apple/swift. Thanks to @CodaFi for making it easy to enable the new `fileprivate` errors -- to discover the necessary changes I applied the following change to https://github.com/apple/swift/pull/3000 (since that pull request now aliases `private` to `fileprivate` for migration purposes):

```diff
diff --git a/lib/AST/NameLookup.cpp b/lib/AST/NameLookup.cpp
index bee6b2f..149c8a4 100644
--- a/lib/AST/NameLookup.cpp
+++ b/lib/AST/NameLookup.cpp
@@ -1017,7 +1017,7 @@ static bool checkAccessibility(const DeclContext *useDC,
 
   switch (access) {
   case Accessibility::Private:
-//      return useDC == sourceDC || useDC->isChildContextOf(sourceDC);
+     return useDC == sourceDC || useDC->isChildContextOf(sourceDC);
   case Accessibility::FilePrivate:
     return useDC->getModuleScopeContext() == sourceDC->getModuleScopeContext();
   case Accessibility::Internal: {
```

We could probably refactor a bit after this pull request, to reduce API from `fileprivate` to `private` where possible.

PR #134, opened 2016-07-11
Update for SE-0025 ('private' and 'fileprivate')

jrose-apple:
[SE-0025](https://github.com/apple/swift-evolution/blob/master/proposals/0025-scoped-access-level.md) makes `private` restrict access to a lexical scope; the Swift 2 behavior is now named `fileprivate`. At the time of this pull request, the compiler treats both `private` and `fileprivate` as having `fileprivate` semantics, but that will soon change.

Supersedes #124; the latest amendment to SE-0025 permits some accesses that the original model (or at least our interpretation of it) would have forbidden.

### DIFF
--- a/Sources/XCTest/Private/PrintObserver.swift
+++ b/Sources/XCTest/Private/PrintObserver.swift
@@ -66,7 +66,7 @@ internal class PrintObserver: XCTestObservation {
         return formatter
     }()
 
-    private func printAndFlush(_ message: String) {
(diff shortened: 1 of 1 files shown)


jrose-apple:
cc @mike-ferris-apple, @modocache

mike-ferris-apple:
Looks reasonable to me...

modocache:
tim-cook-gold-macbook

Looks good to me -- thanks @jrose-apple!

PR #142, opened 2016-07-28
[Performance] Remove redundant state update

modocache:
`PerformanceMeter.state` is updated from within the `recordAPIViolation()` method, so there's no need to update it prior to calling that method as well.

/cc @briancroom


modocache:
@swift-ci please test

modocache:
A straightforward change with passing tests -- I'll merge this and leave any code review for post-commit.

PR #143, opened 2016-07-28
[Performance] Pass correct file/line to methods

modocache:
Were these two callsites to fail for some reason, they would display the file and line number for an internal piece of swift-corelibs-xctest source code, rather than the callsite in the user's test. Thread through the user calliste file and line to fix the issue.

In practice, these two callsites cannot possibly fail:

- The call to `startMeasuring()` occurs before the user has any chance to affect the measurement process, so the meter could not possibly be in an invalid state.
- The call to `stopMeasuring()` is only done if measuring is in progress.

Since these cannot fail, it's impossible to write tests for this code... still, if the structure of this code changes in the future, this commit could prevent issues from popping up.

/cc @briancroom

### DIFF
--- a/Sources/XCTest/Private/PerformanceMeter.swift
+++ b/Sources/XCTest/Private/PerformanceMeter.swift
@@ -186,7 +186,7 @@ public final class PerformanceMeter {
 
     private func stopMeasuringIfNeeded() {
         if state == .iterationStarted {
-            stopMeasuring()
+            stopMeasuring(file: invocationFile, line: invocationLine)
         }
     }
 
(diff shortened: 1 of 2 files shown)


modocache:
@swift-ci please test

modocache:
@swift-ci please test

modocache:
A straightforward change with passing tests -- I'll merge this and leave any code review for post-commit.

briancroom:
Going back through some of these PRs I missed while on vacation.. I agree this is a worthwhile change. Thanks @modocache!

modocache:
Thanks! It'd be great to talk about [performance testing baselines](https://lists.swift.org/pipermail/swift-corelibs-dev/Week-of-Mon-20160718/000809.html) sometime soon as well. These API are great to have, but they won't point out performance regressions on Linux as-is.

PR #145, opened 2016-07-29
Add version number to linux libraries

jdfergason:
Create a versioned libXCTest shared library per SR-2149

### DIFF
--- a/build_script.py
+++ b/build_script.py
@@ -20,6 +20,7 @@ import platform
 import errno
 
 SOURCE_DIR = os.path.dirname(os.path.abspath(__file__))
+VERSION = (3, 0, 0)
 
 
 def note(msg):
@@ -50,6 +51,19 @@ def symlink_force(target, link_name):
             os.symlink(target, link_name)
         else:
             raise e
+
+
+class SetDir():
+    def __init__(self, newdir):
+        self.newdir = newdir
+        self.olddir = os.getcwd()
+
+    def __enter__(self):
+        os.chdir(self.newdir)
+
+    def __exit__(self, *args):
+        os.chdir(self.olddir)
+
 
 class DarwinStrategy:
     @staticmethod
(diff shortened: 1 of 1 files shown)


build_script.py
```diff
@@ -181,15 +195,34 @@ def build(args):
...
+        xctest_soname = '{}.{}.{}.{}'.format(xctest_lib, *VERSION)
+
         run("{swiftc} -emit-library {build_dir}/XCTest.o "
             "-L {foundation_build_dir} -lswiftGlibc -lswiftCore -lFoundation -lm "
             # We embed an rpath of `$ORIGIN` to ensure other referenced
             # libraries (like `Foundation`) can be found solely via XCTest.
             "-Xlinker -rpath=\\$ORIGIN "
-            "-o {build_dir}/libXCTest.so".format(
+            "-Xlinker -soname={soname} "
+            "-o {build_dir}/{soname} "
```
modocache: Instead of hardcoding the `/` path separator, can we use `os.path.join` here?

build_script.py
```diff
@@ -181,15 +195,34 @@ def build(args):
...
                 build_dir=build_dir,
-                foundation_build_dir=foundation_build_dir))
+                foundation_build_dir=foundation_build_dir,
+                soname=xctest_soname))
+
+        # create symlinks for versions
+        with SetDir(build_dir):
+            # symlink lib.so
+            if not os.path.exists(xctest_lib):
+                note('ln -s {} {}'.format(xctest_soname, xctest_lib))
```
modocache: `os.symlink` creates a symlink. Could we use that instead? `ln -s` would not work on Windows. I understand that Windows is a long way off, but it doesn't hurt to be prepared. 😄

build_script.py
```diff
@@ -181,15 +195,34 @@ def build(args):
...
+                os.symlink(xctest_soname, xctest_lib)
+            # symlink lib.so.MAJOR
+            xctest_so_major = '{}.{}'.format(xctest_lib, VERSION[0])
+            if not os.path.exists(xctest_so_major):
+                note('ln -s {} {}'.format(xctest_soname, xctest_so_major))
```
modocache: `os.symlink` could be used here as well.

build_script.py
```diff
@@ -274,9 +307,20 @@ def install(args):
         _mkdirp(library_install_path)
 
         xctest_so = "libXCTest.so"
+        xctest_so_versioned = "{}.{}.{}.{}".format(xctest_so, *VERSION)
```
modocache: Ooh, very cool usage of the `*` splat operator. 👍

modocache:
The code looks good to me, but I'd like to hear what @gribozavr or @compnerd think about this, and about https://bugs.swift.org/browse/SR-2149 in general.

@swift-ci please test

compnerd:
Perhaps my understanding of the API evolution process in XCTest is misguided, but I don't see this as being particularly useful, especially as implemented.

Currently you just splat out the package version.  However, the library versioning is traditionally more nuanced than such an approach.  The major version component of the version is only changed when the API is changed in an incompatible manner (e.g. structure layout changes, removal of an API, change of formal parameters to a previously existing api, etc).  The minor version is changed for compatible things (e.g. additions).  The micro component is updated on every release.  Each smaller component is reset when a larger component is bumped (e.g. bumping the major would reset minor and micro).

My understanding is that traditionally XCTest has made compatible changes.  As such, the current thing should be `1.n.m` rather than `3.0.0`.  Parallel installation of the DSO gains you nothing since the API changes are compatible.

Im trying to understand what the benefits are that are gained from versioning this particular library as I think that the benefits may outweigh the maintenance here.  However, it may be reasonable to for the time being just hardcode the DSO version to `1` corresponding to the `A` version provided on Darwin-based targets.  If and when the library is changed in a way as to necessitate a new dylib version, we could bump to `2.

jdfergason:
I think there are a couple of benefits to adding versioning to the XCTest library:

1. Some linux distributions (ie, Fedora / RHEL) require it before they will distribute pre-built binaries.  These distributions do not often distinguish between the subtleties of whether or not versioning is useful for an individual library--it's more of a blanket. 

2. It's very easy for developers linking to the library to know when there have been ABI incompatible changes.

I think switching to the scheme @compnerd outlines where we set the SONAME to libXCTest.so.1 would be sufficient for my needs.  If you want me to modify the pull request to implement as described let me know.

compnerd:
That sounds like a pretty convincing case to me.  I think that I would prefer the `.1` approach.

modocache:
Cool, thanks @compnerd! Any thoughts, @gribozavr?

gribozavr:
If we are going to do something like this, we should do it uniformly for all libraries that we ship, starting with `libswiftCore.so`, and we should have a well-understood policy of when to bump the SONAME, and what should it be set to in snapshot and preview builds.

modocache:
I agree. @jdfergason, you may want to send an email to the swift-dev mailing list, to get more opinions on [SR-2149](https://bugs.swift.org/browse/SR-2149).

briancroom:
Hey @jdfergason! As mentioned, we would want to have some additionally discussions around what this kind of change would mean for the rest of the swift.org projects before introducing this versioning. I'm going to close this PR for now but let's revisit once some consensus has been reached!

PR #146, opened 2016-07-29
[noescape by default] Add explicit @escaping for escaping closures

milseman:


milseman:
@swift-ci please test

PR #150, opened 2016-08-01
Add 'open' to all public classes and their members.

jrose-apple:
This can probably be refined by the XCTest team in a later commit.

In preparation for landing [SE-0117](https://github.com/apple/swift-evolution/blob/master/proposals/0117-non-public-subclassable-by-default.md).

### DIFF
--- a/Sources/XCTest/Public/XCAbstractTest.swift
+++ b/Sources/XCTest/Public/XCAbstractTest.swift
@@ -15,20 +15,20 @@
 /// An abstract base class for testing. `XCTestCase` and `XCTestSuite` extend
 /// `XCTest` to provide for creating, managing, and executing tests. Most
 /// developers will not need to subclass `XCTest` directly.
-public class XCTest {
+open class XCTest {
     /// Test's name. Must be overridden by subclasses.
-    public var name: String {
+    open var name: String {
         fatalError("Must be overridden by subclasses.")
     }
 
     /// Number of test cases. Must be overridden by subclasses.
-    public var testCaseCount: UInt {
+    open var testCaseCount: UInt {
         fatalError("Must be overridden by subclasses.")
     }
 
     /// The `XCTestRun` subclass that will be instantiated when the test is run
     /// to hold the test's results. Must be overridden by subclasses.
-    public var testRunClass: AnyClass? {
+    open var testRunClass: AnyClass? {
         fatalError("Must be overridden by subclasses.")
     }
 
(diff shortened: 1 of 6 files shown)


jrose-apple:
cc @mike-ferris-apple, @rjmccall 

@swift-ci Please test

Sources/XCTest/Public/XCAbstractTest.swift
```diff
@@ -39,17 +39,17 @@ public class XCTest {
     ///   ensure compatibility of tests between swift-corelibs-xctest and Apple
     ///   XCTest, you should not set this property. See
     ///   https://bugs.swift.org/browse/SR-1129 for details.
-    public public(set) var testRun: XCTestRun? = nil
+    open open(set) var testRun: XCTestRun? = nil
```
jrose-apple: @rjmccall, is this correct?

modocache:
Looks good to me, thanks @jrose-apple!

A later commit should probably revise some of these.
- `XCTestSuite`, `XCTestCaseRun` and `XCTestSuiteRun`, for example, are probably not meant to be `open`, and instead should be marked `public`.
- `XCTestRun` is also probably only meant to be subclassed by `XCTest{Suite,Case}Run`, and so should be `public` (not `open`) as well.

Thoughts, @mike-ferris-apple, @briancroom?

mike-ferris-apple:
I’d say that first goal should be equivalence with Xcode's XCTest…

That said, it might be interesting to assess the API and make changes to the swift side of both Xcode’s and CoreLib's XCTest in concert.

Mike
'

modocache:
@mike-ferris-apple I believe https://github.com/apple/swift/pull/3882 will import all Objective-C classes from Xcode XCTest as `open`. The changes in this pull request makes corelibs-xctest equivalent.

By my comment above, I meant I think both the Swift overlay for Xcode XCTest *and* swift-corelibs-xctest should declare `XCTestSuite`, `XCTestRun`, `XCTestCaseRun`, and `XCTestSuiteRun` as public. I believe these classes are not meant to be subclassed by users of XCTest -- does that sound right?

mike-ferris-apple:
> On Aug 1, 2016, at 2:25 PM, Brian Gesiak <notifications@github.com> wrote:
> 
> @mike-ferris-apple <https://github.com/mike-ferris-apple> I believe apple/swift#3882 <https://github.com/apple/swift/pull/3882> will import all Objective-C classes from Xcode XCTest as open. The changes in this pull request makes corelibs-xctest equivalent.
> 
That's what I figured.
> By my comment above, I meant I think both the Swift overlay for Xcode XCTest and swift-corelibs-xctest should declare XCTestSuite, XCTestRun, XCTestCaseRun, and XCTestSuiteRun as public. I believe these classes are not meant to be subclassed by users of XCTest -- does that sound right?
> 
Well, practically speaking I think that’s true, and it might be good to try to lock things down a bit. I believe that XCTestRun would need to be subclassed by folks adding new subclasses of XCTest, but I am not sure how practical a use case that is.

I am not up to speed on exactly what we’re solving by introducing this distinction, also, and I would be somewhat cautious about cutting off subclassability just because we can’t think of a reason you’d want to…

For example, does closing off XCTestSuite and XCTestSuiteRun gain us anything? It certainly loses us the case of someone discovering something cool they could do by doing this only to find that they can’t. I need a philosophy lesson (and perhaps technology lesson) as to the appropriate use for this distinction in Swift.

I’d say that at the very least this requires some fairly deep thinking. And probably that thinking should consider how prepared the rest of the overall context in which XCTest exists (on the Xcode side) is affected. in fact, what the IDE and larger testing infrastructure is designed/prepared for is potentially one of the more compelling reasons to close stuff off...

Mike

modocache:
Awesome, thanks for the detailed response. I'm looking forward to hearing what the XCTest team decides on which APIs should be `open` or `public`. It'd be great if you all shared that discussion on the mailing lists! 🙏 

Anyway, sorry for the noise on this pull request... 😅

rjmccall:
Okay, we're just going to merge this in; thanks, all.

PR #152, opened 2016-08-11
Reduce dependency on the bridge() API since it will be changing shortly

phausler:
Some of the incoming changes for Foundation will alter bridging of types. This will allow the same functionality but avoid the `.bridge()` calls


parkera:
@swift-ci please test linux

PR #156, opened 2016-08-16
Fix race condition in an async test which was triggering failures (SR-2332)

briancroom:
When the system running these tests is under load, the assumptions that these tests were making about dates and ordering weren't holding. This could cause `test_immediatelyTrueDelayedFalsePredicateAndObject_passes` to fail intermittently, and `test_delayedTruePredicateAndObject_passes` to not be verifying the situation it was written to target.

This removes these tests' dependency on dates, increasing reliability.

Resolves: https://bugs.swift.org/browse/SR-2332

@mike-ferris-apple @parkera

### DIFF
--- a/Tests/Functional/Asynchronous/Predicates/Expectations/main.swift
+++ b/Tests/Functional/Asynchronous/Predicates/Expectations/main.swift
@@ -1,7 +1,6 @@
 // RUN: %{swiftc} %s -o %T/Asynchronous-Predicates
 // RUN: %T/Asynchronous-Predicates > %t || true
-// Disabled due to: https://bugs.swift.org/browse/SR-2332
-// xxx: %{xctest_checker} %t %s
+// RUN: %{xctest_checker} %t %s
 
 #if os(Linux) || os(FreeBSD)
     import XCTest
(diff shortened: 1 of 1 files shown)


briancroom:

briancroom:
@swift-ci Please test

modocache:
Awesome, thanks @briancroom! Do you think similar changes need to be made for other similar async tests, such as [this one](https://github.com/apple/swift-corelibs-xctest/blob/32e23fbf12947c93736439a446e88a85510eb0b0/Tests/Functional/Asynchronous/Expectations/main.swift#L47-L49)?

mike-ferris-apple:
I agree that the technique in this diff might be good to use in other places where we're relying on dates and timing to test similar stuff.

The specific test you reference seems less likely to be an issue immediately because of the different time values involved, though it does seem the same theoretical issues could be avoided with the same sort of technique.

Assuming the tests pass, though, let's get this merged.

modocache:
Our tests seemed to have passed, so in it goes. We'll need to pick this change into `swift-3.0-branch`.

PR #158, opened 2016-08-18
Rename Predicate to NSPredicate

itaiferber:
Following work done in, and blocking, https://github.com/apple/swift-corelibs-foundation/pull/567
Both commits need to be merged at around the same time to keep CI from breaking


itaiferber:
CI currently running in linked PR

itaiferber:
(Tested manually, though, so it should work)

PR #165, opened 2016-08-19
Remove obsolete AnyObject cast

briancroom:
This cast is no longer needed, and produces a warning indicating that it always succeeds.


briancroom:
@swift-ci Please test

PR #166, opened 2016-08-19
Add @discardableResult annotation to applicable public API methods.

briancroom:
This maintains consistency with the Darwin XCTest APIs. This also cleans up some warnings in the functional test suite which were being generated because this annotation wasn't present.

I'll note that there's a case to be made that `expectation(description:file:line:)` should *not* have `@discardableResult`, because the expectation it produces can only be fulfilled by invoking a method on the returned value. Nonetheless, I think it's better to keep the APIs in sync for right now and revisit in the future.

PR #167, opened 2016-08-20
[build_script] Remove grep since it always passes

modocache:
`xcodebuild` always prints "export" once it begins building an Xcode project or workspace, regardless of whether that build eventually passes or fails. This results in the `grep` statement succeeding, and thus causing the build script to pass, even when the build fails.

The `grep` statement was originally added to make the build output more human readable, not change the exit code of any builds. Remove it for now.

---

`grep` originally added here: https://github.com/apple/swift-corelibs-xctest/pull/97

Discussion on its removal here: https://github.com/apple/swift-corelibs-xctest/pull/166

### DIFF
--- a/build_script.py
+++ b/build_script.py
@@ -109,8 +109,7 @@ class DarwinStrategy:
             "-configuration {style_options} "
             "SWIFT_EXEC=\"{swiftc}\" "
             "SWIFT_LINK_OBJC_RUNTIME=YES "
-            "SYMROOT=\"{build_dir}\" OBJROOT=\"{build_dir}\" "
-            "| grep -v \"    export\"".format(
+            "SYMROOT=\"{build_dir}\" OBJROOT=\"{build_dir}\" ".format(
                 swiftc=swiftc,
                 build_dir=build_dir,
                 style_options=style_options,


modocache:
@swift-ci please test

modocache:
@swift-ci Please test

briancroom:
(Addresses https://bugs.swift.org/browse/SR-2433)

Thanks for opening this. I agree this is the right thing to do. The benefit of the `grep` was marginal to begin with. It would be possible to keep it and fix the exit code behavior using `pipefail` or similar, but I'm not convinced it's worthwhile. 

I don't think we will be able to merge this until the other issue (https://bugs.swift.org/browse/SR-2434) is fixed, though, to avoid breaking the bots. I've got a pretty good lead on what's going on there and I expect I can get it resolved on Monday.

modocache:
Perfect! You took the words right out of my mouth. I was worried, since I don't have a macOS environment that can build Swift right now. I'm glad you're on top of it! 💯

briancroom:
With #168 merged, this should hopefully pass CI now.

briancroom:
@swift-ci Please test

briancroom:
There's a Linux failure in the Foundation test suite, but this change doesn't affect the Linux build in any way and the OS X build passed, so I'm going to merge it in. Thanks for the patch @modocache!

modocache:
Thank you! 👏

PR #168, opened 2016-08-22
Don't have Xcode embed the Swift stdlib into SwiftXCTest.framework

briancroom:
This avoids issues arising from dyld trying to load both Xcode's
stdlib version and the just-built stdlib which may not be compatible.

Resolves: [SR-2434](https://bugs.swift.org/browse/SR-2434)


briancroom:
@swift-ci Please test

briancroom:
This looks to have worked! Merging this now as the first step to getting things cleaned up.

Note that I don't think there is much value in adding this to the `swift-3.0-branch` because no OS X products are actually shipped with Swift, this build style only exists as a convenience to corelibs contributors.

PR #172, opened 2016-09-21
[Asynchronous] Move into separate directory

modocache:
Three changes:

1. Move files related to `XCTestCase+Asynchronous` into a separate directory.
2. Move methods on `XCTestCase` that vend expectations for `Predicate` and `Notification` into their own files.
3. Modify the build script in order to traverse deeply nested files.

Besides general project hygiene, the motivation for this change is, unintuitively, related to libdispatch: `import Dispatch` within `XCTestCase+Asynchronous` would otherwise result in an error indicating that `NSObjectProtocol` (used within the `expectation(forNotification:...)` method) is an ambiguous reference. By keeping the source files small and limiting the scope in which `Dispatch` is imported, we can sidestep the error.


modocache:
I haven't tested this on Linux, so I'd be surprised if it worked on the first try. Here goes nothing! :sweat_smile:

@swift-ci please test

modocache:
@swift-ci please test

modocache:
Somehow I doubt this change would've caused a error in LLVM IR debug info 😂 I'll try kicking CI again tomorrow.

modocache:
@swift-ci please test

briancroom:
Cool, I like these changes, although the ambiguous reference issue with libdispatch does seem like a bug that should be addressed at some point. Do you know if there is a Jira issue on that?

Also from the CI log it looks like you've uncovered a compiler crasher, unless perhaps some stale build products are somehow interfering. Let's give it another try.

briancroom:
@swift-ci please test

modocache:
Ooh, neat. Later today I'll file a bug for both the crasher in CI and for the ambiguous reference I described above, then try to work around the issue here.

modocache:
Sure enough, the glob pattern I added for Linux didn't work. The updated pull request has one that does, so all the source files are included in the compilation.

@swift-ci please test

modocache:
@swift-ci please test

modocache:
An error in swift-corelibs-foundation `NSFileManager` caused the macOS build to fail. Considering it passed before, and the only changes I've made since only affect Linux, I'd be fine merging this. I'll leave it to you though, @briancroom!

modocache:
@swift-ci please test

briancroom:
@swift-ci please test

briancroom:
Nice job tracking that issue down. Let's get this merged, and hopefully get some positive feedback on https://github.com/apple/swift-corelibs-foundation/pull/656 now that that one has gone in as well!

Thanks @modocache.

PR #173, opened 2016-09-21
[SR-1901] Use final to prevent linker errors

modocache:
In the discussion on SR-1901, one potential workaround was mentioned: mark internal variables as final. This works for attributes that are not "open get, internal set" -- let's try it out.

### DIFF
--- a/Sources/XCTest/Public/XCTestCase.swift
+++ b/Sources/XCTest/Public/XCTestCase.swift
@@ -53,20 +53,10 @@ open class XCTestCase: XCTest {
     }
 
     /// The set of expectations made upon this test case.
-    /// - Note: FIXME: This is meant to be a `private var`, but is marked as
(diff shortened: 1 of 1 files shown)


modocache:
@swift-ci please test

modocache:
@swift-ci please test Linux

modocache:
Wow, it worked! @briancroom, @mike-ferris-apple how about we merge this to reduce the amount of "overexposed" API in XCTest, at least until [SR-1901](https://bugs.swift.org/browse/SR-1901) is resolved?

mike-ferris-apple:
Definitely good to not have to mark stuff public that is not. I'd say merge it.

briancroom:
Very nice, I wish we had been aware of this workaround sooner! Thanks for putting this out there, @modocache.

PR #175, opened 2016-10-28
Print to stderr, not stdout

modocache:
Apple XCTest prints to stderr, whereas Swift XCTest prints to stdout. Some users expect stderr output, so print to stderr. Also, update the tests to capture stderr output.

---

Haven't run these tests locally yet, I'm expecting some initial failures.

### DIFF
--- a/Sources/XCTest/Private/PrintObserver.swift
+++ b/Sources/XCTest/Private/PrintObserver.swift
@@ -67,8 +67,8 @@ internal class PrintObserver: XCTestObservation {
     }()
 
     fileprivate func printAndFlush(_ message: String) {
-        print(message)
-        fflush(stdout)
+        fputs("\(message)\n", stderr)
+        fflush(stderr)
     }
 
     private func formatTimeInterval(_ timeInterval: TimeInterval) -> String {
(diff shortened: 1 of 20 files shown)


modocache:
@swift-ci please test

briancroom:
Thanks for opening this @modocache. This seems like a good change to make. The test failures look to be due to output scripts mixing stdout and stderr output. Tests that are using `print` to do output aren't flushing the way the XCTest output does, so the stdout lines all end up at the end of the script.

@ddunbar @aciidb0mb3r Do you know if `swift test` depends on the current behavior at all? I'm not sure how the parallel execution mode parses output from each the test runner processes?

aciidb0mb3r:
TL;DR It doesn't depend on current behaviour

It shouldn't matter where (stdout/stderr) the output goes because `swift test`  just shell out the XCTest binary. For parallel we just display the output (again either stream) of tests which failed.

modocache:
Excellent. I've let this languish a bit. @briancroom (or anyone else!) if you'd like to get this in sooner feel free to commandeer it. I'll probably update this by Monday or so. :)

modocache:
Hmm, two months is too long to keep a pull request open, I think. If anyone still wants this done, please create a task on https://bugs.swift.org.

PR #176, opened 2016-11-25
Add watchOS, tvOS, iOS platforms support

larryonoff:
### Motivation

CoreLibs XCTest only supports desktop platforms. What limits us using CoreLibs XCTest on other platforms like iOS / Android. Xcode provides functionality to use XCTest framework on iOS, but this's Objective-C framework that is tightly coupled to Xcode. I see two main goals of this PR:

1. Start making CoreLibs XCTest framework available for other platforms (not only for desktops).
2. Start getting rid of XCTest framework coupled with Xcode and make CoreLibs XCTest as a basic framework for testing. I understand that this is a big deal. So this's only beginning of work.

PS. I guess this PR strongly relates to apple/swift-corelibs-xctest#61 because currently tests only run for macOS platform.

### TODO

- [ ] tests
- [ ] Android?
- [ ] define a way to run tests on mobile devices
- [ ] what else ?

### DIFF
--- a/XCTest.xcodeproj/project.pbxproj
+++ b/XCTest.xcodeproj/project.pbxproj
@@ -29,7 +29,6 @@
 		AE2FE11C1CFE86E6003EF0D7 /* XCTestCaseSuite.swift in Sources */ = {isa = PBXBuildFile; fileRef = AE2FE1131CFE86E6003EF0D7 /* XCTestCaseSuite.swift */; };
 		AE2FE11D1CFE86E6003EF0D7 /* XCTestInternalObservation.swift in Sources */ = {isa = PBXBuildFile; fileRef = AE2FE1141CFE86E6003EF0D7 /* XCTestInternalObservation.swift */; };
 		AE63767E1D01ED17002C0EA8 /* TestListing.swift in Sources */ = {isa = PBXBuildFile; fileRef = AE63767D1D01ED17002C0EA8 /* TestListing.swift */; };
-		DA7805FA1C6704A2003C6636 /* SwiftFoundation.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = DA7805F91C6704A2003C6636 /* SwiftFoundation.framework */; };
 		DA9D44191D920A3500108768 /* XCNotificationExpectationHandler.swift in Sources */ = {isa = PBXBuildFile; fileRef = DA9D44141D920A3500108768 /* XCNotificationExpectationHandler.swift */; };
 		DA9D441A1D920A3500108768 /* XCPredicateExpectationHandler.swift in Sources */ = {isa = PBXBuildFile; fileRef = DA9D44151D920A3500108768 /* XCPredicateExpectationHandler.swift */; };
 		DA9D441B1D920A3500108768 /* XCTestCase+Asynchronous.swift in Sources */ = {isa = PBXBuildFile; fileRef = DA9D44161D920A3500108768 /* XCTestCase+Asynchronous.swift */; };
@@ -93,7 +92,6 @@
 			isa = PBXFrameworksBuildPhase;
 			buildActionMask = 2147483647;
 			files = (
-				DA7805FA1C6704A2003C6636 /* SwiftFoundation.framework in Frameworks */,
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 		};
@@ -391,7 +389,7 @@
 				CLANG_WARN__DUPLICATE_METHOD_MATCH = YES;
 				CODE_SIGN_IDENTITY = "";
 				COPY_PHASE_STRIP = NO;
-				CURRENT_PROJECT_VERSION = 1;
+				CURRENT_PROJECT_VERSION = 1.0;
 				DEBUG_INFORMATION_FORMAT = dwarf;
 				ENABLE_STRICT_OBJC_MSGSEND = YES;
 				ENABLE_TESTABILITY = YES;
@@ -409,13 +407,22 @@
 				GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE;
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				GCC_WARN_UNUSED_VARIABLE = YES;
-				MACOSX_DEPLOYMENT_TARGET = 10.9;
+				INFOPLIST_FILE = Info.plist;
+				IPHONEOS_DEPLOYMENT_TARGET = 10.0;
+				MACOSX_DEPLOYMENT_TARGET = 10.11;
 				MTL_ENABLE_DEBUG_INFO = YES;
 				ONLY_ACTIVE_ARCH = YES;
-				SDKROOT = macosx;
+				PRODUCT_BUNDLE_IDENTIFIER = org.swift.XCTest;
+				PRODUCT_BUNDLE_PACKAGE_TYPE = BNDL;
+				PRODUCT_NAME = SwiftXCTest;
+				SUPPORTED_PLATFORMS = "appletvos appletvsimulator iphoneos iphonesimulator macosx watchos watchsimulator";
+				SWIFT_ACTIVE_COMPILATION_CONDITIONS = DEBUG;
 				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
+				SWIFT_VERSION = 3.0;
+				TVOS_DEPLOYMENT_TARGET = 10.0;
 				VERSIONING_SYSTEM = "apple-generic";
 				VERSION_INFO_PREFIX = "";
+				WATCHOS_DEPLOYMENT_TARGET = 3.0;
 			};
 			name = Debug;
 		};
@@ -440,7 +447,7 @@
 				CLANG_WARN__DUPLICATE_METHOD_MATCH = YES;
 				CODE_SIGN_IDENTITY = "";
 				COPY_PHASE_STRIP = NO;
-				CURRENT_PROJECT_VERSION = 1;
+				CURRENT_PROJECT_VERSION = 1.0;
 				DEBUG_INFORMATION_FORMAT = "dwarf-with-dsym";
 				ENABLE_NS_ASSERTIONS = NO;
 				ENABLE_STRICT_OBJC_MSGSEND = YES;
@@ -452,34 +459,38 @@
 				GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE;
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				GCC_WARN_UNUSED_VARIABLE = YES;
-				MACOSX_DEPLOYMENT_TARGET = 10.9;
+				INFOPLIST_FILE = Info.plist;
+				IPHONEOS_DEPLOYMENT_TARGET = 10.0;
+				MACOSX_DEPLOYMENT_TARGET = 10.11;
 				MTL_ENABLE_DEBUG_INFO = NO;
-				SDKROOT = macosx;
+				PRODUCT_BUNDLE_IDENTIFIER = org.swift.XCTest;
+				PRODUCT_BUNDLE_PACKAGE_TYPE = BNDL;
+				PRODUCT_NAME = SwiftXCTest;
+				SUPPORTED_PLATFORMS = "appletvos appletvsimulator iphoneos iphonesimulator macosx watchos watchsimulator";
 				SWIFT_OPTIMIZATION_LEVEL = "-Owholemodule";
+				SWIFT_VERSION = 3.0;
+				TVOS_DEPLOYMENT_TARGET = 10.0;
 				VERSIONING_SYSTEM = "apple-generic";
 				VERSION_INFO_PREFIX = "";
+				WATCHOS_DEPLOYMENT_TARGET = 3.0;
 			};
 			name = Release;
 		};
 		5B5D86E41BBC74AD00234F36 /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
-				CLANG_ENABLE_MODULES = YES;
-				COMBINE_HIDPI_IMAGES = YES;
 				DEFINES_MODULE = YES;
-				DYLIB_COMPATIBILITY_VERSION = 1;
-				DYLIB_CURRENT_VERSION = 1;
 				DYLIB_INSTALL_NAME_BASE = "@rpath";
-				FRAMEWORK_VERSION = A;
-				INFOPLIST_FILE = Info.plist;
-				INSTALL_PATH = "$(LOCAL_LIBRARY_DIR)/Frameworks";
-				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/../Frameworks @loader_path/Frameworks";
-				MACOSX_DEPLOYMENT_TARGET = 10.11;
-				PRODUCT_BUNDLE_IDENTIFIER = org.swift.XCTest;
-				PRODUCT_NAME = SwiftXCTest;
+				"FRAMEWORK_SEARCH_PATHS[sdk=macosx*]" = $BUILT_PRODUCTS_DIR;
+				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks @loader_path/Frameworks";
+				"LD_RUNPATH_SEARCH_PATHS[sdk=macosx*]" = "@executable_path/../Frameworks @loader_path/Frameworks";
+				"OTHER_LDFLAGS[sdk=macosx*]" = (
+					"-framework",
+					SwiftFoundation,
+				);
+				PRODUCT_BUNDLE_PACKAGE_TYPE = FMWK;
 				SKIP_INSTALL = YES;
-				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
-				SWIFT_VERSION = 3.0;
+				TARGETED_DEVICE_FAMILY = "1,2,3,4";
 				USER_HEADER_SEARCH_PATHS = $BUILT_PRODUCTS_DIR/usr/local/include/CoreFoundation;
 			};
 			name = Debug;
@@ -487,21 +498,18 @@
 		5B5D86E51BBC74AD00234F36 /* Release */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
-				CLANG_ENABLE_MODULES = YES;
-				COMBINE_HIDPI_IMAGES = YES;
 				DEFINES_MODULE = YES;
-				DYLIB_COMPATIBILITY_VERSION = 1;
-				DYLIB_CURRENT_VERSION = 1;
 				DYLIB_INSTALL_NAME_BASE = "@rpath";
-				FRAMEWORK_VERSION = A;
-				INFOPLIST_FILE = Info.plist;
-				INSTALL_PATH = "$(LOCAL_LIBRARY_DIR)/Frameworks";
-				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/../Frameworks @loader_path/Frameworks";
-				MACOSX_DEPLOYMENT_TARGET = 10.11;
-				PRODUCT_BUNDLE_IDENTIFIER = org.swift.XCTest;
-				PRODUCT_NAME = SwiftXCTest;
+				"FRAMEWORK_SEARCH_PATHS[sdk=macosx*]" = $BUILT_PRODUCTS_DIR;
+				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks @loader_path/Frameworks";
+				"LD_RUNPATH_SEARCH_PATHS[sdk=macosx*]" = "@executable_path/../Frameworks @loader_path/Frameworks";
+				"OTHER_LDFLAGS[sdk=macosx*]" = (
+					"-framework",
+					SwiftFoundation,
+				);
+				PRODUCT_BUNDLE_PACKAGE_TYPE = FMWK;
 				SKIP_INSTALL = YES;
-				SWIFT_VERSION = 3.0;
+				TARGETED_DEVICE_FAMILY = "1,2,3,4";
 				USER_HEADER_SEARCH_PATHS = $BUILT_PRODUCTS_DIR/usr/local/include/CoreFoundation;
 			};
 			name = Release;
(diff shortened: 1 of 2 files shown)


Sources/XCTest/Private/PerformanceMeter.swift
```diff
@@ -11,7 +11,7 @@
 //  Measures the performance of a block of code and reports the results.
 //
 
-#if os(Linux) || os(FreeBSD)
+#if os(Linux) || os(FreeBSD) || os(iOS)
```
modocache: As we discussed on Twitter, I think the way forward here isn't to hard-code Linux, FreeBSD, and now iOS to `import Foundation`. Instead, we should provide a configuration option of some kind, to compile swift-corelibs-xctest to use Foundation or SwiftFoundation, as the user desires.

To put it another way, I'm envisioning something like this:

```swift
#if defined(USE_SWIFT_FOUNDATION_XCODEPROJ)
    import SwiftFoundation
#else
    import Foundation
#endif
```

`USE_SWIFT_FOUNDATION_XCODEPROJ` would be specified by the user when they compile swift-corelibs-xctest, or as part of the SwiftXCTest-iOS scheme.

Sources/XCTest/Public/Asynchronous/XCTestCase+NotificationExpectation.swift
```diff
@@ -45,8 +45,7 @@ public extension XCTestCase {
             .default
             .addObserver(forName: Notification.Name(rawValue: notificationName),
                          object: objectToObserve,
-                         queue: nil,
-                         usingBlock: {
+                         queue: nil) {
```
modocache: I think this is a good change, thanks! But please submit these syntax cleanup changes as a separate pull request.

larryonoff:
@modocache thank you for the review. I'll do changes on Monday.

Sources/XCTest/Private/PerformanceMeter.swift
larryonoff: I think it's better assume that SwiftFoundation is default. So I suggest introducing flag with name like USE_OBJC_FOUNDATION.

modocache: Sounds good to me! @briancroom, any thoughts?

Sources/XCTest/Public/Asynchronous/XCTestCase+NotificationExpectation.swift
larryonoff: Added this change in a separate PR, see apple/swift-corelibs-xctest#177. I rebased this PR on PR with the syntax cleanup. In other case iOS build fails. I'll rebase on master as soon as the syntax cleanup approved.

Sources/XCTest/Private/PerformanceMeter.swift
larryonoff: @modocache @briancroom PR was updated.

Sources/XCTest/Private/PerformanceMeter.swift
```diff
@@ -11,7 +11,7 @@
 //  Measures the performance of a block of code and reports the results.
 //
 
-#if os(Linux) || os(FreeBSD)
+#if USE_OBJC_FOUNDATION || os(Linux) || os(FreeBSD)
```
modocache: OK, this is very close to what I had in mind. My only suggestion would be to remove the `|| os(Linux) || os(FreeBSD)` here. Instead, have the build script, `build_script.py`, define `USE_OBJC_FOUNDATION=1` when compiling for a non-Darwin target.

Of course, it's strange that `USE_OBJC_FOUNDATION` should be defined when building for Linux, so I'd change the name here. Maybe `USE_IMPORT_FOUNDATION` or `USE_IMPORT_SWIFTFOUNDATION`.

larryonoff: I thought ObjC `Foundation` exist for Linux. So just curious to introduce better name what kind of foundation we use for Linux / FreeBSD.

modocache:
@swift-ci Please test

modocache:
@swift-ci please test

parkera:
If you want to do this for XCTest, that's fine - but I do want to point out that we have no plans to make swift-corelibs-foundation a shippable product on Darwin platforms. Our plan all along there has been to continue to use the system provided one.

larryonoff:
CI just pointed out that I forgot update build script for Unix. Just did it, could anyone please check that I did it properly.

modocache:
@swift-ci please test

modocache:
@swift-ci Please test

larryonoff:
@swift-ci Please test

larryonoff:
Could anyone please trigger CI? I made changes that should fix tests.

modocache:
@swift-ci please test

modocache:
@swift-ci Please test

larryonoff:
Since I don't have Linux env I only had to check if the changes for Linux are valid with CI help. Sorry, but I need to trigger ci build again :)

modocache:
@swift-ci Please test

modocache:
No problem! :)

modocache:
@swift-ci please test

larryonoff:
build failed on macOS since llvm clone failed. could you please trigger CI again :)

modocache:
@swift-ci Please test

modocache:
@swift-ci please test

CodaFi:
This is going to keep failing until apple/swift#5280 gets merged.

Sources/XCTest/Private/PerformanceMeter.swift
```diff
@@ -11,7 +11,7 @@
 //  Measures the performance of a block of code and reports the results.
 //
 
-#if os(Linux) || os(FreeBSD)
+#if USE_IMPORT_FOUNDATION
```
briancroom: Instead of adding this conditional compilation flag, how about we simplify the expression to something like:

```swift
#if os(macOS)
import SwiftFoundation
#else
import Foundation
#endif
```

See also previous discussion of this: https://github.com/apple/swift-corelibs-xctest/pull/96

XCTest.xcodeproj/project.pbxproj
```diff
@@ -255,6 +299,24 @@
 /* End PBXLegacyTarget section */
 
 /* Begin PBXNativeTarget section */
+		19D66F931DE6F57E00C112C0 /* SwiftXCTest-iOS */ = {
```
briancroom: Rather than adding a whole new target, it should be pretty straightforward to accomplish the same goal by adding `iphoneos` and `iphonesimulator` to the `SUPPORTED_PLATFORMS` build setting of the existing target. You'll probably need to also bring over a couple of additional settings like the iOS deployment target, but I would it expect it to mostly just work for a simple target like this.

Sources/XCTest/Private/PerformanceMeter.swift
larryonoff: I like the idea of removing USE_IMPORT_FOUNDATION define (I generally don't like the flag). But in the case we excluding the case when we want to run both tests for SwiftFoundation and Foundation of SwiftXCTests on macOS.

@modocache @briancroom any thoughts about this?

XCTest.xcodeproj/project.pbxproj
larryonoff: will this work for with SPM?

larryonoff:
@CodaFi thank you for mentioning. It looks that issue introduced recently. Since I was able passing macOS build some time before.

Sources/XCTest/Private/PerformanceMeter.swift
briancroom: Corelibs XCTest is really intended to build on top of Corelibs Foundation. If it works with Darwin Foundation, that's great, but it's not a core configuration for the project. I don't think we should go out of our way to add support for that as a part of this PR, at least.

XCTest.xcodeproj/project.pbxproj
briancroom: I'm not sure I understand what you're asking here. Could you elaborate?

larryonoff: I meant if we leave one unified target (for macOS, iOS) will Swift Package Manage be able compiling it for macOS and iOS platforms? e.g. I think that Carthage won't be able doing it (but I'm not sure).

Sources/XCTest/Private/PerformanceMeter.swift
larryonoff: I got your point. I'll do change soon.

larryonoff: done

ifreecarve:
What's the relationship between this and XCUItest?

XCTest.xcodeproj/project.pbxproj
briancroom: Well, SwiftPM doesn't build from Xcode projects, so there's be no effect there. (It would be interesting to see sometime though if we could do away with this Xcode project altogether and just let SwiftPM generate it for us!)

I'm not sure about Carthage, though I would consider it to be their bug if it doesn't work, since xcodebuild deals with that setup just fine.

Sources/XCTest/Private/PerformanceMeter.swift
briancroom: These look great 👍

modocache:
@ifreecarve XCUITest is a set of methods in Apple's XCTest framework. This project, swift-corelibs-xctest, doesn't implement those methods, and has no immediate goals to do so. You can send an email to the mailing list, swift-corelibs-dev@swift.org, for more information.

XCTest.xcodeproj/project.pbxproj
ikesyo: > I'm not sure about Carthage, though I would consider it to be their bug if it doesn't work, since xcodebuild deals with that setup just fine.

Carthage is supporting such unified schemes from September 2015.

larryonoff: @ikesyo thanks for pointing it out!

larryonoff: I've migrated into unified scheme. Generally I checked https://github.com/mxcl/PromiseKit settings to do this.

The main thing that can be wrong is that I had to remove `SwiftFoundation.framework` from "Linked Frameworks and Libraries" since it doesn't compile for iOS platform. I checked that there's a setting "Link Frameworks Automatically", but I'm not sure that it works properly.

Could anyone please provide any suggestions?

larryonoff:
Are collaborators / contributors fine adding watchOS / tvOS platforms in scope of this PR?

larryonoff:
I think that It may be useful splitting this PR into two PRs:
1. update #if conditionals in code.
2. add unified target.

Should I?

briancroom:
Splitting this into separate PRs sounds great for keeping things easy to follow. Thanks @larryonoff.

larryonoff:
I moved compilation flags update into separate PR, see apple/swift-corelibs-xctest#180.

XCTest.xcodeproj/project.pbxproj
briancroom: Sorry, I missed this before. I think what we're going to need here is to add `-framework SwiftFoundation` into the `Other Linker Flags` build setting, but conditionalized such that it only applies to `Any macOS SDK`. Similar to what is described [here](http://stackoverflow.com/a/40004807/1217794).

briancroom:
@larryonoff Could you rebase this now that #180 has been merged? Thanks!

larryonoff:
@briancroom Thanks! I'll definitely rebase. This's what I expected to do as soon as #180 merged.

larryonoff:
@briancroom I did rebase. I also did the changes mentioned [here](http://stackoverflow.com/a/40004807/1217794). I'm just curios if it's valid using this setting ```"FRAMEWORK_SEARCH_PATHS[sdk=macosx*]" = $BUILT_PRODUCTS_DIR```.

larryonoff:
@ikesyo could you give any advice what to do with Carthage for the XCTest. When I run command `carthage update --no-use-binaries --platform ios` the following error occurs:
```
The following build commands failed:
	ExternalBuildToolExecution SwiftXCTestFunctionalTests
(1 failure)
error: can't exec '/Users/ilaryionau/Desktop/test-thali/Carthage/Checkouts/swift-corelibs-xctest/../llvm/utils/lit/lit.py' (No such file or directory)
error: can't exec '/Users/ilaryionau/Desktop/test-thali/Carthage/Checkouts/swift-corelibs-xctest/../llvm/utils/lit/lit.py' (No such file or directory)
A shell task (/usr/bin/xcrun xcodebuild -workspace /Users/ilaryionau/Desktop/test-thali/Carthage/Checkouts/swift-corelibs-xctest/XCTest.xcworkspace -scheme SwiftXCTestFunctionalTests -configuration Release -sdk iphoneos ONLY_ACTIVE_ARCH=NO BITCODE_GENERATION_MODE=bitcode CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY= CARTHAGE=YES clean build) failed with exit code 65:
```

The reason is because Carthage builds all targets.

ikesyo:
@larryonoff I'm not sure about it. I don't think the project / workspace is meant to be buildable with Carthage.

larryonoff:
@briancroom could you please check this updated PR.

modocache:
@swift-ci please test

modocache:
@swift-ci Please test

larryonoff:
Checks failed with the following error. Looks that the problem doesn't connected with my changes.
```
[xUnit] [INFO] - Stopping recording.
Setting commit status on GitHub for https://github.com/apple/swift-corelibs-xctest/commit/c2070dafcfe1a872121ada9aa8c003aac3f1a6e2
ERROR: [GitHub Commit Status Setter] - {"message":"No commit found for SHA: c2070dafcfe1a872121ada9aa8c003aac3f1a6e2","documentation_url":"https://developer.github.com/v3/repos/statuses/"}, setting build result to FAILURE
Build step 'Set build status on GitHub commit [deprecated]' changed build result to FAILURE
```

modocache:
@swift-ci Please test

modocache:
@swift-ci Please test

modocache:
@swift-ci please test

larryonoff:
CI throws an error: `<unknown>:0: error: no such file or directory: 'x86_64-apple-macosx10.11'`.
Is the reason that I changed MACOSX_DEPLOYMENT_TARGET from 10.9 to 10.11?

modocache:
Possibly, yes. Is that change necessary?

larryonoff:
I don't think that the change is very important, but anyway strange that we're failing since this reason.

larryonoff:
@briancroom there isn't any feedback for a while so just pinging.

modocache:
@larryonoff The CI failure would prevent this from being merged -- what do you think is the next step here? Do you want help diagnosing the CI failure? Or do you want @briancroom to give his LGTM so that you feel motivated to diagnose the CI failure?

larryonoff:
@modocache My plan is:

1.  Receive LGTM from @briancroom.
2. Fix CI failure. I hoped that @briancroom may confirm that my assumption about changing MACOSX_DEPLOYMENT_TARGET from 10.9 to 10.11 is correct.

I'll make a commit very soon with reverting change 'MACOSX_DEPLOYMENT_TARGET from 10.9 to 10.11', so could you please trigger CI then.

Thanks!

larryonoff:
@modocache I just made a change, let's check if it works on CI.

modocache:
@swift-ci please test

modocache:
@swift-ci Please test

larryonoff:
I cannot access CI log, there're not available. Could anyone please trigger CI again?

modocache:
@swift-ci please test

larryonoff:
Could anyone please re-run tests. I believe that I've fixed CI failure.

modocache:
@swift-ci please test

larryonoff:
I'm not sure that CI started, because I don't see any CI results.

modocache:
@swift-ci please test

larryonoff:
Now I don't understand what's wrong, because It builds fine on my machine, but not in CI. I've just made one change. I hope but not sure that it will help.

modocache:
@larryonoff A good next step would be to look at the CI log, and run the exact same invocation locally. Swift CI usually builds a larger portion of all the Swift projects, and sometimes with unique settings.

@swift-ci please test

modocache:
@swift-ci please test

ikesyo:
```
dyld: lazy symbol binding failed: Symbol not found: __os_log_set_nscf_formatter
  Referenced from: /Users/buildnode/jenkins/workspace/swift-corelibs-xctest-PR-osx/Ninja-DebugAssert/xctest-macosx-x86_64/Debug/SwiftFoundation.framework/Versions/A/SwiftFoundation
  Expected in: /usr/lib/libSystem.B.dylib

dyld: Symbol not found: __os_log_set_nscf_formatter
  Referenced from: /Users/buildnode/jenkins/workspace/swift-corelibs-xctest-PR-osx/Ninja-DebugAssert/xctest-macosx-x86_64/Debug/SwiftFoundation.framework/Versions/A/SwiftFoundation
  Expected in: /usr/lib/libSystem.B.dylib
```

It seems that by https://github.com/apple/swift-corelibs-foundation/pull/709, Swift Foundation does not work well on macOS 10.11 now. What is the version of the CI environment?

modocache:
I can see that swift-corelibs-foundation only runs tests on Linux, and its build log prints out the version Ubuntu 16.04. Unfortunately the macOS build doesn't seem to print out the current macOS version. I can see that it's using the MacOSX10.12.sdk, but that doesn't mean much.

@shahmishal, so you know which version of macOS the CI is running? Would you merge a pull request to print out the macOS version in build-script?

ikesyo:
I've submitted a pull request: https://github.com/apple/swift-corelibs-foundation/pull/784

larryonoff:
@modocache PR apple/swift-corelibs-foundation#784 merged. Could you please trigger CI?

modocache:
@swift-ci please test

shahmishal:
@modocache 

```
ProductName:	Mac OS X
ProductVersion:	10.11.5
BuildVersion:	15F34
```

ikesyo:
Another Foundation pull request related to the failure: https://github.com/apple/swift-corelibs-foundation/pull/785.

briancroom:
Sorry guys, I was on holiday for a bit and my notification settings were messed up so I missed the activity and lost track here. I'm going to take a moment to catch up now.

XCTest.xcodeproj/project.pbxproj
```diff
@@ -409,13 +407,22 @@
 				GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE;
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				GCC_WARN_UNUSED_VARIABLE = YES;
-				MACOSX_DEPLOYMENT_TARGET = 10.9;
+				INFOPLIST_FILE = Info.plist;
+				IPHONEOS_DEPLOYMENT_TARGET = 10.0;
```
briancroom: I don't any specific concerns with these deployment targets, but I'm curious how you picked them? Is there a reason why it couldn't deploy further back?

XCTest.xcodeproj/project.pbxproj
```diff
@@ -409,13 +407,22 @@
...
+				MACOSX_DEPLOYMENT_TARGET = 10.11;
 				MTL_ENABLE_DEBUG_INFO = YES;
 				ONLY_ACTIVE_ARCH = YES;
-				SDKROOT = macosx;
```
briancroom: These `SDKROOT` changes are surprising to me. Can you explain what's going on here? This corresponds to the `Base SDK` setting in the Xcode UI.

briancroom:
I left a couple of minor questions, but this looks good to me, @larryonoff! Thanks for your work here.

I would like to wait to merge until we can get CI green on OS X, though. Thanks @ikesyo for helping to deal with the Foundation build failures we're seeing here. I'm hopeful we can get through this soon!

XCTest.xcodeproj/project.pbxproj
larryonoff: There're two reasons, see them below:

1. `swift-corelibs-xctest` uses [Timer.scheduledTimer(withTimeInterval:repeats:block:)](https://github.com/apple/swift-corelibs-xctest/blob/242864abf579529281ed29932c58341fc8962525/Sources/XCTest/Private/XCPredicateExpectation.swift#L36) that requires iOS 10.0+, macOS 10.12+, tvOS 10.0+, watchOS 3.0+. The [link](https://developer.apple.com/reference/foundation/timer/2091889-scheduledtimer) to the documentation.
2. `SwiftFoundation` macOS deployment target is 10.11.

PS. I think that at the moment we don't need deployment targets lower that specified. We can decrease deployment target as soon as people need it, but in this case we need to make some changes in the code.

XCTest.xcodeproj/project.pbxproj
larryonoff: Very good question!

Please note that most of the changes I made are based on [PromiseKit](https://github.com/mxcl/PromiseKit) and the [article](http://promisekit.org/news/2016/08/Multiplatform-Single-Scheme-Xcode-Projects/).

We also should note that we're making the multi-platform single-scheme framework now, so each platform needs it's own `SDKROOT`.

I tried the same setup for `SDKROOT` as for `PromiseKit` (just by removing `SDKROOT` on project level) and it works for `swift-corelibs-xctest`. But I had to add to specify `SDKROOT` for `SwiftXCTestFunctionalTests` target, because without `SDKROOT` it failed to build.

**PS**. I agree that it would be better having `SDKROOT` on project level for each platform, but Xcode only allows configuring it for architecture, not for platform.

It would be awesome if you provide any better solution. Thanks!

XCTest.xcodeproj/project.pbxproj
briancroom: Makes sense! Thanks for the explanation.

XCTest.xcodeproj/project.pbxproj
briancroom: In my testing, Xcode will automatically adjust the `SDKROOT` used for a build such that it matches the platform of the run destination. For example, if the selected SDK is "Latest macOS" and the build has an iOS device destination selected, it will implicitly use the "Latest iOS" SDK instead. Therefore I would prefer keeping the Base SDK setting at the project level with no target overrides.

One other minor point: I think it would be better to specify the new `SUPPORTED_PLATFORMS` value at the level of the `SwiftXCTest` target, rather than at the project level, for now anyway. This is because the `SwiftXCTestFunctionalTests` target only works on macOS, and I don't foresee that changing.

larryonoff: I did the following:

- move SUPPORTED_PLATFORMS from project level to 'XCTest' target.
- move deployment targets  from project level to 'XCTest' target. Because there're relevant only for 'XCTest' target.
- revert SDKROOT to macosx for project level.

larryonoff:
Could anyone please trigger CI again. It should still fail since apple/swift-corelibs-foundation#785 not merged. But I would like to see that nothing is broken since my current changes.

modocache:
@swift-ci please test

larryonoff:
Please trigger CI, since PR apple/swift-corelibs-foundation#785 merged.

rintaro:
@swift-ci Please test

larryonoff:
All CI checks finally passed! Thanks all.

modocache:
🎉 🎉 🎉 

Thanks for all the persistence, everyone! :)

briancroom:
Excellent!

PR #178, opened 2016-12-01
Silence compiler warnings about using optionals with string interpolation.

briancroom:
I'm not sure when this warning was introduced, but I noticed these now when building the project on my machine.

Fixes: `warning: string interpolation produces a debug description for an optional value; did you mean to make this explicit?`


briancroom:
@swift-ci please test

PR #180, opened 2016-12-02
update conditional compilation flags

larryonoff:
This a part of PR apple/swift-corelibs-xctest#176 to simplify base PR.

XCTMain.swift concernes me a lot, but still this file changes look correct for me.

### DIFF
--- a/Sources/XCTest/Public/XCTestMain.swift
+++ b/Sources/XCTest/Public/XCTestMain.swift
@@ -12,12 +12,16 @@
 //  for running tests and some infrastructure for running them.
 //
 
-#if os(Linux) || os(FreeBSD)
-    import Glibc
-    import Foundation
-#else
-    import Darwin
+#if os(macOS)
     import SwiftFoundation
+#else
+    import Foundation
+#endif
+
+#if os(macOS)
+    import Darwin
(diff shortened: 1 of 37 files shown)


Sources/XCTest/Public/XCTestMain.swift
```diff
@@ -12,14 +12,20 @@
...
 #if os(Linux) || os(FreeBSD)
     import Glibc
-    import Foundation
-#else
+#elseif os(macOS)
     import Darwin
-    import SwiftFoundation
 #endif
 
+#if os(macOS) || os(Linux) || os(FreeBSD)
```
briancroom: What's the motivation for introducing this conditional? Does it fail to build in on other platforms for some reason?

larryonoff: @briancroom thank you for pointing this out. I thought that `func XCTMain` won't compile for iOS (e.g. since exit method usage), but I was wrong. I just checked and it compiles. So I removed this condition.

briancroom:
@swift-ci please test

briancroom:
This looks great! Thanks a lot @larryonoff.

PR #183, opened 2017-01-12
Re-export Foundation from XCTest

briancroom:
**Resolves: [SR-3615](https://bugs.swift.org/browse/SR-3615)**

This brings Corelibs XCTest into closer parity with Apple XCTest in that importing XCTest from a test source file is sufficient to allow using Foundation symbols without needing to explicitly import Foundation in the test.

This is not a source-breaking change, as existing code that already imports Foundation can continue to do so with no ill effects.

This change has a particular benefit to the Corelibs XCTest codebase itself because it significantly reduces the number of platform-conditional imports that we need.

### DIFF
--- a/Sources/XCTest/Private/PerformanceMeter.swift
+++ b/Sources/XCTest/Private/PerformanceMeter.swift
@@ -11,12 +11,6 @@
 //  Measures the performance of a block of code and reports the results.
 //
 
-#if os(macOS)
-    import SwiftFoundation
-#else
-    import Foundation
-#endif
-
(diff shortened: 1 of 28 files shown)


briancroom:
@swift-ci please test

briancroom:
The OS X CI build is still failing to build Corelibs Foundation, but I've tested this a bunch locally and am confident that these changes don't regress. Because I want this fix to be on master before the final sync with the swift-3.1-branch, I'm going to go ahead and merge without waiting for the CI trouble to be cleared up.

PR #186, opened 2017-04-13
Add date to time formatted messages

alblue:
On Darwin, messages are printed with the date as well:

    Test Suite 'All tests' started at 2017-04-12 16:50:11.630

On Linux, messages used to be printed without the date:

    Test Suite 'All tests' started at 16:06:28.533

Fixes SR-4573

### DIFF
--- a/Sources/XCTest/Private/PrintObserver.swift
+++ b/Sources/XCTest/Private/PrintObserver.swift
@@ -53,7 +53,7 @@ internal class PrintObserver: XCTestObservation {
 
     private lazy var dateFormatter: DateFormatter = {
         let formatter = DateFormatter()
-        formatter.dateFormat = "HH:mm:ss.SSS"
+        formatter.dateFormat = "yyyy-MM-dd HH:mm:ss.SSS"
         return formatter
     }()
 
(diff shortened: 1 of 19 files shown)


czechboy0:
@swift-ci please test

alblue:
OK there is a test specifically looking for the date format without the year, so I'll fix that.

briancroom:
Unfortunately the testing strategy in use here relies heavily on inspecting the output from the test runner, so a change like this which affects a lot of output will require extensive adjustments in the tests. That said, it should end up being a pretty mechanical set of changes.

alblue:
@swift-ci please test

czechboy0:
@swift-ci please test

briancroom:
It looks like we may have a flaky test in `PredicateExpectationsTestCase.test_delayedTruePredicateAndObject_passes`, in that it is too time-dependent. Let's try the run again and see what happens.

briancroom:
@swift-ci Please test OS X platform

alblue:
@swift-ci please test os x platform

alblue:
@swift-ci please clean test osx platform

alblue:
So this test on OSX is flaky, and the change is for the Linux specific part of the build in any case. What do we do to move this forward?

briancroom:
Since the test failure is clearly unrelated to your change, the Linux build shows that the tests have all been updated appropriately, and the failing OS X build isn't a part of the primary Swift CI, I'm fine with this going in as-is. Thanks for putting this together!

@alblue could you file a bug about the flaky test so that we don't lose track of it?

PR #195, opened 2017-06-27
Avoid importing CoreFoundation in Xcode builds

phausler:
CoreFoundation was being imported as a module which ended up attempting to pull in headers from the system CoreFoundation. This causes build failures when attempting to test Foundation using Xcode.

The change listed avoids this failure by not letting the module map sneak into the XCTest build


briancroom:
@swift-ci please test macOS

briancroom:
@swift-ci please test

PR #198, opened 2017-08-15
[SR-5643] Adopt Swift 4 API adjustments from Apple XCTest

briancroom:
In Xcode 9, XCTest underwent a bit of a "Swiftification" effort to adopt more modern Swift idioms and improve how its APIs get presented in Swift 4. This PR adopts those changes here as well. They include:

* Migrate from `UInt` -> `Int`. Prior to Swift 4, several XCTest APIs were brought into Swift using the
`UInt` type, in contrast to other Apple system frameworks which consistently used `Int`. In Swift 4, XCTest began being treated as a system framework and now uses `Int` as well.
* Make `XCTestObservationCenter.shared` a class property instead of class function.
* Add an `XCTestError` struct. This is to match the struct generated by the Objective-C importer for
Apple XCTest due to its `XCTestErrorCode` having been marked with the `NSErrorDomain` API note.
* Add `XCTPerformanceMetric` string enum. In Xcode 9, the `XCTPerformanceMetric_WallClockTime` string constant was turned into a string enumeration to improve the ability to use strong typing
at call sites using it. At the same time, the `defaultPerformanceMetrics` class function on `XCTestCase` was changed to be a class property.

### DIFF
--- a/Sources/XCTest/Private/PerformanceMeter.swift
+++ b/Sources/XCTest/Private/PerformanceMeter.swift
@@ -40,7 +40,7 @@ internal protocol PerformanceMeterDelegate {
     ///   as average, and standard deviation
     /// - Parameter file: The source file name where the measurement was invoked
     /// - Parameter line: The source line number where the measurement was invoked
-    func recordMeasurements(results: String, file: StaticString, line: UInt)
+    func recordMeasurements(results: String, file: StaticString, line: Int)
 
     /// Reports a test failure from the analysis of performance measurements.
     /// This can currently be caused by an unexpectedly large standard deviation
@@ -48,14 +48,14 @@ internal protocol PerformanceMeterDelegate {
     /// - Parameter description: An explanation of the failure
     /// - Parameter file: The source file name where the measurement was invoked
     /// - Parameter line: The source line number where the measurement was invoked
-    func recordFailure(description: String, file: StaticString, line: UInt)
+    func recordFailure(description: String, file: StaticString, line: Int)
 
     /// Reports a misuse of the `PerformanceMeter` API, such as calling `
     /// startMeasuring` multiple times.
     /// - Parameter description: An explanation of the misuse
     /// - Parameter file: The source file name where the misuse occurred
     /// - Parameter line: The source line number where the misuse occurred
-    func recordAPIViolation(description: String, file: StaticString, line: UInt)
+    func recordAPIViolation(description: String, file: StaticString, line: Int)
 }
 
 /// - Bug: This class is intended to be `internal` but is public to work around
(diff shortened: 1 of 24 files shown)


briancroom:
@swift-ci please test

briancroom:
@parkera I used corelibs-foundation as a model for adding a couple of types here (`XCTestError` and `XCTPerformanceMetric`) that are generated by the clang importer on Darwin. I just wanted to check that what I did here seems sensible to you.

briancroom:
@swift-ci please test

briancroom:
I updated this to no longer make the `UInt` -> `Int` change for the assertion functions. On Darwin, these functions come from the Swift XCTest overlay, rather than XCTest.framework, so the clang importer change due to making XCTest a system framework didn't affect them.

briancroom:
@swift-ci please test

briancroom:
@czechboy0 could you take a look here when you get a chance? Thanks!

Sources/XCTest/Private/PerformanceMeter.swift
```diff
@@ -97,16 +97,16 @@ public final class PerformanceMeter {
...
-    private init(metrics: [PerformanceMetric], delegate: PerformanceMeterDelegate, file: StaticString, line: UInt) {
+    private init(metrics: [PerformanceMetric], delegate: PerformanceMeterDelegate, file: StaticString, line: Int) {
         self.metrics = metrics
         self.delegate = delegate
         self.invocationFile = file
         self.invocationLine = line
     }
 
-    static func measureMetrics(_ metricNames: [String], delegate: PerformanceMeterDelegate, file: StaticString = #file, line: UInt = #line, for block: (PerformanceMeter) -> Void) {
+    static func measureMetrics(_ metricNames: [String], delegate: PerformanceMeterDelegate, file: StaticString = #file, line: Int = #line, for block: (PerformanceMeter) -> Void) {
```
czechboy0: Why not make `metricNames: [PerformanceMetric]` too? As they're strings, we have to map from `PerformanceMetric` to `String` down below.

Sources/XCTest/Public/XCTestCase+Performance.swift
```diff
@@ -103,12 +125,12 @@ public extension XCTestCase {
...
     ///   ensure compatibility of tests between swift-corelibs-xctest and Apple
     ///   XCTest, it is not recommended to pass explicit values for `file` and `line`.
-    func measureMetrics(_ metrics: [String], automaticallyStartMeasuring: Bool, file: StaticString = #file, line: UInt = #line, for block: () -> Void) {
+    func measureMetrics(_ metrics: [XCTPerformanceMetric], automaticallyStartMeasuring: Bool, file: StaticString = #file, line: Int = #line, for block: () -> Void) {
         guard _performanceMeter == nil else {
             return recordAPIViolation(description: "Can only record one set of metrics per test method.", file: file, line: line)
         }
 
-        PerformanceMeter.measureMetrics(metrics, delegate: self, file: file, line: line) { meter in
+        PerformanceMeter.measureMetrics(metrics.map({ $0.rawValue }), delegate: self, file: file, line: line) { meter in
```
czechboy0: We could get rid of this map if we move to `[XCTPerformanceMetric]` across the board.

Sources/XCTest/Private/PerformanceMeter.swift
briancroom: Hmm..

One of the design goals with the PerformanceMeter and WallClockTimeMetric classes was for them to be entirely decoupled from the XCTest-specific stuff. i.e. they are conceptually a lower layer upon which the XCTest API sits. Ideally they would exist in their own separate module, but it's not currently possible to do that while still keeping XCTest library product and module standalone. See https://github.com/apple/swift-corelibs-xctest/pull/109 for earlier discussion.

How do you feel about this given that context?

czechboy0: That makes sense. A possible (future) improvement could be to create a similar enum at the PerformanceMeter layer.

krzyzanowskim:
It didn't make it to the Xcode9 GM either? I mean it works when building with Xcode, but not when launch test from the command line with "swift build"

Apple Swift version 4.0 (swiftlang-900.0.65 clang-900.0.37) Target: x86_64-apple-macosx10.9
Apple Swift Package Manager - Swift 4.0.0-dev (swiftpm-13126)

```
marcinkrzyzanowski:CryptoSwift (develop)$ swift test
Compile Swift Module 'CryptoSwiftTests' (13 sources)
/Users/marcinkrzyzanowski/Devel/CryptoSwift/Tests/CryptoSwiftTests/PBKDF.swift:59:29: error: type 'XCTPerformanceMetric' has no member 'wallClockTime'
            measureMetrics([XCTPerformanceMetric.wallClockTime], automaticallyStartMeasuring: true, for: { () -> Void in
                            ^~~~~~~~~~~~~~~~~~~~ ~~~~~~~~~~~~~
```

PR #199, opened 2017-08-15
Fix String.CharacterView warnings

ianpartridge:
`Warning: 'characters' is deprecated: Please use String or Substring directly`


modocache:
Late to the party, but this seems like a nice improvement to the `String` API. Neat!

@swift-ci please test

modocache:
LGTM, and the tests pass, but I'll leave it to @briancroom to review/merge, just in case.

briancroom:
Looks good. Thanks!

PR #200, opened 2017-08-31
For testing on Android

johnno1962:
Hi Apple,

These are the changes I’ve used to build XCTest for Android. I don’t know if you have other plans for Package.swift as it was but I figure since there are no version tags on this repo it cannot be used anyway.

Cheers.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,3 +1,23 @@
+// swift-tools-version:4.0
+//
+// To build with auto-linking of the .swiftmodule use:
+// $ swift build -Xswiftc -module-link-name -Xswiftc XCTest
+//
+
 import PackageDescription
 
-let package = Package(name: "XCTest")
+let package = Package(
+    name: "XCTest",
+    products: [
+        .library(
+            name: "XCTest",
+            type: .dynamic,
+            targets: ["XCTest"]
+        )
+    ],
+    dependencies: [
+    ],
+    targets: [
+        .target(name: "XCTest", dependencies: [], path: "Sources"),
+    ]
+)
--- a/Sources/XCTest/Private/PrintObserver.swift
+++ b/Sources/XCTest/Private/PrintObserver.swift
@@ -59,7 +59,9 @@ internal class PrintObserver: XCTestObservation {
 
     fileprivate func printAndFlush(_ message: String) {
         print(message)
+        #if !os(Android)
         fflush(stdout)
+        #endif
     }
 
     private func formatTimeInterval(_ timeInterval: TimeInterval) -> String {
--- a/build_script.py
+++ b/build_script.py
@@ -200,13 +200,14 @@ class GenericUnixStrategy:
                 libdispatch_args=libdispatch_args,
                 source_paths=" ".join(sourcePaths)))
         run("{swiftc} -emit-library {build_dir}/XCTest.o "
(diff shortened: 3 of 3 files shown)


briancroom:
Hey @johnno1962, thanks for the PR!

I'd be interested in hearing more about the workflow you're using where you make use of this. In this situation, is XCTest is building against the copy of Foundation from the Swift toolchain? If so, why not take XCTest from the toolchain as well?

Sources/XCTest/Private/PrintObserver.swift
```diff
@@ -59,7 +59,9 @@ internal class PrintObserver: XCTestObservation {
 
     fileprivate func printAndFlush(_ message: String) {
         print(message)
+#if !os(Android)
```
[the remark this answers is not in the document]
johnno1962: It’s the absense of stdout which causes the compilation error.

johnno1962:
Hi @briancroom, I’m having to roll my own XCTest as I’m preparing a toolchain for Android (http://johnholdsworth.com/bothworlds.html) and I don’t believe the build scripts have been adapted for a cross compile of XCTest (I could be wrong there.) Once it’s built I copy the shared lib and .swiftmodule into the toolchain manually and package it up. I’m using the swift-robot-environment scritps for the build - https://github.com/gonzalolarralde/swifty-robot-environment.

Package.swift
```diff
@@ -1,3 +1,23 @@
...
+// To build with auto-linking of the .swiftmodule use:
+// $ swift build -Xswiftc -module-link-name -Xswiftc XCTest
+//
+
 import PackageDescription
 
-let package = Package(name: "XCTest")
+let package = Package(
+    name: "XCTest",
+    products:[
```
xwu: Nit: missing space between `:` and `[`

Package.swift
```diff
@@ -1,3 +1,23 @@
...
+        .library(
+            name: "XCTest",
+            type: .dynamic,
+            targets:["XCTest"]
```
xwu: Nit: missing space between `:` and `[`

Package.swift
```diff
@@ -1,3 +1,23 @@
...
+        )
+    ],
+    dependencies: [
+    ],
+    targets: [
+        .target(name: "XCTest", dependencies: [], path: "Sources"),
+        ]
```
xwu: Nit: de-indent `]` to match the lines above (see, for example, the closing bracket after `dependencies`)

Sources/XCTest/Private/PrintObserver.swift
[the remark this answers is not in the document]
briancroom: (Further discussion of this change [here](https://github.com/apple/swift-corelibs-xctest/pull/201#discussion_r138986199))

briancroom:
@xwu's whitespace comments are fair, and it would be good to address them before merging this, but otherwise I am fine with adding this.

In the future, it would be cool if we could get the test suite to be runnable via `swift test` as well, possibly by bridging to `lit` in some way.

briancroom:
With #201 merged, the remaining change here is just adding the Package.swift file. There's no point in running CI on that since it isn't involved in that build anyway.

johnno1962:
Thanks Brian 👍

PR #203, opened 2017-10-06
[SR-1047] Run xctest_checker tests in tests

tiagomartinho:
swift-corelibs-xctest uses a Python program called xctest_checker to run its functional test suite. That program itself has unit tests, written in Python.
These unit tests should be run as part of the greater XCTest test suite. This is probably possible by modifying the lit.cfg in the swift-corelibs-xctest project such that it picks up on a file with a RUN: directive that triggers the Python tests. See https://github.com/apple/swift/pull/778 for an example.
See llvm.org/docs/CommandGuide/lit.html for additional details on lit--it's the test runner for Swift, LLVM, and other projects, so it's nice to know about it.

### DIFF
--- a/Tests/Functional/lit.cfg
+++ b/Tests/Functional/lit.cfg
@@ -11,7 +11,7 @@
 import os
 import platform
 import tempfile
-
+import sys
 import lit
 
 # Set up lit config.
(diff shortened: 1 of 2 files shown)


modocache:
@swift-ci please test

modocache:
Also +cc @briancroom, in case he knows anyone working on Swift CI who might not want these additional tests to be run. I don't think it'll be a problem, especially considering `swift_build_support` tests are run in the same way, but who knows.

tiagomartinho:
Thank you Brian for the detailed task description 😄

modocache:

briancroom:
Cool! This makes sense to me. I don’t know of any reason not to run these in CI. Let’s get this merged once CI goes green. 

Thanks @tiagomartinho!

tiagomartinho:
Thank you Brian!

PR #204, opened 2017-10-10
Define XCTestCaseClosure typealias

tiagomartinho:
I believe the following type alias can improve the readability of the code, what do you think?

### DIFF
--- a/Sources/XCTest/Public/XCTestCase.swift
+++ b/Sources/XCTest/Public/XCTestCase.swift
@@ -11,12 +11,17 @@
 //  Base class for test cases
 //
 
+/// A block with the test code to be invoked when the test runs.
+///
+/// - Parameter testCase: the test case associated with the current test code.
+public typealias XCTestCaseClosure = (XCTestCase) throws -> Void
+
 /// This is a compound type used by `XCTMain` to represent tests to run. It combines an
 /// `XCTestCase` subclass type with the list of test case methods to invoke on the class.
 /// This type is intended to be produced by the `testCase` helper function.
 /// - seealso: `testCase`
 /// - seealso: `XCTMain`
-public typealias XCTestCaseEntry = (testCaseClass: XCTestCase.Type, allTests: [(String, (XCTestCase) throws -> Void)])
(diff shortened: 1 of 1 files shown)


modocache:
I like it! Could we brainstorm some additional names, tough? In the context of the XCTest framework, `XCTestClosure` could refer to any closure, so I feel it's ambiguous. Maybe `XCTTestMethodClosure`? Or something else?

tiagomartinho:
Of course!

I choose XCTestClosure because I tend to favor shorter names. Initially I was thinking in XCTestCaseClosure, but I also like XCTestMethodClosure. 

Let me know which one you prefer and I will make the change.

modocache:
Personally I like `XCTestCaseClosure` -- in theory these don't have to be methods, after all.

tiagomartinho:
@modocache I changed it to `XCTestCaseClosure`

Sources/XCTest/Public/XCTestCase.swift
```diff
@@ -16,7 +16,9 @@
 /// This type is intended to be produced by the `testCase` helper function.
 /// - seealso: `testCase`
 /// - seealso: `XCTMain`
-public typealias XCTestCaseEntry = (testCaseClass: XCTestCase.Type, allTests: [(String, (XCTestCase) throws -> Void)])
+public typealias XCTestCaseEntry = (testCaseClass: XCTestCase.Type, allTests: [(String, XCTestCaseClosure)])
+
+public typealias XCTestCaseClosure = (XCTestCase) throws -> Void
```
modocache: Sorry, a bit of a nit-pick: how about moving this `typealias` above the first, so that readers don't encounter `XCTestCaseClosure` before it's defined?

Also, this typealias could use some documentation.

tiagomartinho: I totally agree!

Let me know if the documentation is helpful or it needs improvements.

modocache:
@swift-ci please smoke test and merge

modocache:
@swift-ci please test

tiagomartinho:
There was some problem with the merge?

modocache:
@swift-ci Please test

briancroom:
Thanks guys, this does look good!

PR #205, opened 2017-10-21
[SR-906] Allow continueAfterFailure to be set

tiagomartinho:
[After this exchange of ideas](https://bugs.swift.org/browse/SR-906), my opinion is that for the moment we can add the feature for setting continueAfterFailure, and after that write a new task that addresses the fact that all tests stop if a failure occurs and continueAfterFailure is false.

I was thinking in something like:
[SR-XXX] Allow remaining tests to continue after a failure instead of stopping the test process
"Apple XCTest does not throw a fatal error and crash the test process, it merely prevents the remainder of a testClosure from expecting after it's been determined that it has already failed. The following behavior is incorrect."

And migrating all the conclusions we got to this new task.

What do you think?

### DIFF
--- a/Tests/Functional/ContinueAfterFailureTestCase/main.swift
+++ b/Tests/Functional/ContinueAfterFailureTestCase/main.swift
@@ -1,0 +1,62 @@
+// RUN: %{swiftc} %s -o %T/ContinueAfterFailureTestCase
+// RUN: %T/ContinueAfterFailureTestCase > %t || true
+// RUN: %{xctest_checker} %t %s
+
+#if os(macOS)
+    import SwiftXCTest
+#else
+    import XCTest
+#endif
+
+// CHECK: Test Suite 'All tests' started at \d+-\d+-\d+ \d+:\d+:\d+\.\d+
+// CHECK: Test Suite '.*\.xctest' started at \d+-\d+-\d+ \d+:\d+:\d+\.\d+
+
+// CHECK: Test Suite 'ContinueAfterFailureTestCase' started at \d+-\d+-\d+ \d+:\d+:\d+\.\d+
+class ContinueAfterFailureTestCase: XCTestCase {
+    static var allTests = {
+        return [
+            ("testDoesNotContinueAfterFailure", testDoesNotContinueAfterFailure),
+            ("testContinueAfterFailure", testContinueAfterFailure)
+        ]
+    }()
+
+    override func tearDown() {
+        super.tearDown()
+        print("In \(#function)")
+    }
+
+    // CHECK: Test Case 'ContinueAfterFailureTestCase.testDoesNotContinueAfterFailure' started at \d+-\d+-\d+ \d+:\d+:\d+\.\d+
+    // CHECK: First Log in DoesNotContinueAfterFailure test
+    // CHECK: .*/ContinueAfterFailureTestCase/main.swift:[[@LINE+8]]: error: ContinueAfterFailureTestCase.testDoesNotContinueAfterFailure : XCTAssertTrue failed -
+    // CHECK: In tearDown\(\)
+    // CHECK: Terminating execution due to test failure
+    // CHECK-NOT: Second Log in DoesNotContinueAfterFailure test
+    // CHECK-NOT: Test Case 'ContinueAfterFailureTestCase.testDoesNotContinueAfterFailure' failed \(\d+\.\d+ seconds\)
+    func testDoesNotContinueAfterFailure() {
+        continueAfterFailure = false
+        print("First Log in DoesNotContinueAfterFailure test")
+        XCTAssert(false)
+        print("Second Log in DoesNotContinueAfterFailure test")
+    }
+    
+    // CHECK-NOT: Test Case 'ContinueAfterFailureTestCase.testContinueAfterFailure' started at \d+-\d+-\d+ \d+:\d+:\d+\.\d+
+    // CHECK-NOT: First Log in ContinueAfterFailure test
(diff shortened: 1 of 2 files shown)


tiagomartinho:
@modocache do you think it makes sense to follow this idea? Or should we just close the pull request?

Tests/Functional/ContinueAfterFailureTestCase/main.swift
```diff
@@ -0,0 +1,41 @@
...
+    static var allTests = {
+        return [
+            ("testDoesNotContinueAfterFailure", testDoesNotContinueAfterFailure),
+            ("testContinueAfterFailure", testContinueAfterFailure)
+        ]
+    }()
+    
+    // CHECK: Test Case 'ContinueAfterFailureTestCase.testDoesNotContinueAfterFailure' started at \d+-\d+-\d+ \d+:\d+:\d+\.\d+
+    // CHECK: First Log in DoesNotContinueAfterFailure test
+    // CHECK: .*/ContinueAfterFailureTestCase/main.swift:[[@LINE+4]]: error: ContinueAfterFailureTestCase.testDoesNotContinueAfterFailure : XCTAssertTrue failed -
```
modocache: You should include a `CHECK-NOT` for "Second log" here, to verify that it's not being printed.

Tests/Functional/ContinueAfterFailureTestCase/main.swift
```diff
@@ -0,0 +1,41 @@
...
+    func testDoesNotContinueAfterFailure() {
+        continueAfterFailure = false
+        print("First Log in DoesNotContinueAfterFailure test")
+        XCTAssert(false)
+        print("Second Log in DoesNotContinueAfterFailure test")
+    }
+    
+    func testContinueAfterFailure() {
```
modocache: There don't seem to be any assertions for this function, did you forget to add them? What's being tested here?

Tests/Functional/ContinueAfterFailureTestCase/main.swift
tiagomartinho: Ok I will add it! I'm not very familiar with lit

Tests/Functional/ContinueAfterFailureTestCase/main.swift
tiagomartinho: I will add also the CHECK-NOT to this part (thanks for the tip), because with the current behavior all the test process crashes and it doesn't print anything for the second test (testContinueAfterFailure). What's being tested right now is that the remaining of the test suite doesn't execute.

tiagomartinho:
@modocache I added the missing assertions to the integration test.

When the implementation of *continueAfterFailure* matches XCTest, we have to remove all the *-NOT* from *CHECK-NOT* except for this one: 
`// CHECK-NOT: Second Log in DoesNotContinueAfterFailure test`

Does it make sense?

briancroom:
Hey guys! Sorry for not getting involved here sooner.

This is indeed a tricky problem. The implementation of XCTest that ships with Xcode does use Objective-C exceptions to implement the behavior of interrupting test execution after a failure, and no similar construct is available in Swift at this point. As discussed on the bug, aborting the process with a `fatalError`, or similar, does seem like the only way to achieve this sort of interruption behavior.

I *am* a little concerned, though, about dropping in this change without having a mechanism to pick up execution with the next test, since folks could end up with their suites not running to completion, and not realizing it.

There are a couple of things that I could imagine us doing here which might mitigate the problem:
* We should make sure that we are clearly reporting that the test suite is exiting early. From the `CHECK` lines in the new test, it appears that the string being passed to the `fatalError` call may not be making it to stdout/stderr? If that's the case, we should add extra printing.
* If a test running under Apple XCTest is interrupted due to a failure, the test harness will still run the test's `-tearDown` method, which can be important for e.g. cleaning up files that the test might have written out. I wonder if it would be reasonable to run through the tear down sequence before aborting the process?

briancroom:
(Also, I'm going to be on vacation for a while now so don't expect any speedy responses from me!)

briancroom:
This could be interesting for @aciidb0mb3r as well. If we do continue down the path of aborting the test runner process after a failure, it may be that SwiftPM would be a sensible place to perform the overall process management work, and respawn the runner with a new, reduced set of tests to run.

tiagomartinho:
@briancroom I added the extra printing, and also run the **tearDown** before throwing the fatal error.

What about changing the API of XCTest to achieve this sort of interruption behaviour in Swift?
```
func testExample() throws {
    try XCTAssert(true)
}
```

Also if you believe this change does not meet the requirements and expectations we can always drop it.

tiagomartinho:
Hello everyone, any news?

Thank you!

nikita-leonov:
@tiagomartinho @briancroom @modocache is this PR was abandoned for a reason? It will be great to have ‘continueAfterFailure’ work and this PR solves the problem. Not sure how can I help here to get it merged.

tiagomartinho:
Hello @nikita-leonov!

If I remember correctly (it was a long time ago 😄). This PR doesn't fully address the current behaviour of XCTest used by Xcode. 

If `continueAfterFailure` is set it should stop the test suite execution from the current test, gathering the results and stopping the execution gracefully.

Now I believe it just runs the `tearDown` method and aborts the execution. 

I proposed before to change the API to handle this scenario in Swift, but I don't believe it's the best solution and worth the effort to replicate this behaviour.

tiagomartinho:
Should I close the PR?

shahmishal:

PR #208, opened 2017-11-28
Remove XCTAssert(Not)Equal overloads made obsolete by conditional conformances.

DougGregor:
The various array types, Optional, and Dictionary are now Equatable when
their element types are Equatable, so we can remove the overloads meant to
support those use cases. Fixes rdar://problem/17924430.


DougGregor:
@swift-ci please smoke test

PR #213, opened 2018-03-15
SR-2100 - Add support for multi-lib systems

capent:
[SR-2100 - Add support for multi-lib systems](https://bugs.swift.org/browse/SR-2100)

Create a new "libdir" build option for the parser in build-script.py, so Linux distributions
that uses lib64 folder instead of lib for 64-bit libraries, like Fedora, could install Swift
without the need for patches.


modocache:
@swift-ci please test

compnerd:
Note that with #214 this shouldn't be needed as CMake can handle the lib vs lib64 structuring.

capent:
@compnerd That's true! But it is not so automatic. I will take a peek at #214 and test it here to see how it behaves. If it is ok, i believe that going toward cmake is a better option than this one to handle multi-lib systems on Linux and i will close this pull request.

shahmishal:

PR #214, opened 2018-05-25
build: add CMake based build system for XCTest

compnerd:


compnerd:
CC: @modocache

compnerd:
Okay, the newest version has been tested against s-c-f and passes tests as well.  It has testing support too via the `check-xctest` target.  I think that this is pretty much complete at this point.

CMakeLists.txt
```diff
@@ -0,0 +1,122 @@
...
+option(XCTEST_PATH_TO_LIBDISPATCH_SOURCE "Path to libdispatch source" "")
+option(XCTEST_PATH_TO_LIBDISPATCH_BUILD "Path to libdispatch build" "")
+
+option(XCTEST_PATH_TO_FOUNDATION_BUILD "Path to Foundation build" "")
+
+option(XCTEST_PATH_TO_COREFOUNDATION_BUILD "Path to CoreFoundation build" "")
+
+find_package(LLVM CONFIG)
+if(NOT LLVM_FOUND)
+  message(SEND_ERROR "Could not find LLVM; configure with -DCMAKE_PREFIX_PATH=/path/to/llvm/install")
```
modocache: Question for you: my understanding is that there's also a method of configuration in which the `llvm-config` executable is invoked, is that correct? But I believe I heard that means of configuration is deprecated -- I'm not sure. Do you know the current situation is?

To be clear, I'm not asking you to change anything here, this looks right to me, I'm just wondering about `llvm-config`.

CMakeLists.txt
```diff
@@ -0,0 +1,122 @@
+
```
modocache: Tiny nit-pick/question: is there a reason you're leaving an empty line at the top of the CMake files? Same goes for the `cmake/modules/SwiftSupport.cmake` file below.

CMakeLists.txt
```diff
@@ -0,0 +1,122 @@
...
+                    Sources/XCTest/Public/XCTestObservationCenter.swift
+                    Sources/XCTest/Public/XCTestCase+Performance.swift
+                    Sources/XCTest/Public/XCTAssert.swift
+                    Sources/XCTest/Public/Asynchronous/XCTestCase+NotificationExpectation.swift
+                    Sources/XCTest/Public/Asynchronous/XCTestCase+PredicateExpectation.swift
+                    Sources/XCTest/Public/Asynchronous/XCPredicateExpectationHandler.swift
+                    Sources/XCTest/Public/Asynchronous/XCWaitCompletionHandler.swift
+                    Sources/XCTest/Public/Asynchronous/XCNotificationExpectationHandler.swift
+                    Sources/XCTest/Public/Asynchronous/XCTestCase+Asynchronous.swift
+                    Sources/XCTest/Public/Asynchronous/XCTestExpectation.swift
```
modocache: (I vaguely remembered that globbing source files in CMake was discouraged, so for any other reviewers out there, I  looked up the exact reason: https://stackoverflow.com/a/32412044/679254)

cmake/modules/SwiftSupport.cmake
```diff
@@ -0,0 +1,133 @@
+
+include(CMakeParseArguments)
+
+function(add_swift_target target)
```
modocache: This seems like a handy, generic function. Are you anticipating moving it somewhere where all the swift-corelibs-* projects could use it, maybe someday?

aciidb0mb3r: [swift-tools-support-core](https://github.com/apple/swift-tools-support-core) repository is not ready yet but it can be used to hold the shared infrastructure once its open for business!

CMakeLists.txt
compnerd: This is the more modern, CMake way to handle this.  LLVM will install a CMake module that can be used.  If you use the `llvm-config` invocation approach, that requires the user to specify the path to the installation.  If you install LLVM to the system normally, this should JustWork(TM) so it is nicer.

CMakeLists.txt
compnerd: Yeah, ideally, someone would do the hard work of figuring out how to add the authorship information to the CMake content.  I've been lazy and this is something that should be done at some point - it is not for any functional reason, only so that there is uniformity in all the files across the repositories.

cmake/modules/SwiftSupport.cmake
compnerd: @modocache - you are right, this is a handy, generic function.  I use the same basic functionality in swift-corelibs-libdispatch and am using it in swift-corelibs-foundation too.  Once we have the shared repository, moving this would be great.

compnerd:
@briancroom - ping!

briancroom:
@compnerd hey, thanks for the ping! I'm sorry I haven't had a chance to take a proper look here. I'm pretty swamped with WWDC responsibilities right now still but will really try to make some time in the next day or two for this. Thank you for your patience, and for the PR!

compnerd:
@briancroom - ping!

compnerd:
Sure.  Yes, the desire is to deprecate the python build script infrastructure in favor of this.  CMake would also be able to generate an Xcode project if you so desired.  This will create a single unified build system that can target all the platforms.  It also makes it easier to cross-compile XCTest (for say Windows).  I think that switching over the build-script to CMake would be the ideal thing, but that would be a secondary change as we need to ensure that XCTest is ready to be built with CMake first.

It also aids in reducing some of the cognitive overload with different build systems for the different pieces of swift.

briancroom:
Sounds good, thanks!

PR #215, opened 2018-06-13
build: add compatibility with the build_script.py build

compnerd:
This allows us to switch to CMake for the builds of XCTest on non-Darwin
platforms earlier.  This will get us additional testing and ensure
coverage.


compnerd:
@swift-ci please test

compnerd:
@modocache - yeah, that is entirely correct

PR #217, opened 2018-06-22
build: match hardcoded extensions in the compiler

compnerd:
The swiftdoc partials must have the extension .swiftdoc.  Adjust the
output names to match this assumption.  This is needed to repair the
CMake build with the newer compiler.


compnerd:
@swift-ci please test

compnerd:
CC: @modocache @aciidb0mb3r

compnerd:
I can point out where the assumption lies, I think that I had an older swiftc that I was building with and that recent unrelated changes now exposed the bug that had existed in the documentation building.

PR #218, opened 2018-08-16
Fix redundant access-level modifiers.

dingobye:
This patch aims to make `corelibs-xctest` adapted to [this PR]( https://github.com/apple/swift/pull/18623), which produces warnings for redundant access-level modifiers.

For the code pattern shown in the example below, there are two feasible ways to fix the redundancy issue:
```
public extension {            // (1) remove `public` modifier for extension;
   public func foo() {}       // or (2) remove `public` modifier for member.
}
```
This patch goes with (2), keeping consistency with the rest of the project.


stmontgomery:
@swift-ci Please test

dingobye:
Thank you for firing the test, Stuart! Looks like everything goes well. It would be greatly appreciated if you could review and help to land this PR.

dingobye:
Thank you, Stuart!

PR #222, opened 2018-09-06
In test output, unconditionalize parens and unnamed parameter.

rudkx:
Now that https://github.com/apple/swift/pull/19062 has landed (which
makes the output generated for #function more consistent), remove the
conditional matching that was done in
https://github.com/apple/swift-corelibs-xctest/pull/221.


rudkx:
@swift-ci Please test

PR #223, opened 2018-09-19
build: correct the OS directory on install

compnerd:
Convert the OS name to lower case as is the convention.  This installs
the content to the right location to permit use from an installed
package.


compnerd:

PR #225, opened 2018-09-28
Non-Darwin SDKs product results in per-arch directories.

drodriguez:
Depends on apple/swift#19432

Unix (other than Darwin) and Windows do not support fat binaries.
However, the build system was placing the build results in the
platform/OS folder, instead of the per-architecture folder. Having two
architectures side by side was impossible.

After applying the above patch in the Swift compiler, non-Darwin
platforms will look into the per-architecture folders, so the sibling
projects need to leave the products in the right place.


CMakeLists.txt
```diff
@@ -148,4 +148,9 @@ install(FILES
           ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_SHARED_LIBRARY_PREFIX}XCTest${CMAKE_SHARED_LIBRARY_SUFFIX}
         DESTINATION
           ${CMAKE_INSTALL_FULL_LIBDIR}/swift/${swift_os})
-
+if(NOT CMAKE_SYSTEM_NAME STREQUAL Darwin)
```
kevints: It looks like you're now installing two distinct copies of libXCTest.so onto the system here? Would it be possible to use a symlink instead?

drodriguez: Do not merge or take time on this merge request for the moment. I have to review the approach.

There is two copies. This change was a change compatible with the current setup and the setup from apple/swift#19432, which moves the binary libraries into the architecture directories for platforms that do not support fat libraries (every platform except Darwin). The duplication allowed me to merge this while working in the other change, and keep things working.

The symlink approach would work for some platforms, but not others (Windows), so I decided for a simpler copy, since this was going to be a temporal solution.

PR #227, opened 2018-10-23
build: synchronize SwiftSupport with Foundation

compnerd:
Synchronize the SwiftSupport CMake module with Foundation with support
for Resources and correcting the dependency tracking.


compnerd:

PR #229, opened 2018-10-28
Build Fixes

compnerd:
A couple of minor fixes for the build system.  Pass `-target` to the swift driver when linking to enable cross-linking.  Fix the install location for XCTest to use the right architecture spelling as libdispatch and swift-corelibs-foundation do.


briancroom:

PR #237, opened 2018-11-15
Call closures passed as @autoclosure explicitly.

tkremenek:
Needed for Swift 5 migration.


tkremenek:
Thanks @briancroom.  After I started building with this patch I started seeing these problems.  I'll tinker and adjust to match the original semantics.

tkremenek:
I realized these changes make no sense.  I'm going to go back and re-analyze why I saw the errors I did.

tkremenek:
I think the errors I was seeing was due to regressions in the compiler due to the handling of `@autoclosure`.  I should have examined them more closely before creating this PR.

PR #238, opened 2018-11-16
Force corelibs-xctest to build with -swift-version 4

tkremenek:
There may be a regression in the compiler's handling of `@autoclosure`, which prevents
xctest building in `-swift-version 5`.  Temporarily hold it to `-swift-version 4`.


tkremenek:
This appears to workaround the compiler failure when moving the default language mode to Swift 5.  This should be reverted later.

slavapestov:
@tkremenek It's not a regression, it was #20152. The old `@autoclosure` behavior was a bit quirky because we modeled it as a property of a type and not of a parameter.

PR #247, opened 2019-02-26
build: add some workarounds for Windows

compnerd:
This is sufficient to allow us to build for Windows.  This is masking
over underlying deficiencies in the way that we currently build, but
that should not prevent us from using XCTest on Windows for the time
being.


compnerd:

PR #253, opened 2019-02-27
XCTest: use Foundation SPI on Windows

compnerd:
Foundation on Windows does not re-expose the CoreFoundation symbols.  In
order to permit XCTest to continue working with minimum changes, it
exposes a SPI.  Switch to that SPI for Windows.  This allows us to build
XCTest on Windows.

### DIFF
--- a/Sources/XCTest/Public/Asynchronous/XCTWaiter.swift
+++ b/Sources/XCTest/Public/Asynchronous/XCTWaiter.swift
@@ -367,7 +367,11 @@ private extension XCTWaiter {
 
     func cancelPrimitiveWait() {
         guard let runLoop = runLoop else { return }
+#if os(Windows)
+        runLoop._stop()
+#else
         CFRunLoopStop(runLoop.getCFRunLoop())
+#endif
     }
 }
 


Sources/XCTest/Public/Asynchronous/XCTWaiter.swift
```diff
@@ -217,9 +217,17 @@ open class XCTWaiter {
         self.manager = manager
 
         // Begin the core wait loop.
+#if os(Windows)
+        let timeoutTimestamp = NSDate._currentAbsoluteTime() + timeout
```
spevans: Isnt `Date.timeIntervalSinceReferenceDate` the same as `CFAbsoluteTimeGetCurrent()` so I dont think there is a need for this SPI or the Windows/NonWindows differences, it can just be:

`let timeoutTimestamp = Date.timeIntervalSinceReferenceDate + timeout`

Sources/XCTest/Public/Asynchronous/XCTWaiter.swift
```diff
@@ -217,9 +217,17 @@ open class XCTWaiter {
...
+#else
         let timeoutTimestamp = CFAbsoluteTimeGetCurrent() + timeout
+#endif
         while !isFinished {
+#if os(Windows)
+            let remaining = timeoutTimestamp - NSDate._currentAbsoluteTime()
+#else
             let remaining = timeoutTimestamp - CFAbsoluteTimeGetCurrent()
```
spevans: `let remaining = timeoutTimestamp - Date.timeIntervalSinceReferenceDate`

compnerd: lol, that is way better, and allows us to kill a SPI, I like it!

PR #255, opened 2019-02-28
Use `Date.timeIntervalSinceReferenceDate`

compnerd:
Use `Date.timeIntervalSinceReferenceDate` rather than
`CFAbsoluteTimeGetCurrent` to avoid the use of CoreFoundation.  This is
in preparation for moving CoreFoundation interfaces out of Foundation.
It also makes porting XCTest easier for other targets.

### DIFF
--- a/Sources/XCTest/Public/Asynchronous/XCTWaiter.swift
+++ b/Sources/XCTest/Public/Asynchronous/XCTWaiter.swift
@@ -217,9 +217,9 @@ open class XCTWaiter {
         self.manager = manager
 
         // Begin the core wait loop.
-        let timeoutTimestamp = CFAbsoluteTimeGetCurrent() + timeout
+        let timeoutTimestamp = Date.timeIntervalSinceReferenceDate + timeout
         while !isFinished {
-            let remaining = timeoutTimestamp - CFAbsoluteTimeGetCurrent()
+            let remaining = timeoutTimestamp - Date.timeIntervalSinceReferenceDate
             if remaining <= 0 {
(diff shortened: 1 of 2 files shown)


compnerd:
@swift-ci please test

Sources/XCTest/Public/Asynchronous/XCTWaiter.swift
```diff
@@ -217,9 +217,9 @@ open class XCTWaiter {
         self.manager = manager
 
         // Begin the core wait loop.
-        let timeoutTimestamp = CFAbsoluteTimeGetCurrent() + timeout
+        let timeoutTimestamp = Date.timeIntervalSinceReferenceDate + timeout
```
briancroom: Alternatively (and perhaps more idiomatically), this could do something like `let timeoutDate = Date(timeIntervalSinceNow: timeout)`, and then below inspect the result of `timeoutDate. timeIntervalSinceNow`.

briancroom: The same idea could apply to the cases below in the tests as well - working in terms of `Date` rather than `TimeInterval`. It's not critical, though, and I would be ok with this staying as-is.

compnerd:

Sources/XCTest/Public/Asynchronous/XCTWaiter.swift
compnerd: I may try to do that as a follow up.

PR #256, opened 2019-03-05
Tests: adjust test for Windows release build

compnerd:
Running the test in release mode on Windows, the block would completely
get optimized out and the resulting time slice was 0 per measurement.
This would result in the relative standard deviation being undefined,
which would render as `nan(ind)%` failing the expectation.  Add a sleep
to ensure that the block has something to execute.


compnerd:
@swift-ci please test

briancroom:
Interesting! I wonder why that behavior hasn’t occurred on other platforms as well.

compnerd:
Yeah, I agree, it is interesting that we haven't seen this previously.  It could also be a timing thing on Windows (the timers have a slightly larger granularity - primarily for power reasons), as well as the differences in the calling convention cause a different set of optimizations to apply and expose weird issues.

compnerd:
@briancroom @stmontgomery @modocache - okay to merge?

stmontgomery:
Does the test just need _something_ to execute so that the measure block is not empty? Or does it specifically need to be a sleep statement? Could we get away with something trivial which does not slow down test execution by sleeping? I know the 1-second sleep probably won't have much noticeable impact in run time of the test suite, but curious if it has to sleep in the first place.

compnerd:
Yeah, it really is just that it needs *something* to execute, I figured that the sleep was the simplest thing.

compnerd:
@stmontgomery ping?

stmontgomery:
@compnerd So sorry for delay… This is probably fine to merge, I was just concerned about unnecessarily slowing down test execution. I wonder if it could be something trivial but which does not actually take 1 second to execute, especially since this is a performance test so it will be executed 10 times = 10 seconds.

Alternatively, just reducing the time interval from `1` to `0.01` would seem better IMO.

compnerd:
I'll come back around to this, and see if I can get this to work reliably with `0.1`.  If so, I can reduce this timeout in a follow up :-).  In the mean time, this will get XCTest passing the test suite on Windows!

PR #258, opened 2019-03-14
CMake: Use the Swift Driver and Response Files

gmittert:
Rather than rolling our own dependencies tracking, use the Swift driver
to handle dependencies and dispatching jobs. Additionally use response
files in the compilation to prevent extremely long command lines.

As requested to be copied from https://github.com/apple/swift-corelibs-foundation/pull/1959

### DIFF
--- a/cmake/modules/SwiftSupport.cmake
+++ b/cmake/modules/SwiftSupport.cmake
@@ -1,5 +1,37 @@
 
 include(CMakeParseArguments)
+include(ProcessorCount)
+
+# Creates an output file map give a target and its list of sources at
+# output_path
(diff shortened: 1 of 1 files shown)


jmittert:
@compnerd

gmittert:
Rebased and now pass in -j to the swift driver.

@swift-ci please test

gmittert:
@swift-ci please test linux platform

compnerd:
Similar to the libdispatch situation, I think that we should opt for the proper cmake support for this instead.  Thanks for doing this work though, would have been good otherwise.

PR #259, opened 2019-03-21
[SR-10034] test_outerWaiterTimesOut_InnerWaitersAreInterrupted fails non-deterministically

stmontgomery:
This is an attempt to fix this XCTest test case which seems to fail non-deterministically
on non-Darwin platforms. I made two changes:

- More closely matched the structure of the ObjC/Darwin test case which this test was modeled
  after by wrapping the core test inside a dispatch to a global queue. Inside this
  dispatched block, I changed `RunLoop.main.perform` to `RunLoop.current.perform` since the
  run loop inside that block should no longer be main but the thread's current run loop.
- Increased the timeouts of both the outer waiter and the two inner waiters by an order of
  magnitude. When this test fails it often occurs on heavily contended CI machines, and
  also the current timeouts were short enough that they may fall below the "slop" thresholds
  within XCTWaiter, so raising these timeouts generally should also help ensure that the
  ordering of the XCTWaiterDelegate callbacks is deterministic.

Also addresses rdar://problem/47482364

### DIFF
--- a/Tests/Functional/Asynchronous/Expectations/main.swift
+++ b/Tests/Functional/Asynchronous/Expectations/main.swift
@@ -341,36 +341,46 @@ class ExpectationsTestCase: XCTestCase {
     // PRAGMA MARK: - Interrupted Waiters
 
 // CHECK: Test Case 'ExpectationsTestCase.test_outerWaiterTimesOut_InnerWaitersAreInterrupted' started at \d+-\d+-\d+ \d+:\d+:\d+\.\d+
-// CHECK: .*[/\\]Tests[/\\]Functional[/\\]Asynchronous[/\\]Expectations[/\\]main.swift:[[@LINE+22]]: error: ExpectationsTestCase.test_outerWaiterTimesOut_InnerWaitersAreInterrupted : Asynchronous wait failed - Exceeded timeout of 0.1 seconds, with unfulfilled expectations: outer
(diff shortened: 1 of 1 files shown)


stmontgomery:
@swift-ci Please test

Tests/Functional/Asynchronous/Expectations/main.swift
```diff
@@ -341,36 +341,46 @@ class ExpectationsTestCase: XCTestCase {
...
+
+            let start = Date.timeIntervalSinceReferenceDate
+            let outerTimeout = 2.0
+            let result = outerWaiter.wait(for: [outerExpectation], timeout: outerTimeout)
+            let durationOfOuterWait = Date.timeIntervalSinceReferenceDate - start
+            XCTAssertEqual(result, .timedOut)
+
+            // The theoretical best-case duration in the current implementation is:
+            //     `outerTimeout` (2.0) + `outerTimeoutSlop` (0.25) = 2.25
+            // Adding some leeway for other system activity brings us to this value for the assertion.
```
compnerd: These comments are invaluable, particularly when you come back!

stmontgomery:
Discarding this because in further discussion we don't believe this will address the underlying issue. I am going to disable this test in a separate PR shortly and have a Radar tracking re-enabling it after fixing the underlying issue (which we now have a lead on, but don't yet have a fix for).

PR #261, opened 2019-03-24
build: adjust for new build layout

compnerd:
Moving CoreFoundation into the root is needed to actually get the proper
dependency tracking for Foundation builds.  This adjusts the use of it
in XCTest.


compnerd:

compnerd:
@swift-ci please test

PR #268, opened 2019-05-11
Allow running XCTest bundle with multiple comma-separated tests

egorzhdan:
This improves consistency with macOS version of XCTest which supports such behavior.

Right now any tool which runs a subset of tests is required to invoke the XCTest bundle N times, which forces the tool to spawn a lot of processes, and part of the work will be performed over and over again. This is especially painful if any expensive initialization is required (such as starting a web server), since `class func setUp()` will be called every time a new test method is called instead of once per test case.

From user's point of view statistics in this scenario will be shown in a more human-friendly format as `Executed 100 tests, with 0 failures` instead of `Executed 1 tests, with 0 failures` printed 100 times.

### DIFF
--- a/Sources/XCTest/Private/ArgumentParser.swift
+++ b/Sources/XCTest/Private/ArgumentParser.swift
@@ -17,10 +17,10 @@ internal struct ArgumentParser {
 
     /// The basic operations that can be performed by an XCTest runner executable
     enum ExecutionMode {
-        /// Run a test or test suite, printing results to stdout and exiting with
-        /// a non-0 return code if any tests failed. The name of a test or class
-        /// may be provided to only run a subset of test cases.
-        case run(selectedTestName: String?)
+        /// Run tests or test cases, printing results to stdout and exiting with
+        /// a non-0 return code if any tests failed. The names of tests or test cases
+        /// may be provided to only run a subset of them.
+        case run(selectedTestNames: [String]?)
 
         /// The different ways that the tests can be represented when they are listed
         enum ListType {
@@ -39,9 +39,9 @@ internal struct ArgumentParser {
         /// Print Help
         case help(invalidOption: String?)
 
-        var selectedTestName: String? {
-            if case .run(let name) = self {
-                return name
+        var selectedTestNames: [String]? {
+            if case .run(let names) = self {
+                return names
             } else {
                 return nil
             }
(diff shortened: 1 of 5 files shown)


egorzhdan:
@swift-ci please test

briancroom:
Great find @egorzhdan! I forgot that Xcode’s version of the tool supported this.

Sources/XCTest/Private/ArgumentParser.swift
```diff
@@ -17,10 +17,10 @@ internal struct ArgumentParser {
...
     /// The basic operations that can be performed by an XCTest runner executable
     enum ExecutionMode {
-        /// Run a test or test suite, printing results to stdout and exiting with
-        /// a non-0 return code if any tests failed. The name of a test or class
-        /// may be provided to only run a subset of test cases.
-        case run(selectedTestName: String?)
+        /// Run tests or test cases, printing results to stdout and exiting with
+        /// a non-0 return code if any tests failed. The names of tests or test cases
+        /// may be provided to only run a subset of them.
+        case run(selectedTestNames: [String]?)
```
briancroom: I guess we should prefer to use a `Set` for this for efficiency when filtering during execution?

egorzhdan: We will iterate over this array in `TestFiltering.selectedTestFilter` anyway for mapping into `SelectedTest`s, but using a Set in the filter itself is indeed faster. Fixed.

briancroom:
@swift-ci please test

egorzhdan:
Just fixed the recently arisen conflicts, @stmontgomery could you please take a look?

spevans:
@swift-ci please test

briancroom:
@swift-ci please test

briancroom:
@swift-ci Please test

egorzhdan:
The failure on Linux is probably unrelated to this PR. I don't think the changes here could have caused a SwiftPM build script error (the build finished successfully a few days ago and I didn't make any changes since then), it's likely failing because of https://github.com/apple/swift-package-manager/pull/2397

spevans:
@swift-ci test linux

stmontgomery:
@swift-ci Please test

yeswolf:
Hi, could you please share any ETA for these changes to be included into the Swift toolchain on Swift.org? It's essential to have this fix to run separate tests and as I can see they are still not included into the latest DEV.

stmontgomery:
> Hi, could you please share any ETA for these changes to be included into the Swift toolchain on Swift.org? It's essential to have this fix to run separate tests and as I can see they are still not included into the latest DEV.

What specific toolchain are you using? This has been in master since before the 5.3 branch so it should be included in recent snapshots.

yeswolf:
@stmontgomery Sorry, it was my mistake during testing. Found it, everything works with 5.3 branch.

PR #270, opened 2019-05-14
Run tests with a private module cache

brentdax:
We currently believe that the reason Linux CI is so prone to module-loading failures is that different jobs with slightly different compilers may end up using the same system-wide module cache directories. If so, using `-module-cache-path` to point to a directory specific to the working copy should improve CI reliability.

This PR modifies Corelibs XCTest's `lit` configuration to run `swift` with a `-module-cache-path $BUILT_PRODUCTS_DIR/XCTest.dir/ModuleCache` argument, contributing to this solution.

Fixes https://bugs.swift.org/browse/SR-10665.


brentdax:
@swift-ci please test

PR #272, opened 2019-05-14
build: clean up `DEPLOYMENT_TARGET_*`

compnerd:
These flags were used by CoreFoundation whcih was exposed through Foundation.
Now that that is no longer exposed, remove the unnecessary CFLAGS.


compnerd:
@swift-ci please test

PR #274, opened 2019-05-20
build: inline `-O` computation via generator expressions (NFC)

compnerd:
This inlines the `-O` handling as a generator expression into the
add_library to modernise the CMake usage here.  This prepares us for
future CMake changes to support Swift builds properly.


compnerd:
@swift-ci please test

PR #278, opened 2019-06-18
[cmake] Update SwiftSupport.cmake to allow passing link flags

drodriguez:
Also, pass the explicit target to the Swift library creation, so one can
cross compile.

/cc @compnerd

### DIFF
--- a/cmake/modules/SwiftSupport.cmake
+++ b/cmake/modules/SwiftSupport.cmake
@@ -9,7 +9,7 @@ function(add_swift_target target)
   cmake_parse_arguments(AST "${options}" "${single_value_options}" "${multiple_value_options}" ${ARGN})
 
   set(compile_flags ${CMAKE_SWIFT_FLAGS})
-  set(link_flags)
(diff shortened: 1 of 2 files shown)


compnerd:
Lets not duplicate that.  The emission of the output map file will be done by the migration to the newer CMake.  If you want to add the explicit `TARGET`, that is fine.

drodriguez:
Removed all the changes except the essential to cross-compile.

@swift-ci please test

cmake/modules/SwiftSupport.cmake
```diff
@@ -219,7 +219,7 @@ function(add_swift_executable executable)
   add_swift_target(${executable} ${ARGN})
 endfunction()
 
-# Returns the current achitecture name in a variable
+# Returns the current architecture name in a variable
```
compnerd: I spell gud

compnerd:

compnerd:
@swift-ci please test

drodriguez:
Is there a way of fixing those OS X failures? Seems to be something with Xcode, not with the project itself.

PR #279, opened 2019-07-01
XCTUnwrap API

stmontgomery:
Adds a new API `XCTUnwrap` which asserts that an expression is non-nil,
and returns the unwrapped value. This matches a new API added in Apple's
ObjC XCTest framework in Xcode 11 Beta 1 (see [release notes](https://developer.apple.com/documentation/xcode_release_notes/xcode_11_beta_2_release_notes#3318281).)

- Adds an internal user info key `XCTestErrorUserInfoKeyShouldIgnore` which
  will cause errors not to be recorded as failures at the end of a test. This
  will make sure that errors from `XCTUnwrap` are not recorded twice (the assertion
  failure _and_ the thrown error). This can then be further extended to any errors
  that should be ignored by the framework.
- Adds `XCTUnwrap`, which either returns the unwrapped value or records a failure
  and raises an error. The error includes `XCTestErrorUserInfoKeyShouldIgnore`.
- Adds unit tests for `XCTUnwrap`.

I (@stmontgomery) am posting these changes on behalf of my colleague Andrea Fernandez Buitrago, and will be serving as the primary reviewer.

rdar://problem/48810220

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,6 +70,7 @@ add_swift_library(XCTest
                     Sources/XCTest/Private/ArgumentParser.swift
                     Sources/XCTest/Private/SourceLocation.swift
                     Sources/XCTest/Private/WaiterManager.swift
+                    Sources/XCTest/Private/IgnoredErrors.swift
                     Sources/XCTest/Public/XCTestRun.swift
                     Sources/XCTest/Public/XCTestMain.swift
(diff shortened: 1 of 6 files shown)


stmontgomery:
@swift-ci Please test

stmontgomery:
@swift-ci Please test

stmontgomery:
@swift-ci Please test

PR #283, opened 2019-07-26
Implement XCTestCase.addTeardownBlock

drodriguez:
Seems that this API was missing from the open source implementation.

Includes tests.

### DIFF
--- a/Sources/XCTest/Public/XCTestCase.swift
+++ b/Sources/XCTest/Public/XCTestCase.swift
@@ -132,6 +132,7 @@ open class XCTestCase: XCTest {
                     expected: false)
             }
         }
+        runTeardownBlocks()
         tearDown()
     }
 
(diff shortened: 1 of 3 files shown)


Sources/XCTest/Public/XCTestCase.swift
```diff
@@ -178,6 +179,29 @@ open class XCTestCase: XCTest {
...
+        }
+    }
+
+    private func runTeardownBlocks() {
+        let blocks = teardownBlocksQueue.sync { () -> [() -> Void] in
+            let blocks = self.teardownBlocks
+            self.teardownBlocks = []
+            return blocks
+        }
+
```
stmontgomery: FYI, the Apple/ObjC implementation uses an internal flag to track whether or not the teardown blocks have been run yet or not, and raises an ObjC exception if a test attempts to add _more_ teardown blocks after they've already been run. This is to prevent weird cases like a test calling `addTeardownBlock(_:)` again inside a teardown block.

So you might consider matching that behavior here as well, using perhaps a `fatalError` instead of an exception.

stmontgomery:
Thanks @drodriguez!

@swift-ci Please test

Sources/XCTest/Public/XCTestCase.swift
drodriguez: The last version includes this idea and my attempt at a test that proves the usage. The test is not perfect, because the precondition text doesn't seem to be printed even if the test is crashing.

stmontgomery:
@swift-ci Please test

stmontgomery:
@drodriguez Can you rebase this on latest master to get the fix from #292? That should fix the macOS CI failure, and the Linux failure was resolved via a separate config change. So once you rebase we can re-run Swift CI and hopefully land this.

stmontgomery:
@swift-ci Please test

drodriguez:
Thanks!

PR #293, opened 2020-01-03
build: use the modern spelling for the Swift compiler (NFC)

compnerd:
This uses `CMAKE_Swift_COMPILER` which is used by CMake rather than
`CMAKE_SWIFT_COMPILER` which was used by the local Swift build system.


compnerd:
@swift-ci please test

PR #294, opened 2020-01-10
Make XCTAssertEqual with accuracy more generic

gormster:
Equality with accuracy isn't limited to floating point tests. Sometimes integer (or other numeric types) need to be accurate within certain bounds for the purposes of testing. This change allows any numeric type that has magnitude and distance to be used with `XCTAssertEqual(_:_:accuracy:)` and `XCTAssertNotEqual(_:_:accuracy:)`.

I'll be honest - I don't have a clue where in the Tests directory I'm supposed to put a regression test for this. Should I create a new test target? Seems a bit excessive, but I'm lost otherwise.

### DIFF
--- a/Sources/XCTest/Public/XCTAssert.swift
+++ b/Sources/XCTest/Public/XCTAssert.swift
@@ -171,10 +171,10 @@ public func XCTAssertEqual<T: Equatable>(_ expression1: @autoclosure () throws -
     }
 }
 
-public func XCTAssertEqual<T: FloatingPoint>(_ expression1: @autoclosure () throws -> T, _ expression2: @autoclosure () throws -> T, accuracy: T, _ message: @autoclosure () -> String = "", file: StaticString = #file, line: UInt = #line) {
+public func XCTAssertEqual<T: Strideable & Numeric>(_ expression1: @autoclosure () throws -> T, _ expression2: @autoclosure () throws -> T, accuracy: T, _ message: @autoclosure () -> String = "", file: StaticString = #file, line: UInt = #line) {
     _XCTEvaluateAssertion(.equalWithAccuracy, message: message(), file: file, line: line) {
         let (value1, value2) = (try expression1(), try expression2())
-        if abs(value1.distance(to: value2)) <= abs(accuracy.distance(to: T(0))) {
+        if abs(value1.distance(to: value2)) <= abs(accuracy.distance(to: T.zero)) {
             return .success
         } else {
             return .expectedFailure("(\"\(value1)\") is not equal to (\"\(value2)\") +/- (\"\(accuracy)\")")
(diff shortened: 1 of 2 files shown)


stmontgomery:
Thanks @gormster !

>I'll be honest - I don't have a clue where in the Tests directory I'm supposed to put a regression test for this.

I would recommend augmenting these two existing test files to add test coverage for your additions here:

```
Tests/Functional/NegativeAccuracyTestCase/main.swift
Tests/Functional/FailureMessagesTestCase/main.swift
```

gormster:
Cheers @stmontgomery. I just added some to NegativeAccuracyTestCase; seemed kind of pointless to have the exact same test in two places.

spevans:
@swift-ci test

gormster:
weird - the builds both succeeded but are both still showing as waiting for status to be reported. something up with Jenkins, or just slow?

spevans:
@swift-ci test

spevans:
Often if means the messages between jenkins and github got lost.

Sources/XCTest/Public/XCTAssert.swift
```diff
@@ -171,10 +171,10 @@ public func XCTAssertEqual<T: Equatable>(_ expression1: @autoclosure () throws -
     }
 }
 
-public func XCTAssertEqual<T: FloatingPoint>(_ expression1: @autoclosure () throws -> T, _ expression2: @autoclosure () throws -> T, accuracy: T, _ message: @autoclosure () -> String = "", file: StaticString = #file, line: UInt = #line) {
+public func XCTAssertEqual<T: Strideable & Numeric>(_ expression1: @autoclosure () throws -> T, _ expression2: @autoclosure () throws -> T, accuracy: T, _ message: @autoclosure () -> String = "", file: StaticString = #file, line: UInt = #line) {
```
stmontgomery: I've done some local experimentation with the version of this code which is included in Xcode, and I believe it's possible to simplify this code to _only_ require conformance to the `Numeric` protocol and remove the `Strideable` requirement. I think the latter is needed _currently_ because the code uses `distance(to:)` to calculate differences, but I'm not sure if that way of getting the difference is strictly necessary. If we distill the core "are two numbers equal w/accuracy" calculation down to a helper function, I think that can look something like this:

```swift
private func areEqual<T : Numeric>(_ exp1: T, _ exp2: T, accuracy: T) -> Bool {
    if exp1 == exp2 {
        return true
    } else {
        // NaN values are handled implicitly, since the <= operator returns false when comparing any value to NaN.
        let difference = (exp1.magnitude > exp2.magnitude) ? exp1 - exp2 : exp2 - exp1
        return difference.magnitude <= accuracy.magnitude
    }
}
```

And this removes the need for `distance(to:)` and anything else which currently requires `Strideable`.

Could we employ this technique in this PR too, so we can change the type requirements to only need `Numeric`? (Or, if anyone knows a strong reason to _keep_ the current implementation, let me know.) Thanks!

shahmishal:

stmontgomery:
This was closed due to the branch change noted above, but I'm landing this in #319

PR #295, opened 2020-01-30
build: do not attempt to link against dispatch/Foundation

compnerd:
On macOS, these libraries are provided by the system and are implicit.
This allows building XCTest on macOS.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,9 +46,11 @@ add_library(XCTest
   Sources/XCTest/Public/Asynchronous/XCTWaiter.swift
   Sources/XCTest/Public/Asynchronous/XCTestCase+Asynchronous.swift
   Sources/XCTest/Public/Asynchronous/XCTestExpectation.swift)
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

briancroom:
When building corelibs-xctest using the .xcworkspace in the repo, as happens in CI for this repo, it actually builds and links against corelibs-foundation instead of using the system Foundation. Is there a reason to do it differently for the CMake build?

(I don’t feel too strongly about this though, in any case, since corelibs-xctest doesn’t actually ship on macOS.)

compnerd:
libdispatch cannot be built for macOS, which means that Foundation cannot be built for macOS.  There are other projects which may benefit from this work.

compnerd:
(I'm going to merge this for now, its the next one that is more ... interesting)

briancroom:
Alright, I'm curious to see what you're working towards. We probably should converge the Xcode and CMake builds at some point, though.

PR #296, opened 2020-01-31
build: support building against Foundation.framework

compnerd:
This enables building XCTest against the System Foundation.framework on
macOS.  This requires that the executable path is computed using `URL`
(`NSURL`) as the `lastPathComponent` has been explicitly made
unavailable.

```
/Users/compnerd/SourceCache/swift-corelibs-xctest/Sources/XCTest/Public/XCTestMain.swift:99:48: error: 'lastPathComponent' is unavailable: Use lastPathComponent on URL instead.
        let exeName = CommandLine.arguments[0].lastPathComponent
                                               ^~~~~~~~~~~~~~~~~
Foundation.StringProtocol:7:16: note: 'lastPathComponent' has been explicitly marked unavailable here
    public var lastPathComponent: String { get }
               ^
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,7 @@ list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules)
 project(XCTest LANGUAGES Swift)
 
 option(BUILD_SHARED_LIBS "Build shared libraries" ON)
+option(USE_FOUNDATION_FRAMEWORK "Use Foundation.framework on Darwin" NO)
 
 if(NOT CMAKE_SYSTEM_NAME STREQUAL Darwin)
   find_package(dispatch CONFIG REQUIRED)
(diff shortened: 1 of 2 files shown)


compnerd:
@swift-ci please test

compnerd:
CC: @briancroom

So, with this work, I can actually build XCTest easily on macOS.  This has the benefit of being able to easily write tests for macOS/Linux/Windows as `XCTest` on macOS does not have `XCTMain` and `testCase` available in Swift.  There isn't a XCTest runner on Windows, and on Linux, you either need to use either swift-package-manager to generate the scaffolding or use `XCTMain`.  So, on Windows and Linux (to run from the command line) you need that.  On macOS, you need to build up the bundle if you want to use XCTest from the system.  For a little bit more work, you can have the same code run on all three platforms from the console without having to setup the bundle.

briancroom:
Ok, cool. Yeah, I think using SwiftPM as the cross-platform workflow entry point is the recommended path, but if you find it helpful/convenient to be able to use Corelibs XCTest on macOS, I don't see any particular reason to put up roadblocks for that.

PR #301, opened 2020-02-20
Expose arguments to allow overriding them.

pschuh:


Sources/XCTest/Public/XCTestMain.swift
```diff
@@ -59,10 +59,10 @@
 ///
 /// - Parameter testCases: An array of test cases run, each produced by a call to the `testCase` function
 /// - seealso: `testCase`
-public func XCTMain(_ testCases: [XCTestCaseEntry]) -> Never {
+public func XCTMain(_ testCases: [XCTestCaseEntry], arguments: [String] = CommandLine.arguments) -> Never {
     let testBundle = Bundle.main
 
-    let executionMode = ArgumentParser().executionMode
+    let executionMode = ArgumentParser(arguments: arguments).executionMode
```
stmontgomery: The `ArgumentParser` initializer has a default value for its `arguments:` parameter of `CommandLine.arguments`. Can we remove that default parameter value now, to enforce that the caller must specify it? That seems best now, given that the caller is the place where this default value is defined.

stmontgomery:
@swift-ci Please test

Sources/XCTest/Public/XCTestMain.swift
```diff
@@ -59,10 +59,10 @@
 ///
 /// - Parameter testCases: An array of test cases run, each produced by a call to the `testCase` function
 /// - seealso: `testCase`
-public func XCTMain(_ testCases: [XCTestCaseEntry]) -> Never {
+public func XCTMain(_ testCases: [XCTestCaseEntry], arguments: [String] = CommandLine.arguments) -> Never {
```
stmontgomery: Ah, just realized one more thing…

Since `XCTMain` is public API, there's a small possibility that this addition of a new parameter could break source compatibility, even though it includes a default value. If any existing code is grabs a _reference_ to this function, that reference would include all the parameters, but adding new params breaks that. (This was discussed at 22:16 of the WWDC 2019 session ["Binary Frameworks in Swift"](https://developer.apple.com/videos/play/wwdc2019/416/?time=1336).)

So I think we should modify this function's signature to not specify a default value for the `arguments:` parameter, then add an overload of the `XCTMain` function with the previous signature, and then call the 2-param overload from it, passing `CommandLine.arguments`.

stmontgomery:
@swift-ci Please test

stmontgomery:
Thanks @pschuh!

stmontgomery:
@swift-ci Please test

PR #302, opened 2020-04-23
build: use `find_package` to find the python interpreter

compnerd:
This removes the dependency on the user specifying `PYTHON_EXECUTABLE`.


compnerd:
@swift-ci please test

PR #306, opened 2020-06-14
Make XCTMain test observations configurable

MaxDesiatov:
Related to [SR-8436](https://bugs.swift.org/browse/SR-8436).

Currently, XCTest uses `PrintObserver` as a default observation. While it is possible to add other custom observations through `XCTestObservationCenter.shared.addTestObserver`, it doesn't prevent the default `PrintObserver` from printing the default output. In certain situations this is undesirable, especially when you'd like to get a custom observation output through stdout. This change allows passing a custom observation through a third new argument to `XCTMain` to replace the default observation. The existing two-argument overload is still available, so this doesn't have any impact on existing users.

To my knowledge, the only widely adopted way to call `XCTMain` directly is in `LinuxMain.swift`, which I think could be the primary use for custom test reporters for now. In the future, XCTest could provide a few other observers in addition to `PrintObserver`.

### DIFF
--- a/Sources/XCTest/Public/XCTestMain.swift
+++ b/Sources/XCTest/Public/XCTestMain.swift
@@ -64,13 +64,13 @@ public func XCTMain(_ testCases: [XCTestCaseEntry]) -> Never {
 }
 
 public func XCTMain(_ testCases: [XCTestCaseEntry], arguments: [String]) -> Never {
-  XCTMain(testCases, arguments: arguments, observation: PrintObserver())
+  XCTMain(testCases, arguments: arguments, observers: [PrintObserver()])
 }
 
 public func XCTMain(
   _ testCases: [XCTestCaseEntry],
   arguments: [String],
-  observation: XCTestObservation
+  observers: [XCTestObservation]
 ) -> Never {
     let testBundle = Bundle.main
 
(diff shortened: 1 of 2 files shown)


Sources/XCTest/Public/XCTestMain.swift
```diff
@@ -62,7 +62,16 @@
...
 }
+
 public func XCTMain(_ testCases: [XCTestCaseEntry], arguments: [String]) -> Never {
+  XCTMain(testCases, arguments: arguments, observation: PrintObserver())
+}
+
+public func XCTMain(
+  _ testCases: [XCTestCaseEntry],
+  arguments: [String],
+  observation: XCTestObservation
```
stmontgomery, line 73: - Can we make the type of the `observation:` parameter an Array of XCTestObservation elements, so callers can pass multiple observers if needed? I think expanding this API to allow passing _multiple_ observers would make it even more flexible.
- Can we rename `observation:` to `observers:` to better reflect that that it's plural (after changing to Array)?

stmontgomery:
@swift-ci Please test

stmontgomery:
ci.swift.org seems to be down, so waiting for that to return and run CI

MaxDesiatov:
Awesome, thanks!

stmontgomery:
@swift-ci Please test

PR #307, opened 2020-06-19
Fix SR-13047 (failure in XCTAssertEqual(Inf, Inf, accuracy: 1e-6))

bhamiltoncx:
`XCTAssertEqual(Inf, Int, accuracy: 1e-6)` currently fails.

To fix the issue, this PR changes `XCTAssertEqual()` to compare values for exact equality *before* checking distance. (Otherwise, distance calculation on infinity returns NaN, which fails all comparisons).

Fixes https://bugs.swift.org/browse/SR-13047 .

### DIFF
--- a/Sources/XCTest/Public/XCTAssert.swift
+++ b/Sources/XCTest/Public/XCTAssert.swift
@@ -174,7 +174,9 @@ public func XCTAssertEqual<T: Equatable>(_ expression1: @autoclosure () throws -
 public func XCTAssertEqual<T: FloatingPoint>(_ expression1: @autoclosure () throws -> T, _ expression2: @autoclosure () throws -> T, accuracy: T, _ message: @autoclosure () -> String = "", file: StaticString = #file, line: UInt = #line) {
     _XCTEvaluateAssertion(.equalWithAccuracy, message: message(), file: file, line: line) {
         let (value1, value2) = (try expression1(), try expression2())
-        if abs(value1.distance(to: value2)) <= abs(accuracy.distance(to: T(0))) {
+        // Test with equality first to handle comparing inf/-inf with itself.
+        if value1 == value2 ||
+             abs(value1.distance(to: value2)) <= abs(accuracy.distance(to: T(0))) {
             return .success
         } else {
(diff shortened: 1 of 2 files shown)


stmontgomery:
Thanks!

@swift-ci Please test

stmontgomery:
I confirmed that the copy of XCTest included in Xcode does not suffer from this issue, it includes an `==` check similar to the one added here, before it calculates distance.

bhamiltoncx:
> I confirmed that the copy of XCTest included in Xcode does not suffer from this issue, it includes an `==` check similar to the one added here, before it calculates distance.

Thanks for the approval and merge.

The bug definitely reproduces with the copy of XCTest included in Xcode 11.5 (11E608c), swiftlang-1103.0.32.9. Perhaps it was fixed separately after that release?

```
% swift -version
Apple Swift version 5.2.4 (swiftlang-1103.0.32.9 clang-1103.0.32.53)
Target: x86_64-apple-darwin19.4.0

% cat Tests/DoubleBugTests/DoubleBugTests.swift 
import XCTest

final class DoubleBugTest : XCTestCase {
  func testCompareWithAccuracy() throws {
    XCTAssertEqual(Double.infinity, Double.infinity, accuracy: 1e-6)
  } 
}

% swift test
(snip)
Test Case '-[DoubleBugTests.DoubleBugTest testCompareWithAccuracy]' started.
/Users/benhamilton/Developer/OpenSource/swift-bug/Tests/DoubleBugTests/DoubleBugTests.swift:5: error: -[DoubleBugTests.DoubleBugTest testCompareWithAccuracy] : XCTAssertEqualWithAccuracy failed: ("inf") is not equal to ("inf") +/- ("1e-06")
Test Case '-[DoubleBugTests.DoubleBugTest testCompareWithAccuracy]' failed (0.250 seconds).
```

stmontgomery:
>The bug definitely reproduces with the copy of XCTest included in Xcode 11.5 (11E608c), swiftlang-1103.0.32.9. Perhaps it was fixed separately after that release?

Yes, please retry that test in the recent Xcode 12 Beta

PR #309, opened 2020-08-30
An XCTUnwrap failure in `setUpWithError` causes the test method to not be invoked

stmontgomery:
This fixes an issue where an `XCTUnwrap` failure in `setUpWithError()` incorrectly causes the test method to not be invoked. If there is an unwrap failure during setUpWithError, the failure should be recorded and the overall test should be marked as a Failure (which is working correctly), but the test method should still be invoked. This is similar to all other XCT assert APIs, where a failure does not cause the remainder of the test to stop executing, regardless of whether that failure occurs in a setUp* method or in the test method itself.


stmontgomery:
@swift-ci Please test

shahmishal:

PR #310, opened 2020-09-05
test: adjust library search paths for Foundation

compnerd:
In order to support a uniform build tree layout, we need to add a new
library search path to XCTest.  This structures the tree similar to all
the other products (including XCTest itself).


compnerd:
@swift-ci please smoke test

compnerd:
@swift-ci please test

PR #311, opened 2020-09-26
Remove unnecessary `import CoreFoundation` on Windows

compnerd:
This removes the unnecessary import of CoreFoundation in the XCTest
module.  It is unused and the removal enables the removal of the
CoreFoundation headers from the distribution which are internal
implementation details on non-Darwin platforms.


compnerd:
@swift-ci please test

PR #312, opened 2020-10-02
test: quote the path to the python interpreter

compnerd:
This is needed to ensure that paths with spaces don't break the test
suite.


compnerd:
@swift-ci please test

PR #304, opened 2020-05-26
SR-1872 Using RunLoopSource in XCTestCase.waitForExpectations

pranavshenoy:
[SR-1872](https://bugs.swift.org/browse/SR-1872) Replacing polling expectations to check whether an expectation is fulfilled with RunloopSource

PR #314, opened 2020-10-11
Sr 1872 using runloopsource in wait for expectation

pranavshenoy:
SR-1872 Replacing polling expectations to check whether an expectation is fulfilled with RunloopSource

### DIFF
--- a/Sources/XCTest/Public/Asynchronous/XCTWaiter.swift
+++ b/Sources/XCTest/Public/Asynchronous/XCTWaiter.swift
@@ -118,6 +118,7 @@ open class XCTWaiter {
     private weak var manager: WaiterManager<XCTWaiter>?
     private var runLoop: RunLoop?
 
+    private var runLoopSource: CFRunLoopSource?
     private weak var _delegate: XCTWaiterDelegate?
     private let delegateQueue = DispatchQueue(label: "org.swift.XCTest.XCTWaiter.delegate")
 
@@ -217,14 +218,7 @@ open class XCTWaiter {
         self.manager = manager
 
         // Begin the core wait loop.
-        let timeoutTimestamp = Date.timeIntervalSinceReferenceDate + timeout
-        while !isFinished {
-            let remaining = timeoutTimestamp - Date.timeIntervalSinceReferenceDate
-            if remaining <= 0 {
-                break
-            }
-            primitiveWait(using: runLoop, duration: remaining)
-        }
+        primitiveWait(using: runLoop, duration: timeout)
 
         manager.stopManaging(self)
         self.manager = nil
@@ -356,13 +350,20 @@ open class XCTWaiter {
 
 private extension XCTWaiter {
     func primitiveWait(using runLoop: RunLoop, duration timeout: TimeInterval) {
-        // The contract for `primitiveWait(for:)` explicitly allows waiting for a shorter period than requested
-        // by the `timeout` argument. Only run for a short time in case `cancelPrimitiveWait()` was called and
-        // issued `CFRunLoopStop` just before we reach this point.
-        let timeIntervalToRun = min(0.1, timeout)
+        var context: CFRunLoopSourceContext   = CFRunLoopSourceContext(version: 0,
+                                                               info: nil,
+                                                               retain: nil,
+                                                               release: nil,
+                                                               copyDescription: nil,
+                                                               equal: nil,
+                                                               hash: nil,
+                                                               schedule: { _, _, _ in },
+                                                               cancel: {  _, runLoop, _ in  },
+                                                               perform: { _ in })
 
-        // RunLoop.run(mode:before:) should have @discardableResult <rdar://problem/45371901>
-        _ = runLoop.run(mode: .default, before: Date(timeIntervalSinceNow: timeIntervalToRun))
+        runLoopSource = CFRunLoopSourceCreate(nil, 0, &context)
+        CFRunLoopAddSource(runLoop.getCFRunLoop() , runLoopSource, CFRunLoopMode.defaultMode)
+        _ = runLoop.run(until: .init(timeIntervalSinceNow: timeout))
     }
 
     func cancelPrimitiveWait() {
@@ -370,7 +371,8 @@ private extension XCTWaiter {
 #if os(Windows)
         runLoop._stop()
 #else
-        CFRunLoopStop(runLoop.getCFRunLoop())
+        guard let runLoopSource = runLoopSource else { return }
+        CFRunLoopRemoveSource(runLoop.getCFRunLoop(), runLoopSource, CFRunLoopMode.defaultMode)
 #endif
     }
 }
@@ -405,7 +407,7 @@ extension XCTWaiter: ManageableWaiter {
         dispatchPrecondition(condition: .onQueue(XCTWaiter.subsystemQueue))
 
         queue_validateExpectationFulfillment(dueToTimeout: true)
-        manager!.queue_handleWatchdogTimeout(of: self)
+        manager?.queue_handleWatchdogTimeout(of: self)
         cancelPrimitiveWait()
     }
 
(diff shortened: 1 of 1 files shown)


pranavshenoy:
@compnerd  can we test the changes??

compnerd:
@swift-ci please test

pranavshenoy:
Previous discussions: https://github.com/apple/swift-corelibs-xctest/pull/304

Sources/XCTest/Public/Asynchronous/XCTWaiter.swift
```diff
@@ -405,7 +389,7 @@ extension XCTWaiter: ManageableWaiter {
         dispatchPrecondition(condition: .onQueue(XCTWaiter.subsystemQueue))
 
         queue_validateExpectationFulfillment(dueToTimeout: true)
-        manager!.queue_handleWatchdogTimeout(of: self)
+        manager?.queue_handleWatchdogTimeout(of: self)
```
stmontgomery, line 392: What was the reason for changing this, did you encounter a case where the force-unwrap failed after your changes? I vaguely recall using force unwrap intentionally here to represent the expectation that, at this point in the waiter lifecycle, it definitely should still have a WaiterManager, but I may have missed something or this PR could have changed that invariant.

Sources/XCTest/Public/Asynchronous/XCTWaiter.swift
```diff
@@ -356,22 +349,13 @@ open class XCTWaiter {
...
     }
 
     func cancelPrimitiveWait() {
-        guard let runLoop = runLoop else { return }
-#if os(Windows)
-        runLoop._stop()
-#else
-        CFRunLoopStop(runLoop.getCFRunLoop())
-#endif
+        runLoopSource?.invalidate()
```
stmontgomery, line 358: This `cancelPrimitiveWait()` method is always called from the `XCTWaiter.subsystemQueue` DispatchQueue, but with the changes in `primitiveWait` above, the `runLoopSource` stored property will now be read here on that private queue but assigned in `primitiveWait` which is _not_ on that same queue.

So I think there could be a data race lurking here. The current code in `primitiveWait` takes the RunLoop as a function parameter, which is done intentionally because the _assignment_ of that RunLoop to the `runLoop` stored property is done earlier, inside a `DispatchQueue.sync {...}` block. I think we ought to take a similar approach here by creating and storing the `RunLoop._Source` instance earlier in the calling `wait(...)` function, and assign it to the stored properties inside that protected `.sync {...}` block. Then, accessing the `runLoopSource` stored property here in `cancelPrimitiveWait()` should become safe. We might even consider adding a

```swift
dispatchPrecondition(condition: .onQueue(XCTWaiter.subsystemQueue))
```

to it, like other on-queue funcs have, to help ensure this.

Sources/XCTest/Public/Asynchronous/XCTWaiter.swift
```diff
@@ -117,7 +117,7 @@ open class XCTWaiter {
     internal var waitSourceLocation: SourceLocation?
     private weak var manager: WaiterManager<XCTWaiter>?
     private var runLoop: RunLoop?
-
+    private var runLoopSource: RunLoop._Source?
```
stmontgomery, line 120: I notice the `RunLoop._Source` type and several of its related RunLoop methods are marked as deprecated in Foundation… does that mean this PR will introduce new build warnings to the XCTest codebase? I wonder what we can do about that, since Swift has no way to "silence" those warnings.

stmontgomery (changes requested):
Hi @pranavshenoy, thanks for taking this on. I have familiarized myself with the discussion on the previous PR, #304, and I have a few questions. One "larger" question, and some smaller inline suggestions about the code…

My biggest question about this change is: Is it missing the step of "signaling" the run loop source? I had to refresh my memory on how these work, but IIRC the idea behind run loop sources is that, once you assign a source to a run loop and begin running the run loop, it begins and doesn't finish until either its source is signaled _or_ it times out.

In this PR, it looks like you're adding the source to the run loop, and at the very end (once it's no longer needed) you're calling `invalidate()` on the source… but I don't see a place where `RunLoop._Source.signal()` is called. My understanding was that, each time one of the waiter's XCTestExpectations is fulfilled, we would need to `signal()` the source to cause the run loop to spin again. And looking at Apple's [documentation for `CFRunLoopSourceSignal`](https://developer.apple.com/documentation/corefoundation/1543700-cfrunloopsourcesignal?language=objc) (the underlying C API), it seems like one would need to call `CFRunLoopWakeUp` after `signal()` as well, since we'd want the run loop to process it and fire immediately. (I'm not sure there's a Swift `RunLoop._Source` wrapper for `CFRunLoopWakeUp`, yet.)

Can you clarify if I am misunderstanding something or if this signaling happens a different way?

pranavshenoy:
Hi @stmontgomery 
Thanks for the detailed [review](https://github.com/apple/swift-corelibs-xctest/pull/314#pullrequestreview-537666423).

 From my understanding of the code, the RunLoop in the `primitiveWait()` is expecting only invalidation either when all expectations are fulfilled or the time is elapsed. I believe invalidation of runLoopSource would suffice for this requirement and “signaling” the runLoopSource might not not required.  When `fulfill()` is invoked for an expectation, the `validateExpectations()` method within it checks whether all of the waiter’s XCTestExpectations are fulfilled. If all are fulfilled, the `queue_finished()` is invoked, which in turn invokes `cancelPrimitiveWait()`, invalidating the runLoop and resuming the execution of `wait()`. 

From my understanding of the task, the while loop in the `wait()`, which checks for `!isFinished`. must be replaced with RunLoopSource. Additionally, I feel the same could be achieved without RunLoopSource (starting the RunLoop and stopping it). 

Please let me know if I am misinterpreting the code and/or the task.

Sources/XCTest/Public/Asynchronous/XCTWaiter.swift
pranavshenoy: Have made the changes. Thanks

Sources/XCTest/Public/Asynchronous/XCTWaiter.swift
pranavshenoy: Initially, this was causing crashes. Hence, felt optional unwrapping would be safer to avoid crashes. 
Have reverted the changes. Thanks.

pranavshenoy:
@compnerd  could you please trigger the build? Thanks

spevans:
@swift-ci test

Sources/XCTest/Public/Asynchronous/XCTWaiter.swift
pranavshenoy: @stmontgomery  was looking into the root cause of the crash. It looks like for the scenario where the duration of wait() is less than or equal to zero (there are two functional test cases for this), there exists a race-condition between `queue_handleWatchdogTimeout()` and `self.manager = nil` with the latest changes. This is because  the `isFinished` property, which has `subsystemQueue.sync` that prevents the race-condition, is not needed in the latest changes. 

Please let me know if the approach needs to be changed.

pranavshenoy:
@stmontgomery 
from this [FIXME](https://github.com/apple/swift-corelibs-xctest/blob/4971c8f73be61b991f8a8f8a6c354002d89cc1df/Sources/XCTest/Public/XCTestCase%2BAsynchronous.swift#L102-L105), I feel that the expectation of the task was to signal RunLoopSource when each expectation was fulfilled() instead of polling. But with the additional capabilities like InverseExpectation, innerWaiter and outerWaiter, this is handled by  `validateExpectations()`.  Hence I feel, calling `runLoop.run(until: Date(timeIntervalSinceNow: timeout))` wait() and letting `queue_finish()` cancel the RunLoop would suffice. 

I tried it out locally and all tests passed. 

Could you please tell me if there is some gap in this approach that I am missing?  If this one is good, I can proceed and make the necessary changes.

Thanks

pranavshenoy:
@stmontgomery @millenomi  Can you share your thoughts on this?

pranavshenoy:
@stmontgomery @millenomi @compnerd  Could you please provide your inputs on this?

Thanks

pranavshenoy:
> @stmontgomery @millenomi @compnerd Could you please provide your inputs on this?

@stmontgomery @millenomi @compnerd Could you please provide your inputs on this?

PR #316, opened 2020-10-17
Only use the last CoreFoundation reference on Apple platforms

buttaface:
Now that CoreFoundation is no longer exposed on new platforms, ie except for Darwin and linux, apple/swift-corelibs-foundation#2878, change this last reference that was added years ago, before `RunLoop._stop()` was available. Removing the CoreFoundation interface broke building XCTest for Android, just like on Windows, which this patch fixes.

@millenomi and @compnerd, if this won't work for some reason, just let me know and I could try another approach.

### DIFF
--- a/Sources/XCTest/Public/Asynchronous/XCTWaiter.swift
+++ b/Sources/XCTest/Public/Asynchronous/XCTWaiter.swift
@@ -10,7 +10,7 @@
 //  XCTWaiter.swift
 //
 
-#if !os(Windows)
+#if os(macOS) || os(iOS) || os(tvOS) || os(watchOS)
 import CoreFoundation
 #endif
 
@@ -369,10 +369,10 @@ private extension XCTWaiter {
 
     func cancelPrimitiveWait() {
         guard let runLoop = runLoop else { return }
-#if os(Windows)
-        runLoop._stop()
-#else
+#if os(macOS) || os(iOS) || os(tvOS) || os(watchOS)
         CFRunLoopStop(runLoop.getCFRunLoop())
+#else
+        runLoop._stop()
 #endif
     }
 }


spevans:
@swift-ci test

Sources/XCTest/Public/Asynchronous/XCTWaiter.swift
```diff
@@ -369,11 +365,7 @@ private extension XCTWaiter {
 
     func cancelPrimitiveWait() {
         guard let runLoop = runLoop else { return }
-#if os(Windows)
         runLoop._stop()
```
stmontgomery, line 368: I notice `RunLoop._stop()` is marked `deprecated`:

https://github.com/apple/swift-corelibs-foundation/blob/main/Sources/Foundation/RunLoop.swift#L238

Does this change mean there will be a deprecation warning here at this call site in XCTest? I'm not aware of any way to "silence" such deprecation warnings, so just curious since this feels somewhat unfortunate.

buttaface: Yes, that's the only drawback, it now puts out a deprecation warning saying this function is "For XCTest use only." The good news is that it clearly says it's okay for XCTest, so nobody will be confused when they see it.

[The plan is to remove CF as a public API for all non-Darwin platforms](https://forums.swift.org/t/formalizing-the-unavailability-of-core-foundation/40216), so this is doing that. I figure that once a public API is decided, this can be changed to that. In the meantime, this avoids breakage on platforms like Android and Windows where CF APIs were just removed.

An alternative solution would be to keep it the way it is now and just extend the non-CF platforms using `RunLoop._stop()` whenever needed, while keeping Darwin and linux using the CF API for now.

compnerd:
@buttaface it is possible to build XCTest against Foundation.framework from Darwin which wont provide `_stop`.  I think that the work you have done here is actually desirable, but perhaps we need to back-pedal a small amount and leave the old path around, or alternatively, provide a shim in that case that bridges `_stop`.  I think that the shim is actually nicer because it keeps a single path through the code.

buttaface:
Alright, changed it to keep using the CF API on Darwin platforms, and the provisional `_stop()` API everywhere else.

spevans:
@swift-ci test

compnerd (approved):
@stmontgomery any objections?  CoreFoundation isn't meant to be accessible on non-Darwin platforms, this is now strictly a bug fix, even though it has the side effect of printing the deprecation warning.

stmontgomery:
> @stmontgomery any objections? CoreFoundation isn't meant to be accessible on non-Darwin platforms, this is now strictly a bug fix, even though it has the side effect of printing the deprecation warning.

Yes thanks @compnerd and @buttaface, this LGTM now.

Not necessarily for this PR, but should we eventually be adopting `@_implementationOnly` for the `import CoreFoundation`, even on Darwin, to avoid re-exporting CoreFoundation in XCTest's module?

stmontgomery (approved):

compnerd:
I think that it might be a good idea - it would mean that the behavioural differences would be minimal when using XCTest on the various platforms.

PR #317, opened 2020-10-20
Add an additional search path for llvm-lit

gmittert:
If we already have an llvm config file in a build tree, we can also
check there for an llvm-lit to test with. See also:
https://reviews.llvm.org/D77110


gmittert:
@swift-ci please test

gmittert:
cc @stmontgomery @compnerd I don't have write access, could I get a review please?

stmontgomery (approved):

stmontgomery:
Thanks @gmittert!

PR #318, opened 2020-11-18
[cmake] Make dispatch/Foundation non required

gmittert:
When building using an install tree which already contains
dispatch/Foundation, a dispatch/foundation packages do not need to be
specified since they are already automatically linked against.


gmittert:
@swift-ci please test platform

PR #320, opened 2020-11-24
[SR-12575] Over-fulfilling XCTestExpectation records test failure even when `assertForOverFulfill` is disabled

stmontgomery:
This fixes an issue where calling `XCTestExpectation.fulfill()` more times than expected will still
record a test failure, even if its `assertForOverFulfill` property is set to `false` (the default).
In this Corelibs version of XCTest, `assertForOverFulfill` is only guarding a `fatalError`, but we
still unconditionally record a test failure which causes this symptom. This means the behavior is
different than in the Xcode copy of XCTest, where if `assertForOverFulfill` is false, neither an
NSAssert exception nor a test failure occurs. Swift doesn't support exceptions, so I believe the
unconditional test failure was added here in the past in an attempt to emulate that behavior.

The fix here removes the `fatalError()` entirely, and leaves the test failure-recording logic but
makes it conditional on `assertForOverFulfill` being true. The behavior will still differ from Xcode's
copy of XCTest, but I think it's justifiable and closer to what users expect. The use of NSAssert in
the (ObjC) Xcode copy of XCTest was largely an attempt to get better diagnostics when an over-fulfill
occurred by capturing the backtrace of that fulfillment. This is especially important with
XCTestExpectations, since they are meant for testing async conditions, and an over-fulfillment could
occur after their test has completed, so without a backtrace or some better state tracking the test
failure could be misattributed to a subsequent test once it becomes "current". Corelibs XCTest doesn't
suffer from this diagnostics problem as much since it already captures `file:line:` in `fulfill()`.
Moreover, there have been requests to change that behavior in ObjC XCTest and use a "plain" test failure
everywhere instead of NSAssert.

So this PR gets us closer to that goal, starting here in Corelibs, by removing the fatalError and only
leaving the existing test failure-recording behavior. To fully realize it and ensure these test
failures are attributed to the correct test, we'll need to make further internal state tracking
enhancements (63874139), but for now this seems like a good incremental step, considering that Corelibs
already performed this test failure recording. I've also added a new test and modified an existing one.

rdar://problem/62202297

### DIFF
--- a/Sources/XCTest/Public/Asynchronous/XCTestExpectation.swift
+++ b/Sources/XCTest/Public/Asynchronous/XCTestExpectation.swift
@@ -259,19 +259,12 @@ open class XCTestExpectation {
             //        expectations have completed. Similarly, this should cause an
             //        error as well.
 
-            if queue_isFulfilled {
-                // FIXME: No regression tests exist for this feature. We may break it
-                //        without ever realizing (similar to `continueAfterFailure`).
-                if _assertForOverFulfill {
-                    fatalError("API violation - multiple calls made to fulfill() for \(queue_expectationDescription)")
-                }
+            if queue_isFulfilled, _assertForOverFulfill, let testCase = XCTCurrentTestCase {
+                testCase.recordFailure(
+                    description: "API violation - multiple calls made to XCTestExpectation.fulfill() for \(queue_expectationDescription).",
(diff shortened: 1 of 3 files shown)


stmontgomery:
@swift-ci please test

briancroom (approved):
Awesome! Thanks for the fix.

PR #325, opened 2021-08-03
Implement XCTAssertIdentical & XCTAssertNotIdentical

egorzhdan:
These functions are available in Xcode 12.5, but aren't available on Linux/Windows.
This change adds the implementations for them.

### DIFF
--- a/Tests/Functional/FailureMessagesTestCase/main.swift
+++ b/Tests/Functional/FailureMessagesTestCase/main.swift
@@ -28,6 +28,8 @@ class FailureMessagesTestCase: XCTestCase {
             ("testAssertFalse", testAssertFalse),
             ("testAssertGreaterThan", testAssertGreaterThan),
             ("testAssertGreaterThanOrEqual", testAssertGreaterThanOrEqual),
+            ("testAssertIdentical", testAssertIdentical),
+            ("testAssertNotIdentical", testAssertNotIdentical),
             ("testAssertLessThan", testAssertLessThan),
             ("testAssertLessThanOrEqual", testAssertLessThanOrEqual),
             ("testAssertNil", testAssertNil),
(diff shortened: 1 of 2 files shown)


egorzhdan:
@swift-ci please test

Tests/Functional/FailureMessagesTestCase/main.swift
```diff
@@ -121,6 +123,20 @@ class FailureMessagesTestCase: XCTestCase {
...
+// CHECK: Test Case 'FailureMessagesTestCase.testAssertIdentical' failed \(\d+\.\d+ seconds\)
+    func testAssertIdentical() {
+        XCTAssertIdentical("a" as AnyObject, "b" as AnyObject, "message", file: "test.swift")
+    }
+
+// CHECK: Test Case 'FailureMessagesTestCase.testAssertNotIdentical' started at \d+-\d+-\d+ \d+:\d+:\d+\.\d+
+// CHECK: test.swift:[[@LINE+3]]: error: FailureMessagesTestCase.testAssertNotIdentical : XCTAssertNotIdentical failed: \("nil"\) is identical to \("nil"\) - message
+// CHECK: Test Case 'FailureMessagesTestCase.testAssertNotIdentical' failed \(\d+\.\d+ seconds\)
+    func testAssertNotIdentical() {
+        XCTAssertNotIdentical(nil, nil, "message", file: "test.swift")
```
stmontgomery, line 137: Thank you for this change! I think we can land it soon

My only request is that you add a few more assertions to these two tests, in particular **success** cases which validate that both `XCTAssertIdentical` and `XCTAssertNotIdentical` _succeed_ when passed identical or not identical values, respectively. Currently, the tests only validate failure scenarios, but we should be sure to have tests covering when they pass as well. Personally, I'm fine just adding those inside the existing two test method bodies, to keep things simple.

stmontgomery (changes requested):

Tests/Functional/FailureMessagesTestCase/main.swift
egorzhdan: Oh that's a good point. I added a few succeeding assertions to the test.

egorzhdan:
@swift-ci please test

Tests/Functional/FailureMessagesTestCase/main.swift
```diff
@@ -121,6 +123,27 @@ class FailureMessagesTestCase: XCTestCase {
         XCTAssertGreaterThanOrEqual(-1, 0, "message", file: "test.swift")
     }
 
+// CHECK: Test Case 'FailureMessagesTestCase.testAssertIdentical' started at \d+-\d+-\d+ \d+:\d+:\d+\.\d+
+// CHECK: test.swift:[[@LINE+7]]: error: FailureMessagesTestCase.testAssertIdentical : XCTAssertIdentical failed: \("a"\) is not identical to \("b"\) - message
+// CHECK: Test Case 'FailureMessagesTestCase.testAssertIdentical' failed \(\d+\.\d+ seconds\)
+    func testAssertIdentical() {
+        let object = XCTestExpectation()
+        XCTAssertIdentical(object as AnyObject, object as AnyObject)
```
stmontgomery, line 131: Is it actually necessary to cast these using `as AnyObject` ? I would expect it to work without the casts

egorzhdan: With this specific assertion the cast wasn't needed indeed, I removed it.
All the other assertions require a cast to avoid a compiler error (`type of expression is ambiguous without more context`).

stmontgomery (approved):

egorzhdan:
@swift-ci please test

briancroom:
Looks great. Thank you!

PR #326, opened 2021-08-04
[REVIEW] Add Async Support to XCTest

SeanROlszewski:
This is a work-in-progress pull request that @stmontgomery and I have been working on which adds first-class `async/await` support to the cross-platform XCTest. 

Please withhold comments until this PR changes from [WIP] to [REVIEW]. 

**17/08/21 Update**: This PR has now been opened for review and is being actively iterated on.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ xcuserdata
 *.xcscmblueprint
 .build/
 Output/
+Tests/Functional/.lit_test_times.txt
--- a/Sources/XCTest/Public/XCTestCase.TearDownBlocksState.swift
+++ b/Sources/XCTest/Public/XCTestCase.TearDownBlocksState.swift
@@ -23,11 +23,8 @@ extension XCTestCase {
         
         @available(macOS 12, *)
         func append(_ block: @Sendable @escaping () async throws -> Void) {
-            XCTWaiter.subsystemQueue.sync {
-                precondition(wasFinalized == false, "API violation -- attempting to add a teardown block after teardown blocks have been dequeued")
-                blocks.append {
-                    try awaitUsingExpectation { try await block() }
-                }
+            self.append {
+                try awaitUsingExpectation { try await block() }
             }
         }
         
--- a/Sources/XCTest/Public/XCTestCase.TearDownBlocksState.swift
+++ b/Sources/XCTest/Public/XCTestCase.TearDownBlocksState.swift
@@ -22,7 +22,7 @@ extension XCTestCase {
         private var blocks: [() throws -> Void] = []
         
         @available(macOS 12, *)
-        func append(_ block: @escaping () async throws -> Void) {
+        func append(_ block: @Sendable @escaping () async throws -> Void) {
             XCTWaiter.subsystemQueue.sync {
                 precondition(wasFinalized == false, "API violation -- attempting to add a teardown block after teardown blocks have been dequeued")
                 blocks.append {
--- a/Sources/XCTest/Public/XCTestCase.TearDownBlocksState.swift
+++ b/Sources/XCTest/Public/XCTestCase.TearDownBlocksState.swift
@@ -1,0 +1,48 @@
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2021 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//
+//  XCTestCase.ClosureType.swift
+//  Extension on XCTestCase which encapsulates ClosureType
+//
+
+///  XCTestCase.TeardownBlocksState
+///  A class which encapsulates teardown blocks which are registered via the `addTearDownBlock(_block:)` method.
+///  Supports async and sync throwing methods
+
+extension XCTestCase {
+    final class TeardownBlocksState {
+        
+        private var wasFinalized = false
+        private var blocks: [() throws -> Void] = []
+        
+        @available(macOS 12, *)
+        func append(_ block: @escaping () async throws -> Void) {
+            XCTWaiter.subsystemQueue.sync {
+                precondition(wasFinalized == false, "API violation -- attempting to add a teardown block after teardown blocks have been dequeued")
+                blocks.append {
+                    try awaitUsingExpectation { try await block() }
+                }
+            }
+        }
+        
+        func append(_ block: @escaping () throws -> Void) {
+            XCTWaiter.subsystemQueue.sync {
+                precondition(wasFinalized == false, "API violation -- attempting to add a teardown block after teardown blocks have been dequeued")
+                blocks.append(block)            }
+        }
+        
+        func finalize() -> [() throws -> Void] {
+            XCTWaiter.subsystemQueue.sync {
+                precondition(wasFinalized == false, "API violation -- attempting to run teardown blocks after they've already run")
+                wasFinalized = true
+                return blocks
+            }
+        }
+    }
+}
--- a/Sources/XCTest/Public/XCTestCase.swift
+++ b/Sources/XCTest/Public/XCTestCase.swift
@@ -206,7 +206,7 @@ open class XCTestCase: XCTest {
     /// method ends.
     @available(macOS 12.0, *)
     open func addTeardownBlock(_ block: @Sendable @escaping () async throws -> Void) {
-        teardownBlocksState.append(block)
+        teardownBlocksState.appendAsync(block)
     }
 
     private func performSetUpSequence() {
--- a/Sources/XCTest/Public/XCTestCase.swift
+++ b/Sources/XCTest/Public/XCTestCase.swift
@@ -35,7 +35,7 @@ open class XCTestCase: XCTest {
     private let testClosure: XCTestCaseClosure
 
     private var skip: XCTSkip?
-    private let teardownBlocksState: XCTestCase.TeardownBlocksState
+    private let teardownBlocksState = XCTestCase.TeardownBlocksState()
 
     /// The name of the test case, consisting of its class name and the method
     /// name it will run.
@@ -201,6 +201,13 @@ open class XCTestCase: XCTest {
     open func addTeardownBlock(_ block: @escaping () -> Void) {
         teardownBlocksState.append(block)
     }
+    
+    /// Registers a block of teardown code to be run after the current test
+    /// method ends.
+    @available(macOS 12.0, *)
+    open func addTeardownBlock(_ block: @escaping () async throws -> Void) {
+        teardownBlocksState.append(block)
+    }
 
     private func performSetUpSequence() {
         func handleErrorDuringSetUp(_ error: Error) {
@@ -236,35 +243,25 @@ open class XCTestCase: XCTest {
         self.setUp()
     }
     
-    static let asyncTestTimeout = 5
     
-    @available(macOS 12.0.0, *)
     private func performTearDownSequence() {
-        @Sendable func handleErrorDuringTearDown(_ error: Error) {
+        func handleErrorDuringTearDown(_ error: Error) {
             if error.xct_shouldRecordAsTestFailure {
                 recordFailure(for: error)
             }
         }
         
-        @Sendable func runTeardownBlocks(errorHandler: @escaping (Error) -> Void) async {
-            for block in teardownBlocks.finalize().reversed() {
+        func runTeardownBlocks(errorHandler: @escaping (Error) -> Void) {
+            for block in self.teardownBlocksState.finalize().reversed() {
                 do {
-                    try await block()
+                    try block()
                 } catch {
                     errorHandler(error)
                 }
             }
         }
         
-        let runTeardownBlocksExpectation = XCTestExpectation(description: "runTeardownBlocksExpectation")
-        Task {
-            defer { runTeardownBlocksExpectation.fulfill() }
-            await runTeardownBlocks(errorHandler: handleErrorDuringTearDown(_:))
-        }
-        _ = XCTWaiter.wait(for: [runTeardownBlocksExpectation], timeout: Self.asyncTestTimeout)
-        
-        
-        runTeardownBlocks()
+        runTeardownBlocks(errorHandler: handleErrorDuringTearDown(_:))
 
         tearDown()
 
@@ -282,19 +279,6 @@ open class XCTestCase: XCTest {
             }
         } catch {
             handleErrorDuringTearDown(error)
-        }
-    }
-
-    private func runTeardownBlocks() {
-        let blocks = teardownBlocksQueue.sync { () -> [() -> Void] in
-            self.teardownBlocksDequeued = true
-            let blocks = self.teardownBlocks
-            self.teardownBlocks = []
-            return blocks
-        }
-
-        for block in blocks.reversed() {
-            block()
         }
     }
 
--- a/Sources/XCTest/Public/XCTestCase.swift
+++ b/Sources/XCTest/Public/XCTestCase.swift
@@ -209,11 +209,9 @@ open class XCTestCase: XCTest {
     }
 
     private func performSetUpSequence() {
-        do {
-            try setUpWithError()
-        } catch {
+        func handleErrorDuringSetUp(_ error: Error) {
             if error.xct_shouldRecordAsTestFailure {
-                recordFailure(for: error)
+                self.recordFailure(for: error)
             }
 
             if error.xct_shouldSkipTestInvocation {
@@ -225,10 +223,30 @@ open class XCTestCase: XCTest {
             }
         }
 
-        setUp()
+        do {
+            try awaitUsingExpectation {
+                try await self.setUp()
+            }
+        } catch {
+            handleErrorDuringSetUp(error)
+        }
+
+        do {
+            try self.setUpWithError()
+        } catch {
+            handleErrorDuringSetUp(error)
+        }
+
+        self.setUp()
     }
 
     private func performTearDownSequence() {
+        func handleErrorDuringTearDown(_ error: Error) {
+            if error.xct_shouldRecordAsTestFailure {
+                recordFailure(for: error)
+            }
+        }
+
         runTeardownBlocks()
 
         tearDown()
@@ -236,9 +254,15 @@ open class XCTestCase: XCTest {
         do {
             try tearDownWithError()
         } catch {
-            if error.xct_shouldRecordAsTestFailure {
-                recordFailure(for: error)
+            handleErrorDuringTearDown(error)
+        }
+
+        do {
+            try awaitUsingExpectation {
+                try await self.tearDown()
             }
+        } catch {
+            handleErrorDuringTearDown(error)
         }
     }
 
@@ -292,3 +316,57 @@ private func test<T: XCTestCase>(_ testFunc: @escaping (T) -> () throws -> Void)
         try testFunc(testCase)()
     }
 }
+
+public func asyncTest<T: XCTestCase>(
+    _ testClosureGenerator: @escaping (T) -> () async throws -> Void
+) -> (T) -> () throws -> Void {
+    return { (testType: T) in
+        let testClosure = testClosureGenerator(testType)
+        return {
+            try awaitUsingExpectation(testClosure)
+        }
+    }
+}
+
+private func awaitUsingExpectation(
+    _ closure: @escaping () async throws -> Void
+) throws -> Void {
+    let expectation = XCTestExpectation(description: "async test completion")
+    let thrownErrorWrapper = ThrownErrorWrapper()
+
+    Task {
+        defer { expectation.fulfill() }
+
+        do {
+            try await closure()
+        } catch {
+            thrownErrorWrapper.error = error
+        }
+    }
+
+    _ = XCTWaiter.wait(for: [expectation], timeout: asyncTestTimeout)
+
+    if let error = thrownErrorWrapper.error {
+        throw error
+    }
+}
+
+private final class ThrownErrorWrapper: @unchecked Sendable {
+
+    private var _error: Error?
+
+    var error: Error? {
+        get {
+            XCTWaiter.subsystemQueue.sync { _error }
+        }
+        set {
+            XCTWaiter.subsystemQueue.sync { _error = newValue }
+        }
+    }
+}
+
+
+// This time interval is set to a very large value due to their being no real native timeout functionality within corelibs-xctest.
+// With the introduction of async/await support, the framework now relies on XCTestExpectations internally to coordinate the addition async portions of setup and tear down.
+// This time interval is the timeout corelibs-xctest uses with XCTestExpectations.
+private let asyncTestTimeout: TimeInterval = 60 * 60 * 24 * 30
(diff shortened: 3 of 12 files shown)


SeanROlszewski:
@swift-ci Please test

Sources/XCTest/Public/XCTestCase.swift
```diff
@@ -108,41 +126,185 @@ open class XCTestCase: XCTest {
...
+        }
+        
+        switch testClosure {
+        case .regular(let closure):
+            
+            do {
+                try setUpWithError()
+            } catch {
+                handleError(error)
+                return
```
stmontgomery, line 194: We won't want to early `return` here if `setUpWithError` throws an error, since that will mean we'll skip the non-throwing `setUp` as well as `tearDown` and `tearDownWithError` later.

If either the `setUp() async throws` or `setUpWithError()` methods throw an error, we need to skip invoking the test closure but still invoke all remaining set up method(s) and all tear down methods. The existing implementation does this by setting `self.skip` if an Error is caught during one of the set up methods, and then that property is checked before invoking the test closure

Sources/XCTest/Public/XCTestCase.swift
```diff
@@ -108,41 +126,185 @@ open class XCTestCase: XCTest {
...
+            
+            setUp()
+            
+            try closure(self)
+            
+            do {
+                try tearDownWithError()
+            } catch {
+                handleError(error)
+                return
```
stmontgomery, line 205: (Same comment here about the early `return`)

Sources/XCTest/Public/XCTestCase.swift
```diff
@@ -108,41 +126,185 @@ open class XCTestCase: XCTest {
...
+            }
+            
+            tearDown()
```
stmontgomery, line 208: Need to invoke `tearDown` _before_ `tearDownWithError`: the tear down methods are invoked in the opposite order as the set up methods

Sources/XCTest/Public/XCTestCase.swift
```diff
@@ -108,41 +126,185 @@ open class XCTestCase: XCTest {
...
+            
+            
+            
+            Task {
+                defer { expectation.fulfill() }
+                await runTeardownBlocks(errorHandler: handleError(_:))
```
stmontgomery, line 214: We'll need to `runTeardownBlocks` _before_ all `tearDown` methods

The complete order of execution should be:

```
setUp() async throws
setUpWithError() throws
setUp()

// the test closure, if none of the setUp methods above threw an error. The test closure may or may not be async.

runTeardownBlocks
tearDown()
tearDownWithError() throws
tearDown() async throws
```

Sources/XCTest/Public/XCTestCase.swift
```diff
@@ -108,41 +126,185 @@ open class XCTestCase: XCTest {
...
+                        do {
+                            try await self.performSetUpSequence()
+                            try await closure(self)
+                        } catch {
+                            handleError(error)
+                        }
+                    }
+                    
+                    group.addTask {
+                        await self.performTearDownSequence()
```
stmontgomery: I don't think we can safely use the `withThrowingTaskGroup` or `group.addTask` APIs here since it means their order of execution is unspecified, so for example tearDown could end up running before or concurrently with `setUp`. I think we'll probably want to have a code flow similar to that of the `.regular` closure type above

```swift
  /// Unconditionally create a child task in the group.
  /// 
  /// The child task will be executing concurrently with the group, and its result 
  /// may be collected by calling `group.next()` or iterating over the group gathering 
  /// all submitted task results from the group.
  mutating func addTask(
```
https://github.com/apple/swift-evolution/blob/main/proposals/0304-structured-concurrency.md#creating-taskgroup-child-tasks

.gitignore
```diff
@@ -4,3 +4,4 @@ xcuserdata
 *.xcscmblueprint
 .build/
 Output/
+*.lit_test_times.txt
```
SeanROlszewski, line 7: This was a drive-by change I added as `lit` kept generating this file and `git` kept marking it as untracked. 

I'm uncertain if this has always been the case, or if I should be leveraging this file. If we _don't_ want to ignore it, I can back this out and consider it a quality-of-life change for my local dev experience.

Sources/XCTest/Public/XCTestCase.swift
```diff
@@ -136,13 +153,140 @@ open class XCTestCase: XCTest {
...
+                            nonAsyncSetUp()
+                            
+                            try await closure(self)
+                            
+                        } catch {
+                            handleError(error)
+                        }
+                    }
+                    
+                    group.addTask {
```
stmontgomery, line 242: I don't think we can safely use the `withThrowingTaskGroup` or `group.addTask` APIs here since it means their order of execution is unspecified, so for example tearDown could end up running before or concurrently with `setUp`. I think we'll probably want to have a code flow similar to that of the `.regular` closure type above

```swift
  /// Unconditionally create a child task in the group.
  /// 
  /// The child task will be executing concurrently with the group, and its result 
  /// may be collected by calling `group.next()` or iterating over the group gathering 
  /// all submitted task results from the group.
  mutating func addTask(
```
https://github.com/apple/swift-evolution/blob/main/proposals/0304-structured-concurrency.md#creating-taskgroup-child-tasks

SeanROlszewski:
* Rename `ClosureType` -> `XCTestMethodConvention`
* Potentially make an `XCTFI` equivalent

SeanROlszewski:
@swift-ci please test 😇

Sources/XCTest/Public/XCTestCase.swift
```diff
@@ -195,66 +304,37 @@ open class XCTestCase: XCTest {
     /// class.
     open class func tearDown() {}
 
-    private var teardownBlocks: [() -> Void] = []
-    private var teardownBlocksDequeued: Bool = false
-    private let teardownBlocksQueue: DispatchQueue = DispatchQueue(label: "org.swift.XCTest.XCTestCase.teardownBlocks")
+    private class TeardownBlocksState {
```
briancroom, line 307: We’ve lost the thread safety on the list of teardown blocks here. We need to either figure out how to successfully make this an actor again, or (probably easier / better for starting) reintroduce the serial dispatch queue being used as a lock around this state.

Sources/XCTest/Public/XCTestCase.swift
```diff
@@ -108,41 +130,128 @@ open class XCTestCase: XCTest {
...
     }
 
+    public required init(name: String, closureType: ClosureType) {
+        _name = "\(type(of: self)).\(name)"
+        self.testClosure = closureType
+    }
     /// Invoking a test performs its setUp, invocation, and tearDown. In
     /// general this should not be called directly.
     open func invokeTest() {
-        performSetUpSequence()
```
briancroom, line 119: I think we ought to reintroduce the performSetUpSequence and performTearDownSequence helpers here, since they make the overall flow through this method a bit clearer. What do you think?

briancroom (commented):
Great progress here! Most of my comments are on the theme of tidying up what we’ve got.

Sources/XCTest/Public/XCTestCase.swift
SeanROlszewski: Oh this is likely a change I unintentionally committed. Let me back this out.

SeanROlszewski: I'm starting to get the hang of this whole `actor` concept. 😎  I made quick work of this comment this morning.

Sources/XCTest/Public/XCTestCase.swift
```diff
@@ -15,14 +15,36 @@
...
 /// - seealso: `XCTMain`
+//@available(*, deprecated, message: "Use XCTRunnableTest")
 public typealias XCTestCaseEntry = (testCaseClass: XCTestCase.Type, allTests: [(String, XCTestCaseClosure)])
 
+public struct XCTTestMethodMapping {
+    let name: String
+    let closure: XCTestCase.ClosureType
+}
+
+public class XCTRunnableTestCase {
```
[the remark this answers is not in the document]
SeanROlszewski: I can get behind these suggestions. 👍

Sources/XCTest/Public/XCTestMain.swift
```diff
@@ -58,17 +58,58 @@
...
+//
+//    testCases.map(XCTestCaseSuite.init).forEach(currentTestSuite.addTest)
+//
+//    handleExecutionMode(executionMode,
+//                        rootTestSuite: rootTestSuite,
+//                        arguments: arguments,
+//                        observers: [])
+//}
+
+//@available(*, deprecated, message: "Use XCTMain(_ runnableTests:) instead.")
```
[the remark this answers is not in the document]
SeanROlszewski: So, I went and added them throughout the stack as it cleaned up a lot of spurious project warnings where we create a deprecation warning in our project because we're using the deprecated API. 👼

Sources/XCTest/Public/Factory_Methods.swift
```diff
@@ -0,0 +1,108 @@
...
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//
+//  Factory_Methods.swift
+//  A collection of methods to create XCTestCase.ClassEntries for use in test code.
+//
+
+/// Wrapper function allowing an array of static test case methods to fit
```
SeanROlszewski, line 14: Fix

Sources/XCTest/Public/Factory_Methods.swift
```diff
@@ -0,0 +1,108 @@
...
+    })
+}
+
+public func testCase<T: XCTestCase>(_ entries: [(String, (T) -> () throws -> Void)]) -> XCTestCase.ClassEntries {
+    return XCTestCase.ClassEntries(testCaseClass: T.self, entries: entries.map {
+        XCTestCase.Entry(name: $0.0, closure: test($0.1))
+    })
+}
+
+/// Wrapper function for the non-async, throwing variant of tests.
```
SeanROlszewski, line 39: fix

Sources/XCTest/Public/XCTestCase.swift
```diff
@@ -108,41 +110,130 @@ open class XCTestCase: XCTest {
     /// - Note: Like the designated initializer for Apple XCTest's XCTestCase,
     ///   `-[XCTestCase initWithInvocation:]`, it's rare for anyone outside of
     ///   XCTest itself to call this initializer.
-    public required init(name: String, testClosure: @escaping XCTestCaseClosure) {
```
briancroom, line 111: Explicitly overriding this initializer is quite unusual, I believe, but it seems like this has the *potential* for source breakage if somebody is overriding it. I wonder if there's anything we can do to avoid breaking change here?

Sources/XCTest/Public/XCTestMain.swift
```diff
@@ -58,17 +58,33 @@
 ///     ./FooTests FooTestCase          # Run all the tests in FooTestCase
 ///
 /// - Parameter testCases: An array of test cases run, each produced by a call to the `testCase` function
-/// - seealso: `testCase`
+/// - seealso: `testCase
```
briancroom, line 61:
```suggestion
/// - seealso: `testCase`
```

SeanROlszewski:
@swift-ci please test ❤️

Sources/XCTest/Public/XCTestCase.swift
```diff
@@ -108,41 +110,130 @@ open class XCTestCase: XCTest {
...
+            try tearDownWithError()
+        } catch {
+            recordError(error)
         }
 
-        performTearDownSequence()
+        let asyncTeardownExpectation = XCTestExpectation(description: "asyncTeardownExpectation")
+        Task {
+            defer { asyncTeardownExpectation.fulfill() }
+            try await tearDown()
```
stmontgomery, line 232: We'll need to wrap this in a `do/catch` and call `recordError` like the other, non-async tearDown API calls above

Sources/XCTest/Public/XCTestCase.swift
```diff
@@ -195,66 +286,40 @@ open class XCTestCase: XCTest {
     /// class.
     open class func tearDown() {}
 
-    private var teardownBlocks: [() -> Void] = []
-    private var teardownBlocksDequeued: Bool = false
-    private let teardownBlocksQueue: DispatchQueue = DispatchQueue(label: "org.swift.XCTest.XCTestCase.teardownBlocks")
+    private actor TeardownBlocksState {
+        
+        private var wasFinalized = false
+        private var blocks: [() async throws -> Void] = []
```
stmontgomery, line 292: I just realized that having this `Array` of `addTeardownBlock` closures include the `async` keyword will break existing behavior by executing **all** the closures via `Task {}`, because that will mean that existing, non-async teardown blocks will execute off the main thread and that could break assumptions. (This is a similar problem as the one we faced for XCTestCase closures themselves.)

I think we could employ a similar solution here of having this Array store instance of that `enum Closure` introduced elsewhere in the PR, and conditionally executing each teardown block as either `.sync` or `.async`

Sources/XCTest/Public/XCTestCase.swift
```diff
@@ -108,41 +110,130 @@ open class XCTestCase: XCTest {
...
+    
     /// Invoking a test performs its setUp, invocation, and tearDown. In
     /// general this should not be called directly.
     open func invokeTest() {
-        performSetUpSequence()
+        if skip == nil {
+            performSetup()
+            performTest()
+            performTeardown()
+        }
```
stmontgomery, line 131: Two related comments here:
- I think we should more narrowly focus this `if skip == nil` guard and have it only surround the call to `performTest`, not `performSetup` or `performTeardown`.
- I think we should remove the `var shouldInvokeTestClosure = true` stored property below entirely, and instead let the nullability of `self.skip` decide whether the test closure gets invoked.

The `shouldInvokeTestClosure` Bool property is basically redundant, in other words, because in the scenarios where it is set, `self.skip` will _also_ already be set to a non-nil value. Additionally, there is some logic elsewhere which checks `if error.xct_shouldSkipTestInvocation` and decides whether or not an Error thrown during a set up method should cause the test closure to be skipped. We should preserve that behavior, but currently the `shouldInvokeTestClosure` property gets unconditionally set to `true` upon any Error being caught

Sources/XCTest/Public/Factory_Methods.swift
```diff
@@ -0,0 +1,110 @@
...
+/// the signature required by `XCTMain`
+/// - seealso: `XCTMain`
+@available(*, deprecated, message: "Use the variant returning XCTestCase.ClassEntries instead.")
+@_disfavoredOverload
```
stmontgomery, line 45: Will using `@_disfavoredOverload` here cause the new overloads above which accept `async` closures to be preferred implicitly over their non-async counterparts? I wonder if that will break existing behaviors, by interpreting _every_ test closure as `async` and thus invoking them all via `Task {}`

Sources/XCTest/Public/XCTestCase.swift
SeanROlszewski: Ah right! Great memory. We had discussed this earlier in the PR's history on Slack. I'll add that change in.

Sources/XCTest/Public/XCTestCase.swift
SeanROlszewski: That all makes sense. I've taken care of this in the latest diff. Check it out! :)

Sources/XCTest/Public/XCTestCase.swift
SeanROlszewski: So, `TeardownBlocksState` needs to be a `class` and not an `actor` so we can make sure that teardown blocks are added to our stack from the correct thread. I've reverted the kind of type it is and used the existing `XCTWaiter.subsystemQueue.sync` method to synchronize adding to the teardown block stack. After doing this, I was able to make a new `enum` called `XCTestCase.TeardownBlocksState.BlockType` which affords us the same flexibility as the newly introduced `XCTestCase.ClosureType`. 

The reason I created a new `enum` is because teardown blocks have a different function signatures than the `XCTestCase.ClosureType` enum, and it seems like a bad design to need to figure out when to deal with some cases and not the other based on the usage site. The best thing is a new type that's used identically, but returns `Void` and is argumentless,.

Sources/XCTest/Public/Factory_Methods.swift
SeanROlszewski: I don't know if the compiler would do that. This sounds like a good question for someone on the concurrency team to explain how the overloads would evaluate.

My suspicion is it _wouldn't_ promote something which isn't explicitly `async` to `async`. That sounds very un-Swifty. But I just don't know. ¯\_(ツ)_/¯

briancroom: Stuart and I were just chatting about this - we're thinking a good next step for this would be to add a bunch of new `isMainThread` assertions into existing *non-async* test cases into the suite, in -setUp, -tearDown, the test method body, and teardown blocks, to help make sure that we're maintaining the expected threading behaviors for existing tests. If Stuart's concern here turns out to be a real problem, we expect that those new assertions would be failing. How does that sound?

SeanROlszewski: Sounds like a great plan to me! :D

millenomi:
This work is now blocking landing https://github.com/apple/swift-corelibs-foundation/pull/3036

millenomi:
(because it has asynchronous tests.)

SeanROlszewski:
@swift-ci please test ❤️

SeanROlszewski:
@swift-ci please test

u the GOAT.

Sources/XCTest/Public/XCTestCase.swift
```diff
@@ -225,33 +224,61 @@ open class XCTestCase: XCTest {
...
             }
         }
-    }
+        
+        func runTeardownBlocks(errorHandler: @escaping (Error) -> Void) {
+            for block in self.teardownBlocksState.finalize().reversed() {
+                do {
+                    try block()
+                } catch {
+                    errorHandler(error)
```
stmontgomery, line 259: I don't think we need the `errorHandler` closure to be marked `@escaping` anymore, since it's no longer being passed to a `Task {}`

Oh and actually! We don't even need it to be a parameter; we can simply reference `func handleErrorDuringTearDown` above directly
```suggestion
        func runTeardownBlocks() {
            for block in self.teardownBlocksState.finalize().reversed() {
                do {
                    try block()
                } catch {
                    handleErrorDuringTearDown(error)
```

Sources/XCTest/Public/XCTestCase.TearDownBlocksState.swift
```diff
@@ -0,0 +1,48 @@
...
+///  Supports async and sync throwing methods
+
+extension XCTestCase {
+    final class TeardownBlocksState {
+        
+        private var wasFinalized = false
+        private var blocks: [() throws -> Void] = []
+        
+        @available(macOS 12, *)
+        func append(_ block: @escaping () async throws -> Void) {
```
stmontgomery, line 25: Similar my comment on the API which calls this, I think this closure param should be marked `@Sendable`
```suggestion
        func append(_ block: @Sendable @escaping () async throws -> Void) {
```

Sources/XCTest/Public/XCTestCase.TearDownBlocksState.swift
```diff
@@ -0,0 +1,48 @@
...
+            XCTWaiter.subsystemQueue.sync {
+                precondition(wasFinalized == false, "API violation -- attempting to add a teardown block after teardown blocks have been dequeued")
+                blocks.append {
+                    try awaitUsingExpectation { try await block() }
+                }
+            }
```
stmontgomery, line 31: I'm realizing now that we can reduce duplication by implementing this async overload in terms of the non-async one, by having it call down to it:
```suggestion
            self.append {
                try awaitUsingExpectation { try await block() }
            }
```
Advantage there is that we'd consolidate the places where we do the DispatchQueue synchronization and precondition checking

Sources/XCTest/Public/XCTestCase.swift
```diff
@@ -194,26 +195,24 @@ open class XCTestCase: XCTest {
...
-
+    
     /// Registers a block of teardown code to be run after the current test
     /// method ends.
     open func addTeardownBlock(_ block: @escaping () -> Void) {
-        teardownBlocksQueue.sync {
-            precondition(!self.teardownBlocksDequeued, "API violation -- attempting to add a teardown block after teardown blocks have been dequeued")
-            self.teardownBlocks.append(block)
-        }
+        teardownBlocksState.append(block)
```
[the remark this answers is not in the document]
SeanROlszewski: Sounds good to me. It seems like a foot-gun in Swift's implementation of Effects, so I'm down to document it in the method name.

Sources/XCTest/Public/XCTestCase.swift
```diff
@@ -194,26 +195,24 @@ open class XCTestCase: XCTest {
...
+    }
+    
+    /// Registers a block of teardown code to be run after the current test
+    /// method ends.
+    @available(macOS 12.0, *)
+    open func addTeardownBlock(_ block: @Sendable @escaping () async throws -> Void) {
```
stmontgomery, line 208: I just remembered, the Darwin/Xcode copy of this API is `public` and not `open` since it's in a Swift overlay. I think it may be safer to match that here, and make this `public`. It'll be slightly more conservative which is probably safer for now anyway
```suggestion
    public func addTeardownBlock(_ block: @Sendable @escaping () async throws -> Void) {
```

briancroom (commented):
Thanks for all the iteration on this, I think it's looking good! Let's see what further work might be needed to get CI into good shape, particularly around the macOS deployment target?

briancroom:
@swift-ci please test

Sources/XCTest/Public/XCTestCase.swift
```diff
@@ -225,33 +224,61 @@ open class XCTestCase: XCTest {
             }
         }
 
-        setUp()
-    }
+        do {
```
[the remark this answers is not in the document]
stmontgomery: I believe it should be handling these correctly, due to the following two guards above in `handleErrorDuringSetUp(_:)`:

`if error.xct_shouldRecordAsTestFailure {` and
`if error.xct_shouldSkipTestInvocation {`

briancroom:
The Linux build errored with:
```
/home/buildnode/jenkins/workspace/swift-corelibs-xctest-PR-Linux/swift-corelibs-xctest/Sources/XCTest/Public/XCTestCase.swift:38:39: error: cannot find 'TeardownBlocksState' in scope
11:10:59     private let teardownBlocksState = TeardownBlocksState()
```

I had forgotten that for platforms building with CMake, we need to add a reference to any new source files to the `CMakeLists.txt` file in the root of the repository.

briancroom:
@swift-ci please test

stmontgomery:
@swift-ci please test

SeanROlszewski:
@swift-ci please test ty <3

SeanROlszewski:
@swift-ci please test ty <3

SeanROlszewski:
@swift-ci please test thank yoooouu <3

SeanROlszewski:
@swift-ci please test

briancroom:
@swift-ci please test

briancroom:
@swift-ci please test

stmontgomery (approved):

compnerd:
This seems to have regressed the Windows builds :-(
https://dev.azure.com/compnerd/3133d6ab-80a8-4996-ac4f-03df25cd3224/_apis/build/builds/54170/logs/148

PR #328, opened 2021-09-14
Fix Xcode 13 scheme warning

keith:
Xcode 13b5 warns about these being set to manual order, it doesn't
appear to be necessary in this case.


briancroom:
@swift-ci please test

briancroom:
The macOS CI run appears to have failed as a result of the XCTest and Foundation framework targets building concurrently, and I'm able to reproduce that locally as well with some regularity. Could you try additionally turning on "Find implicit dependencies" for these schemes? That ought to let the build system pick up the XCTest -> Foundation dependency and continue to build them in the correct order.

keith:
Huh, I guess I got lucky, done

briancroom:
@swift-ci please test

briancroom:
Thank you!

keith:
Thanks!

PR #329, opened 2021-09-14
Remove default arguments from private functions

keith:
Since these are only ever called internally, these arguments should
always be passed so it doesn't accidentally fall back to this file for
expanding `#file` and friends.


stmontgomery:
@swift-ci Please test

stmontgomery (approved):
Good catch, thanks @keith !

keith:
Thanks!

PR #330, opened 2021-09-26
Revert "[REVIEW] Add Async Support to XCTest"

compnerd:
Reverts apple/swift-corelibs-xctest#326

This breaks swift-corelibs-foundation on Windows; given that the retrench is so close, I think that reverting it temporarily is safer unless someone can immediately spot what is going on.

CI logs: https://dev.azure.com/compnerd/3133d6ab-80a8-4996-ac4f-03df25cd3224/_apis/build/builds/54170/logs/148


compnerd:
@swift-ci please test

compnerd:
CC: @briancroom - sadly this is somewhat urgent in light of the rebranch

stmontgomery:
@compnerd I'll go ahead and land this revert. I think we may need assistance debugging the Windows linker failure — it definitely looks related to this change, but it's not clear to me at first glance what may be going wrong.

(Side note: can you clarify what you mean by "in light of the rebranch"?)

compnerd:
@stmontgomery - the swift repositories are meant to update the LLVM base (rebranching) for updates, and it would be nice to keep everything green across that large change.

I should be able to help with the debugging, I suspect that there is a silly bug in the compiler (with the async function pointer being referenced incorrectly across module boundaries).

PR #331, opened 2021-09-28
Add Async Support to XCTest, Take 2

SeanROlszewski:
This PR contains the changes from #326, plus a change which conditionally compiles out the new `async` `setUp/tearDown` variants on Windows, since there's a compilation error in XCTest. We plan on removing this limitation once the underlying compilation error is reproduced and resolved in a CI environment.

### DIFF
--- a/Sources/XCTest/Public/XCAbstractTest.swift
+++ b/Sources/XCTest/Public/XCAbstractTest.swift
@@ -52,6 +52,9 @@ open class XCTest {
         perform(testRun!)
     }
 
+    /// Async setup method called before the invocation of `setUpWithError` for each test method in the class.
+    @available(macOS 12.0, *)
+    open func setUp() async throws {}
     /// Setup method called before the invocation of `setUp` and the test method
     /// for each test method in the class.
     open func setUpWithError() throws {}
@@ -68,6 +71,10 @@ open class XCTest {
     /// for each test method in the class.
     open func tearDownWithError() throws {}
 
+    /// Async teardown method which is called after the invocation of `tearDownWithError`
+    /// for each test method in the class.
+    @available(macOS 12.0, *)
+    open func tearDown() async throws {}
     // FIXME: This initializer is required due to a Swift compiler bug on Linux.
     //        It should be removed once the bug is fixed.
     public init() {}
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ xcuserdata
 *.xcscmblueprint
 .build/
 Output/
+Tests/Functional/.lit_test_times.txt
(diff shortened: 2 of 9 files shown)


Sources/XCTest/Public/XCAbstractTest.swift
```diff
@@ -52,6 +52,11 @@ open class XCTest {
         perform(testRun!)
     }
 
+    /// Async setup method called before the invocation of `setUpWithError` for each test method in the class.
```
SeanROlszewski, line 55: This hunk, and the hunk below that starts on 78, are the changes that this PR adds which #326 did not contain.

briancroom (approved):

stmontgomery (approved):

briancroom:
@swift-ci Please test

SeanROlszewski:
@swift-ci Please test

kirilltitov:
What about tests discovery and LinuxMain autogeneration, by the way?

MaxDesiatov:
AFAIK test discovery is implemented in SwiftPM and this question may not be relevant to this PR. Additionally, it has been enabled by default since Swift 5.4, and [`--generate-linuxmain` has been deprecated](https://github.com/apple/swift-package-manager/blob/75115d404bd8e86c597844a373a59d52828e4f8f/Sources/Commands/SwiftTestTool.swift#L539).

kirilltitov:
Of course I know, I mean, SPM doesn't know anything about `asyncTest` wrapper, does it?

joshuawright11:
Curious, would this not be available  until Swift 5.6? Considering adding some async tests to a library of mine, but plan to wait until this is publicly available.

kirilltitov:
@joshuawright11 I managed to run [async tests](https://github.com/kirilltitov/FDBSwift/blob/master/Tests/FDBTests/FDBTests.swift) in Linux using some code (in a polyfill form) from this PR. Rather dirty solution, but the alternative is nothing.

compnerd (approved):
Tested this locally on Windows after apple/swift#39998:

```
Failed Tests (1):
  SwiftXCTestFunctionalTests :: Asynchronous/Use/main.swift


Testing Time: 12.70s
  Passed: 24
  Failed:  1
```

I have an idea of how to improve the Asynchronous.Use.main test so it passes most of the time on WIndows.  Worst case, we can disable the one test on Windows.  But thank you for the support on this, it allows people to continue to be able to use XCTest uniformly on all the platforms.

The workaround that I have in mind would flush the output stream on exit.  Since the issue is more about synchronising the output, I'm less concerned about that.  We can address that subsequently after the feature is merged.  I don't think it is valuable to keep the feature waiting for the ideal state.

tomerd:
🥳

ktoso:
Wohoo!

PR #334, opened 2021-11-05
Tests: add a workaround for Windows

compnerd:
The Windows test regularly fails if the stdout is not flushed on
completion.  This increases the stability of the tests on Windows
(though there is another async related issue lurking that causes a
sporadic failure).

### DIFF
--- a/Tests/Functional/Asynchronous/Use/main.swift
+++ b/Tests/Functional/Asynchronous/Use/main.swift
@@ -3,6 +3,8 @@
 // RUN: %{xctest_checker} %t %s
 // REQUIRES: concurrency_runtime
 
+// UNSUPPORTED: OS=windows
+
 #if os(macOS)
     import SwiftXCTest
 #else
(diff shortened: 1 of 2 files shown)


compnerd:
@swift-ci please test

Tests/Functional/Asynchronous/Use/main.swift
```diff
@@ -56,7 +60,11 @@ class AsyncAwaitTests: XCTestCase {
     
     override func setUp() async throws {}
     
-    override func tearDown() async throws {}
+    override func tearDown() async throws {
```
stmontgomery, line 63: Makes sense, although I wonder if it would be ever-so-slightly more robust to place this flush in an override of the class-level `tearDown` method rather than this instance one? The class one isn't `async` but it runs later, after all tests and their `tearDown` methods complete, so it would theoretically be more inclusive in case there were stdout generated after an instance `tearDown`. I.e.
```suggestion
    override class func tearDown() {
```

stmontgomery (approved):

compnerd:
Hmm, when I went to test this, I happened to notice an assertion in the runtime.  I think that it might be that we should just temporarily disable this one test on Windows.  I can try to look into the runtime assertion subsequently.

compnerd:
@swift-ci please test

Tests/Functional/Asynchronous/Use/main.swift
compnerd: I think that I am going to disable the test on Windows instead.  I noticed that the runtime was triggering asserts, and I think that is the core issue and that the flush might be a red herring.  Sorry about the change of approach.

PR #340, opened 2021-12-27
Support XCTestExpectation creation APIs on XCTestCase from non-main threads

stmontgomery:
This matches a behavior change made in the Xcode copy of XCTest and allows tests which call `XCTestCase.expectation(description:)` — as well as other `XCTestCase` APIs that return an `XCTestExpectation` — to do so from any thread. This removes a previous requirement that tests call these APIs from the main thread, and it allows existing tests using them to adopt `async` without requiring `@MainActor`.

rdar://85344643
rdar://79163972
rdar://81024086

### DIFF
--- a/Sources/XCTest/Public/XCTestCase.swift
+++ b/Sources/XCTest/Public/XCTestCase.swift
@@ -48,6 +48,7 @@ open class XCTestCase: XCTest {
         return 1
     }
 
+    // FIXME: Once `waitForExpectations(timeout:...handler:)` gains `@MainActor`, this may be able to add it as well.
     internal var currentWaiter: XCTWaiter?
 
     /// The set of expectations made upon this test case.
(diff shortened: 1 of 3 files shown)


Sources/XCTest/Public/XCTestCase.swift
```diff
@@ -99,7 +97,10 @@ open class XCTestCase: XCTest {
         XCTCurrentTestCase = self
         testRun.start()
         invokeTest()
-        failIfExpectationsNotWaitedFor(_allExpectations)
+
+        let allExpectations = XCTWaiter.subsystemQueue.sync { _allExpectations }
```
stmontgomery, line 101: This is a drive-by fix to an apparently un-synchronized read of the `_allExpectations` property

stmontgomery:
@swift-ci Please test

briancroom (approved):
Nice, thank you for taking care of this!

PR #413, opened 2022-06-08
Remove last #filePath usage

keith:
Every other assertion function is using #file instead, which I think
will be preferred once SE-0274 is entirely rolled out.


parkera:
@swift-ci please test

parkera (approved):

keith:
Thanks!

PR #414, opened 2022-06-23
Make XCTestExpectation Sendable

stmontgomery:
This makes the `XCTestExpectation` class hierarchy conform to `Sendable`. This class and its descendants are designed to be used concurrently and are already internally concurrency-safe, so this conformance formally permits them to be used in Swift Concurrency APIs which enforce sendability.

Includes a build-only test which has concurrency warnings enabled to validate the fix.

rdar://95746454

### DIFF
--- a/Sources/XCTest/Public/Asynchronous/XCTestExpectation.swift
+++ b/Sources/XCTest/Public/Asynchronous/XCTestExpectation.swift
@@ -11,7 +11,7 @@
 //
 
 /// Expectations represent specific conditions in asynchronous testing.
-open class XCTestExpectation {
+open class XCTestExpectation: @unchecked Sendable {
 
     private static var currentMonotonicallyIncreasingToken: UInt64 = 0
     private static func queue_nextMonotonicallyIncreasingToken() -> UInt64 {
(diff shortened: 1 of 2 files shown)


stmontgomery:
@swift-ci Please test

dennisweissmann (approved):
Thanks Stuart!

briancroom (approved):

Covidnining:
> This makes the `XCTestExpectation` class hierarchy conform to `Sendable`. This class and its descendants are designed to be used concurrently and are already internally concurrency-safe, so this conformance formally permits them to be used in Swift Concurrency APIs which enforce sendability.
> 
> 
> 
> Includes a build-only test which has concurrency warnings enabled to validate the fix.
> 
> 
> 
> rdar://95746454

Covidnining:
> @swift-ci Please test

Sources/XCTest/Public/Asynchronous/XCTestExpectation.swift
```diff
@@ -11,7 +11,7 @@
 //
 
 /// Expectations represent specific conditions in asynchronous testing.
-open class XCTestExpectation {
+open class XCTestExpectation: @unchecked Sendable {
```
grynspan, line 14: This is actually dangerous, because the class is `open` and subclasses could capture data that is _not_ sendable. We should only make instances sendable when they actually _are_.

This is an unfortunate problem with the current design of the class: there's no distinction between the base class `open XCTestExpectation` and some concrete `final XCTDescribedExpectation` class that takes a description and can't possibly become unsendable.

I don't think that's solvable in the general case, and XCTest as it ships with Xcode has this same problem.

PR #420, opened 2022-09-14
Enable C language in CMakeList to avoid error finding LLVM

drodriguez:
With the absortion of
https://github.com/apple/llvm-project/commit/b66339575a9b541e67ce5ad2ba7e88da07cf9305 in stable/20220421, the LLVMConfig might try to execute C code if Terminfo library is found in the system. However the project line of XCTest does not enable the C language, and finding the package might fail in that case.

This is not a problem in Swift CI because the CI machines does not seem to have the Terminfo library (as reported in their logs during the configuration of XCTest). This might only affect developers in their local machines.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,6 +69,7 @@ set_target_properties(XCTest PROPERTIES
 if(ENABLE_TESTING)
   enable_testing()
 
+  enable_language(C)
   find_package(LLVM CONFIG)
   if(LLVM_FOUND)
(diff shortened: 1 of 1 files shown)


stmontgomery:
@swift-ci Please test

stmontgomery (approved):

stmontgomery:
Going ahead and merging, although @compnerd, feel free to chime in with any objections, since you've most frequently maintained this project's CMake configuration.

compnerd (commented):
Seems safe to me.  Thanks for the ping @stmontgomery!

PR #426, opened 2023-01-26
XCTWaiter.subsystemQueue.sync is always non-null

AtariDreams:
No need to have the null check via "?"


stmontgomery:
This is optional because the value **returned by** the closure passed to this call to `XCTWaiter.subsystemQueue.sync` is optional. It's returning the value `queue_didFulfillHandler`, which is a property whose type is an optional closure. So I believe this is correct as written

I would expect this branch to not build successfully as a result… have you tried?

PR #430, opened 2023-02-23
install: place the library in architecture-specific directories on ELF platforms

buttaface:
This is needed for apple/swift#63782, which changes the Unix toolchain to look for libraries in architecture-specific directories.


stmontgomery (approved):

buttaface:
Updated this pull to account for the recently merged #433, that also modified a few of these lines.

buttaface:
Closing since [the Swift devs want to take a different approach](https://github.com/apple/swift/pull/63782#pullrequestreview-1353429744).

PR #431, opened 2023-03-03
tests: add the ability to pass along CMAKE_Swift_FLAGS

compnerd:
Pass along any compiler flags to the tests.  This is required on Windows for testing where we add a VFS overlay through the flags now to avoid modifying the Visual Studio installation.


compnerd:
@swift-ci please test

PR #433, opened 2023-03-15
Honor CMAKE_INSTALL_***DIR variables

tristanlabelle:
This allows the CMake invocation to redirect library or binary files to other locations which is useful to place binaries where the installer build expects them in Windows build scripts.

How tested:
- Rebuilt and verified that the install paths are the same
- Rebuilt with `CMAKE_INSTALL_BINDIR` and verified that the path is honored


compnerd:
https://github.com/apple/swift-corelibs-xctest/blob/main/CMakeLists.txt#L17 - we should be able to use these variables.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test Windows platform

PR #434, opened 2023-03-15
Disambiguate the lib install path for different architectures

tristanlabelle:
The CMake install path for XCTests libs overlaps for different architectures on Windows, complicating building for multiple architectures on the same machine. This change adds an architecture subfolder, consistent with what we have for `XCTest.swiftdoc` and `XCTest.swiftmodule`.

The same problem applies to `bin`, but is better resolved by https://github.com/apple/swift-corelibs-xctest/pull/433

Co-dependent on https://github.com/apple/swift-installer-scripts/pull/181


tristanlabelle:
Abandonning in favor of handling this in the build scripts. It seems like we need CMake install to better handle Swift modules in the long term.

PR #435, opened 2023-03-29
Escape CMAKE_Swift_FLAGS to preserve Windows path separators when running tests

tristanlabelle:
Swift flags can include paths, which lost their `\` separators to CMake's escaping when running tests.


compnerd:
@swift-ci please test

compnerd (approved):

PR #437, opened 2023-04-02
Tests: add vfs overlay for dipsatch

compnerd:
Introduce an additional compile time parameter for swift builds in the tests to pass along the VFS overlay for dispatch if the build directory is present.  This ensures that we prefer the just built dispatch over the installed dispatch which is a prerequisite for permitting building Foundation et al with a build tree of dispatch when we have an installed toolchain.


compnerd:
Please test with following PRs:
https://github.com/apple/swift-corelibs-libdispatch/pull/785

@swift-ci please test

PR #446, opened 2023-06-06
Tests: add VFS overlay for Dispatch

compnerd:
Introduce an additional compile time parameter for swift builds in the tests to pass along the VFS overlay for dispatch if the build directory is present.  This ensures that we prefer the just built dispatch over the installed dispatch which is a prerequisite for permitting building Foundation et al with a build tree of dispatch when we have an installed toolchain.


grynspan (commented):
I have no objection, but this is out of my wheelhouse so I'm not sure what the various effects might be.

compnerd:
This is needed as a dependency for https://github.com/apple/swift-corelibs-libdispatch/pull/794.  This is a change to the configuration for the test suite and has no impact on the resulting binaries.

etcwilde (approved):

PR #421, opened 2022-09-30
Lower macOS availability of async APIs from 12.0 to 10.15

stmontgomery:
After async test support was added to this version of XCTest in #331, the Swift async/await feature became available in older Apple OSes, and on macOS specifically it changed from requiring 12.0 to 10.15. So this lowers the availability of all macOS 12.0 requirements of async APIs to 10.15.

PR #449, opened 2023-06-14
build_script.py: use `python3` by default

MaxDesiatov:
We're seeing sporadic CI failures, since apparently some nodes don't have `python` symlinking to a correct Python version. Changing this to `python3` will also make it consistent with the rest of the Python scripts in the Swift project.

Unblocks macOS CI check for https://github.com/apple/swift-corelibs-xctest/pull/421.


MaxDesiatov:
@swift-ci test

grynspan (approved):

PR #467, opened 2023-10-08
Stop importing CF as a framework

etcwilde:
Frameworks don't exist off of Darwin resulting in link failures in this test. The CF modulemap declares CF as a framework everywhere, which is wrong, but importing CF from swift-corelibs-foundation is also equally wrong, which this test does as well, so there's a lot of wrongness.

To get the rebranch working, we need to stop importing CoreFoundation as a framework in this test, or disable the test on non-darwin platforms. Given that corelibs-xctest is for non-Darwin platforms, that seems like the wrong course of action. Technically, things outside of swift-corelibs-Foundation should not be importing the CoreFoundation module at all.

This is affecting rebranch as clang changed modulemap import behavior in a way that affects the corelibs-Foundation CoreFoundation module.


etcwilde:
@swift-ci please test

etcwilde:
@swift-ci please test

parkera (approved):

MaxDesiatov:
@swift-ci test macos

MaxDesiatov:
@swift-ci test windows

PR #476, opened 2024-02-15
Set the minimum macOS deployment target to 10.13 for the test target

kateinoigakukun:
The lit test suite respects `MACOSX_DEPLOYMENT_TARGET` to determine the `-target` option passed to compilers. However, the `SwiftXCTestFunctionalTests` target and the `XCTest` project do not set `MACOSX_DEPLOYMENT_TARGET` in their build settings, so the default value (14.2 in the current Xcode installed on the CI) is used and the macOS version of the CI machine is older than it. This patch fixes the issue by setting `MACOSX_DEPLOYMENT_TARGET` to 10.13, which is currently used as the minimum macOS version requirement in ci.swift.org.

### DIFF
--- a/XCTest.xcodeproj/project.pbxproj
+++ b/XCTest.xcodeproj/project.pbxproj
@@ -561,6 +561,7 @@
 				DEBUG_INFORMATION_FORMAT = dwarf;
 				GCC_GENERATE_DEBUGGING_SYMBOLS = YES;
 				GCC_OPTIMIZATION_LEVEL = 0;
+				MACOSX_DEPLOYMENT_TARGET = 10.13;
 				OTHER_CFLAGS = "";
 				"OTHER_LDFLAGS[sdk=macosx*]" = (
 					"-framework",
@@ -575,6 +576,7 @@
 			isa = XCBuildConfiguration;
 			buildSettings = {
 				DEBUG_INFORMATION_FORMAT = "dwarf-with-dsym";
+				MACOSX_DEPLOYMENT_TARGET = 10.13;
 				OTHER_CFLAGS = "";
 				"OTHER_LDFLAGS[sdk=macosx*]" = (
 					"-framework",


kateinoigakukun:
@swift-ci test

kateinoigakukun:
@swift-ci test macOS

kateinoigakukun:
```
objc[11791]: Class _TtC10Foundation20_PropertyListDecoder is implemented in both /System/Library/Frameworks/Foundation.framework/Versions/C/Foundation (0x7ff847d9c528) and /Users/ec2-user/jenkins/workspace/swift-corelibs-xctest-PR-macos/branch-main/build/Ninja-DebugAssert/xctest-macosx-x86_64/Debug/SwiftFoundation.framework/Versions/A/SwiftFoundation (0x1071895c8). One of the two will be used. Which one is undefined.
```

Hmm, I guess some of the linker input uses system's Foundation instead of just built one... Not sure what brought it.

kateinoigakukun:
@swift-ci test

kateinoigakukun:
@swift-ci test

kateinoigakukun:
Hmm, I managed to get rid of CoreFoundation.framework dependency from the corelibs-foundation (In fact, we used to depend on it implicitly for a long!). However, it looks like `TARGET_OS_MAC` code paths have several subtle assumptions that work only with CoreFoundation.framework, so test suites crashed around memory management stuff. 
IIUC macOS support of swift-corelibs-foundation and swift-corelibs-xctest is just for compatibility checks, so I gave up repairing the build for now.

For those who want to fix this, I pushed my WIP fixes to the following branches:
- https://github.com/kateinoigakukun/swift-corelibs-xctest/tree/yt/explicit-min-deployment-target
- https://github.com/kateinoigakukun/swift-corelibs-foundation/tree/yt/get-rid-of-cf-dep

@parkera Do you think it's worth maintaining macOS build and checking on CI?

parkera:
On Darwin the "upward" dependency of CoreFoundation.framework on Foundation has been there since the release of macOS 14 and aligned releases. That is how we're achieving [our promise](https://www.swift.org/blog/future-of-foundation/) of a single unified implementation written in Swift (e.g., all calendar calculations are now calling into the same [open source implementation](https://github.com/apple/swift-foundation/blob/main/Sources/FoundationEssentials/Calendar/Calendar.swift)).

In parallel with your effort here we've been considering how to migrate everyone using swift-corelibs-foundation to this new implementation. I don't have a full answer yet, but it seems reasonable in the short term to focus this project on building correctly for non-Darwin platforms.

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

ikesyo:
Looks like we need to check availability (or bump the deploymen target to 10.15) for these async tests:

- https://github.com/apple/swift-corelibs-xctest/blob/e0c38689de605f45998767a4667792dbf62664e7/Tests/Functional/Asynchronous/ExpectationsWithConcurrency/main.swift
- https://github.com/apple/swift-corelibs-xctest/blob/e0c38689de605f45998767a4667792dbf62664e7/Tests/Functional/Asynchronous/Use/main.swift
- https://github.com/apple/swift-corelibs-xctest/blob/e0c38689de605f45998767a4667792dbf62664e7/Tests/Functional/Asynchronous/Expectations/main.swift

grynspan:
Sorry for the delay in reviewing!

At this time, we do not expect swift-corelibs-xctest to compile and run on Apple platforms. On Apple platforms, use Apple's XCTest.framework instead (included with Xcode and Xcode Command-Line Tools).

PR #478, opened 2024-03-10
Add support for WASI platform

kateinoigakukun:
WASI does not have thread spawning method yet, but the existing implementation blocks threads to wait async test cases synchronously. This commit introduced a new waiter method for running async test cases in single-threaded WASI environments, enabled by USE_SWIFT_CONCURRENCY_WAITER flag.

With the new waiter, `XCTMain` is async runs the given test suites without blocking the thread by bypassing some synchronous public APIs like `XCTest.perform` and `XCTest.run`. This ignores those APIs even if they are overridden by user-defined subclasses, so it's not 100% compatible with the existing XCTest APIs. This is a trade-off to support async test execution in single-threaded environments, but it should be fine because the APIs are seldom overridden by user code.


kateinoigakukun:
@swift-ci test

MaxDesiatov (approved):

MaxDesiatov:
@swift-ci test macos

kateinoigakukun:
@stmontgomery If we can get this merged before the next branch cut, it would be appreciated because this patch is the last difference from the fork. Most of the changes are guarded by a macro and should not affect other platforms. :bow:

Sources/XCTest/Public/XCTestSuite.swift
```diff
@@ -38,6 +38,27 @@ open class XCTestSuite: XCTest {
...
     }
 
+    #if USE_SWIFT_CONCURRENCY_WAITER
+    override func _performAsync(_ run: XCTestRun) async {
+        guard let testRun = run as? XCTestSuiteRun else {
+            fatalError("Wrong XCTestRun class.")
+        }
+
+        run.start()
+        func syncSetUp() { setUp() }
```
briancroom, line 48: What's the motivation for wrapping the `setUp` call in a helper function like this (and similarly below for tearDown)? Does it influence the executor that the method runs on?

CMakeLists.txt
```diff
@@ -8,7 +8,15 @@ project(XCTest LANGUAGES Swift)
...
 option(USE_FOUNDATION_FRAMEWORK "Use Foundation.framework on Darwin" NO)
 
-if(NOT CMAKE_SYSTEM_NAME STREQUAL Darwin)
+set(USE_SWIFT_CONCURRENCY_WAITER_default NO)
+
+if(CMAKE_SYSTEM_PROCESSOR STREQUAL wasm32)
+  set(USE_SWIFT_CONCURRENCY_WAITER_default ON)
+endif()
+
+option(USE_SWIFT_CONCURRENCY_WAITER "Use Swift Concurrency-based waiter implementation" "${USE_SWIFT_CONCURRENCY_WAITER_default}")
```
briancroom, line 17: Does this option work on other platforms besides WASM/WASI? I wonder if that would be a path to improving test coverage for this new code, by running the existing lit-based tests on Darwin or Linux, with this flag turned on?

Sources/XCTest/Public/Asynchronous/XCTWaiter.swift
```diff
@@ -187,9 +189,16 @@ open class XCTWaiter {
     ///   these environments. To ensure compatibility of tests between
     ///   swift-corelibs-xctest and Apple XCTest, it is not recommended to pass
     ///   explicit values for `file` and `line`.
+    #if USE_SWIFT_CONCURRENCY_WAITER
+    @available(*, unavailable, message: "Expectation-based waiting is not available when using the Swift concurrency waiter.")
```
briancroom, line 193: I'm a little bit confused about the messaging of the `USE_SWIFT_CONCURRENCY_WAITER` option, the continued availability of the `XCTWaiter` class, and these unavailability messages about "expectation-based waiting".

Is there anything left that `XCTWaiter` *can* do in this mode? Or should the entire class be removed / made unavailable?

Sources/XCTest/Public/XCAbstractTest.swift
```diff
@@ -36,20 +36,43 @@ open class XCTest {
     /// testRunClass. If the test has not yet been run, this will be nil.
     open private(set) var testRun: XCTestRun? = nil
 
+    internal var performTask: Task<Void, Never>?
```
briancroom, line 39: This is only used in `USE_SWIFT_CONCURRENCY_WAITER` mode, right? Can it move inside the `#if` block?

Sources/XCTest/Public/XCTestCase.swift
```diff
@@ -325,18 +416,31 @@ private func test<T: XCTestCase>(_ testFunc: @escaping (T) -> () throws -> Void)
...
         }
     }
+#endif
 }
 
 @available(macOS 12.0, *)
 func awaitUsingExpectation(
     _ closure: @escaping () async throws -> Void
 ) throws -> Void {
+#if !USE_SWIFT_CONCURRENCY_WAITER
```
briancroom, line 443: Can this guard move up a few lines to enclose the entire function declaration?

Sources/XCTest/Public/XCTestCase.swift
```diff
@@ -114,9 +135,46 @@ open class XCTestCase: XCTest {
         self.testClosure = testClosure
     }
 
+    #if USE_SWIFT_CONCURRENCY_WAITER
+    internal func _invokeTestAsync() async {
+        await performSetUpSequence()
+
+        do {
+            if skip == nil {
+                try testClosure(self)
```
briancroom, line 144: It's very important for behavioral consistency in this mode that when calling out to *non-*`async` user-provided code (test methods, and `setUp` and `tearDown` overrides) that those synchronous functions run on the main actor. I'm not certain, but I don't think that invariant is maintained by the current code. Can you comment on that?

(This was something that we struggled with a bit, and initially got wrong when we were working on support for `async` test methods a while back in https://github.com/apple/swift-corelibs-xctest/pull/326)

briancroom (changes requested):
Thank you for the PR and all the work on this, @kateinoigakukun! (And @MaxDesiatov as well, it sounds like.) I've left a number of questions and comments, some pretty important.

I wish there were a way to achieve this with less branching, but without `reasync` support in the compiler, I doubt we can do much better.

Besides the details of the code changes in the PR, I'm interested in learning more about the plans for ongoing support and maintenance of this compilation mode and platform. In particular, what kind of CI and testing is in place, or expected to come in the near future?

Sources/XCTest/Public/XCAbstractTest.swift
kateinoigakukun: Ah, yes, this should be inside #if block

Sources/XCTest/Public/XCTestCase.swift
kateinoigakukun: I think we can keep the consistency by adding `@MainActor` on this method.

kateinoigakukun:
> I wish there were a way to achieve this with less branching, but without reasync support in the compiler, I doubt we can do much better.

Yeah, that's a very unfortunate situation to have many branching here due to function-colors. I agree that we can improve it with `reasync` support in the future.

> Besides the details of the code changes in the PR, I'm interested in learning more about the plans for ongoing support and maintenance of this compilation mode and platform. In particular, what kind of CI and testing is in place, or expected to come in the near future?

Yes, at least I'm going to include XCTest and Foundation build in the WebAssembly build CI after this will get merged: https://ci.swift.org/job/oss-swift-pr-test-crosscompile-wasm-ubuntu-20_04/

Also, as you mentioned, we can build and run tests with this mode on other platforms.

Sources/XCTest/Public/XCTestSuite.swift
briancroom: Ohhhhh yes, of course. Thanks for that reminder.

CMakeLists.txt
briancroom: Cool. It's fine for that to come afterwards, but I would like to see it happen!

Sources/XCTest/Public/Asynchronous/XCTWaiter.swift
briancroom: Do you think it would be workable to adjust the scope and semantics of this conditional a little bit, perhaps calling it `DISABLE_XCTWAITER` or something along those lines, and have it guard the entirety of the `XCTWaiter` and `XCTestExpectation` classes, along with the everything in `XCTestCase+Asynchronous.swift`, and the expectation-related state tracking in `XCTestCase.swift`.

I do see that usage of the `XCTWaiter.subsystemQueue` has leaked out to non-waiter code in a few places (`ThrownErrorWrapper` and `TeardownBlocksState`). Those are using it for just basic locking purposes, and quite independently from the actual waiter subsystem. Perhaps we introduce a ` XCTestCase.subsystemQueue` to use for those instead? Or even a different locking solution altogether? This would be a general-purpose change, not conditionalized by the new `USE_SWIFT_CONCURRENCY_WAITER`/`DISABLE_XCTWAITER` option.

grynspan: If we introduce a second queue, we should be very wary of deadlocks, data races, etc. where the current code assumes a single global lock.

kateinoigakukun: Right, I used the new queue in `TeardownBlocksState` and `ThrownErrorWrapper`. As far as I checked, they are well independent from the other, and they won't acquire the other lock while acquiring itself lock. But I need more eyes here :pray:

kateinoigakukun:
@swift-ci test

Sources/XCTest/Public/Asynchronous/XCTWaiter.swift
briancroom: Agreed, considering the fine-grained scoping of the locking provided by the queue for these, I think we should be ok in that regard.

Sources/XCTest/Private/XCTestCase.TearDownBlocksState.swift
```diff
@@ -12,29 +12,42 @@ extension XCTestCase {
...
                 try awaitUsingExpectation { try await block() }
             }
+            #endif
         }
 
         func append(_ block: @escaping () throws -> Void) {
-            XCTWaiter.subsystemQueue.sync {
+            XCTestCase.subsystemQueue.sync {
                 precondition(wasFinalized == false, "API violation -- attempting to add a teardown block after teardown blocks have been dequeued")
                 blocks.append(block)
```
briancroom, line 45: In `DISABLE_XCTWAITER` mode, this is implicitly converting the synchronous `block` to an `async` block, right? I think we also need to wrap that in an `@MainActor` closure to preserve the expected execution context on it?

kateinoigakukun: Ah, nice catch!

kateinoigakukun:
@swift-ci test

briancroom:
Let’s take a good look at CI results, but the diff looks ok to me now! Thanks for discussing and the quick turnaround on the revisions.

briancroom (approved):

MaxDesiatov:
@swift-ci test windows

kateinoigakukun:
Thanks a lot for your review @briancroom! I assume we need @shahmishal's help to get this merged by temporarily bypassing PR status check, right?

kkk669:
This change might have broken some tests especially if there are many test cases. (I'm sorry if I was wrong.)

- https://github.com/kkk669/wasm-test-example
- https://github.com/kkk669/swift-syntax/pull/18

kateinoigakukun:
@kkk669 Yes, the implementation is changed from the version used in the downstream, so some of large async test cases might not work. It will be mitigated after tail-call support.

Issue #473, opened 2024-01-10
Misleading string-ified test failure output


dabrahams:
When this test fails:

```swift
      XCTAssertEqual(g.rules.count, 1)
```

The output is 

> XCTAssertEqual failed: ("3") is not equal to ("1")

The quotes around the values are both needless and misleading; they give the impression that I was testing for string equality.

grynspan:
I think we'd still want the quotes for anything conforming to `StringProtocol`, but I agree they (and probably the parentheses too?) aren't immediately meaningful for other data types.

abrarShariar:
The parenthesis might be helpful to separate out the value we are asserting vs. the logging text.

grynspan:
I don't morally object, though I know a lot of code that integrates with XCTest is sensitive to formatting changes. @briancroom what do you think?

PR #482, opened 2024-03-24
Removing additional quotes around values in a XCTAssertEqual

abrarShariar:
https://github.com/apple/swift-corelibs-xctest/issues/473


abrarShariar:
Taking it to draft to gather feedback from team and add additional enhancement. https://github.com/apple/swift-corelibs-xctest/issues/473#issuecomment-2016714708

PR #483, opened 2024-03-28
Use the package swift-corelibs-foundation

parkera:
Update `Package.swift` to use the `package` branch of `swift-corelibs-foundation`.

Note: This is work done on a new branch of `swift-corelibs-xctest`, also called `package`. It doesn't affect the main branch.


briancroom:
Cool, thanks for looking at this, @parkera!

briancroom (approved):

PR #485, opened 2024-04-17
Add `XCTAssertNoThrowAsync` and `XCTAssertThrowsErrorAsync`

### LINKED ISSUE #419: Support async calls with XCTAssertNoThrow, XCTAssertThrowsError, XCTExpectFailure...
Hello,

I'm trying to use `XCTAssertNoThrow` method with async function. I'm trying to use it as this:

```swift
        XCTAssertNoThrow {
            let item = try await map.getItem(with: "āž$*")
            XCTAssertNil(item)
        }
```

It turns out the callback is never run. In my case `try` returns with an error, which is thrown from `getItem`.

I ended up with this solution, which is suboptimal:

```swift
        var caughtError: Error?
        do {
            let item = try await map.getItem(with: "āž$*")
            XCTAssertNil(item)
        } catch {
            caughtError = error
        }
        XCTExpectFailure("test fails for the time being") {
            XCTAssertNil(caughtError)
        }
```

Side note: `XCTExpectFailure` doesn't run the async closure as well.

jiyimeta:
Add `XCTAssertNoThrowAsync` and `XCTAssertThrowsErrorAsync`, which are the async versions of `XCTAssertNoThrow` and `XCTAssertThrowsError`.

I adopted different names from sync ones in order to avoid confusion of overloads.
For example, if the sync and async versions have the same name `XCTAssertThrowsError`,
```swift
func testExample() async {
    XCTAssertThrowsError(try functionThatDoesThrowError())
}
```
such a sync assertion in an async context will cause compile error with message `Expression is 'async' but is not marked with 'await'`, despite success before this change.

Resolves #419.

### DIFF
--- a/Sources/XCTest/Public/XCTAssert.swift
+++ b/Sources/XCTest/Public/XCTAssert.swift
@@ -28,7 +28,9 @@ private enum _XCTAssertion {
     case `false`
     case fail
     case throwsError
+    case throwsErrorAsync
     case noThrow
+    case noThrowAsync
 
     var name: String? {
         switch(self) {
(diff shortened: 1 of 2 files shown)


briancroom:
Hi @jiyimeta, thanks for your interest in improving XCTest. Unfortunately we aren't accepting new APIs here on the swift-corelibs-xctest repository, because its primary purpose is to provide a compatible implementation of the APIs provided by the XCTest that ships with Xcode, for use when targeting platforms that Xcode doesn't support.

I recommend that you keep these additional APIs in your own project/package, and file an Xcode enhancement request at http://bugreport.apple.com.

PR #493, opened 2024-06-24
Add new include path into Foundation when building tests

jmschonfeld:
Now that Foundation is being re-cored ontop of swift-foundation, we're adding some new (internal) C modules. XCTest's tests add an explicit include path into the Foundation build directory to pickup these modules, but it does not account for the new location of these modules. Instead of adding more paths into Foundation's build directory, I'm adding one new path to a new folder named `_CModulesForClients`. We're updating the Foundation build to place all C modules that XCTest's tests need into this folder, so that the tests here need to only include this directory.

For now, I'm leaving the old paths as-is in case they are load-bearing with the current Foundation build. I've validated that all XCTest tests pass in my local build with Foundation re-cored and will use swift-ci to validate that this does not break the existing build either.

### DIFF
--- a/Tests/Functional/lit.cfg
+++ b/Tests/Functional/lit.cfg
@@ -99,7 +99,7 @@ else:
         '-L', os.path.join(foundation_dir, 'lib'),
         '-I', foundation_dir,
         '-I', os.path.join(foundation_dir, 'swift'),
-        '-I', os.path.join(foundation_dir, '_CModulesForClients',
+        '-I', os.path.join(foundation_dir, '_CModulesForClients'),
         '-Xcc', '-F', '-Xcc', foundation_dir,
     ])
 
--- a/Tests/Functional/lit.cfg
+++ b/Tests/Functional/lit.cfg
@@ -99,6 +99,7 @@ else:
         '-L', os.path.join(foundation_dir, 'lib'),
         '-I', foundation_dir,
         '-I', os.path.join(foundation_dir, 'swift'),
+        '-I', os.path.join(foundation_dir, '_CModulesForClients',
         '-Xcc', '-F', '-Xcc', foundation_dir,
     ])
 


jmschonfeld:
@swift-ci please test

parkera (approved):

briancroom:
Out of curiosity, do you know if other clients/scenarios need this as well? I don’t have a good sense of what (if anything) is special about the way that these tests are built.

briancroom (approved):

jmschonfeld:
> Out of curiosity, do you know if other clients/scenarios need this as well? I don’t have a good sense of what (if anything) is special about the way that these tests are built.

XCTest and LLBuild are the only projects that depend on Foundation but build/run tests before Foundation is installed into the destination toolchain directory. Other projects like SwiftPM just automatically pickup these modules from the installed location, but XCTest and LLBuild depend on Foundation's build directory directly since Foundation hasn't been installed yet. I also ran LLBuild's tests and it builds and runs tests successfully without any changes, so it doesn't appear that they are relying on this type of include path the same way that XCTest is (but if I encounter issues I'll go take a look to see if LLBuild needs the same treatment)

stmontgomery (approved):

jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test Windows platform

jmschonfeld:
macOS failed with:

```
dyld[71565]: Symbol not found: _swift_getTypeByMangledNameInContext2
  Referenced from: <CD22481B-792E-329E-98AC-4017FD4CD50F> /Users/ec2-user/jenkins/workspace/swift-corelibs-xctest-PR-macos/branch-main/swift-corelibs-xctest/Tests/Functional/Asynchronous/Misuse/Output/Misuse (built for macOS 14.2 which is newer than running OS)
  Expected in:     <087DADAD-D590-35EA-AC4D-7B0E966804EB> /usr/lib/swift/libswiftCore.dylib
```

I don't think this is related to my change, it looks like this is a misconfigured deployment target? I think this is the same issue that SCL-F was hitting. Can this failure be ignored?

PR #503, opened 2024-09-05
[6.0.1] Guard parse_version

shahmishal:
The most recent versions of `parse_version` throw an exception if the version is empty. The version passed in is only set on Darwin (call to `mac_ver()`, so it's causing test failures on newer versions of Linux since the test suite can't even start.

Now, the only reason for the version parse is because the tests are looking at whether or not concurrency is available on the OS. This is only a limitation if we're working with Darwin. Swift 5.10 on Windows and Linux always have a Swift 5.10 concurrency runtime, so we don't even need to check for a version.

rdar://128502662
(cherry picked from commit 65e6ecdce29101129911c49760b0f85ac18d21e5) (cherry picked from commit f26bc2138b8cd8f73317d7b7c019bf016ca5bc33)


shahmishal:
@swift-ci test

PR #509, opened 2025-03-13
build: avoid doubly nesting the swift module

compnerd:
We always install the swift module into the platform directory. The additional nesting is not required.


compnerd:
@swift-ci please test

stmontgomery (approved):

PR #514, opened 2025-09-11
build: optionalise Foundation and dispatch

compnerd:
Rather than requiring the build tree for the dependencies, optionalise them. This allows us to build against an existing SDK or against the build tree. This flexibility allows us to change the build order when building the Windows toolchain against a SDK that is being built up. Without this change, we must provide a dispatch build tree which will inject the modulemap for the dispatch runtime which conflicts with the verison in the SDK that is being built against.


compnerd:
@swift-ci please smoke test

compnerd:

stmontgomery (approved):

PR #522, opened 2025-12-06
Clean up expectations when done before stopping the test run

LumioseSil:
When a test creates expectations but never waits on them (e.g., expectation(forNotification:) without a corresponding wait(for:timeout:)), those expectations are never cleaned up. 

> Note from reviewer @stmontgomery:
>
> As one example of the consequence of this bug, the handler for an unwaited NSNotification expectation may be called for a notification during an unrelated test. I haven’t tried, but I expect there would be a similar problem for XCTNSPredicateExpectation.

### DIFF
--- a/Sources/XCTest/Public/XCTestCase.swift
+++ b/Sources/XCTest/Public/XCTestCase.swift
@@ -141,6 +141,8 @@ open class XCTestCase: XCTest {
         let allExpectations = XCTWaiter.subsystemQueue.sync { _allExpectations }
         failIfExpectationsNotWaitedFor(allExpectations)
 
+        cleanUpExpectations()
+
         testRun.stop()
         XCTCurrentTestCase = nil
     }
--- a/Tests/Functional/Asynchronous/Expectations/Cleanup/main.swift
+++ b/Tests/Functional/Asynchronous/Expectations/Cleanup/main.swift
@@ -1,0 +1,57 @@
+// RUN: %{swiftc} %s -o %T/ExpectationCleanup
+// RUN: %T/ExpectationCleanup > %t || true
+// RUN: %{xctest_checker} %t %s
+
+#if os(macOS)
+    import SwiftXCTest
+#else
+    import XCTest
+#endif
+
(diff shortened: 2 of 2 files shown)


grynspan:
Thanks for filing @AZero13. PRs should include descriptions that explain the purpose of the change. If there is a GitHub issue associated with this change, please include it here.

We cannot accept PRs without explanation. Thanks!

AZero13:
> Thanks for filing @AZero13. PRs should include descriptions that explain the purpose of the change. If there is a GitHub issue associated with this change, please include it here.
> 
> We cannot accept PRs without explanation. Thanks!

Done!

grynspan:
@swift-ci test

grynspan:
Can you add a unit test for this scenario?

AZero13:
> Can you add a unit test for this scenario?

Done!

grynspan:
@swift-ci test

Sources/XCTest/Public/Asynchronous/XCTNSPredicateExpectation.swift
```diff
@@ -97,7 +97,7 @@ open class XCTNSPredicateExpectation: XCTestExpectation {
             }
         }
 
-        runLoop.add(timer, forMode: .default)
+        runLoop.add(timer, forMode: RunLoopMode.defaultRunLoopMode)
```
grynspan: Were you testing on macOS? This library is not expected to compile or run on Apple targets. The correct symbol name is [`default`](https://github.com/swiftlang/swift-corelibs-foundation/blob/main/Sources/Foundation/RunLoop.swift#L35).

AZero13: Okay, so let's restart the tests then!

grynspan:
@swift-ci test

AZero13:
@grynspan Fixed the error!

grynspan:
@swift-ci test

AZero13:
Had to fix formatting and forgot a few check lines that are there now. I apologize @grynspan

plemarquand:
@swift-ci test

grynspan:
@swift-ci test Windows

PR #523, opened 2026-01-05
Write back expectation to self.state

LumioseSil:
waitingState needs to be written back to state.


grynspan:
Would you mind fleshing out the PR description a bit more? In particular, can you describe what bug you're fixing ("without this change, X does Y and Z doesn't W" or similar). It would also help us if you could explain how you came across that bug and then what let you to this solution. Thanks!

grynspan:
No reply since January. Closing this PR. Please feel free to reopen when you've added more info as requested. Thanks!

PR #524, opened 2026-01-23
DNM: Intentionally introduce breaking changes to see if CI works as expected

jerryjrchen:
I tested this in a container locally using the build_script, and it would find the copy of libXCTest.so in /usr/lib/swift/ because it's prepended to the runpath of the test executables built by lit.

If CI is correctly using the built copy of libXCTest.so, then it should see test failures with this breaking change.

Do not merge this please!!!


jerryjrchen:
@swift-ci test

jerryjrchen:
Closing since CI does appear to fail as expected.

PR #531, opened 2026-03-04
`XCTWaiter` should spin the run loop in common modes, not default.

grynspan:
`XCTWaiter` spins the run loop in the default mode, which is insufficient to avoid a deadlock if another subsystem schedules work in another common mode.

I don't have a unit test for this because it's non-deterministic and may take many test runs to trigger. @stmontgomery and I verified at-desk that the hang goes away with this change in place.

Resolves rdar://139710145.


grynspan:
@swift-ci test

grynspan:
Nope.

PR #541, opened 2026-03-30
Guard _TestingInterop link on SwiftTesting_FOUND

rjmansfield:
find_package(SwiftTesting CONFIG) is non-REQUIRED so it's possible to build without swift-testing and fail linking _TestingInterop

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -81,10 +81,12 @@ if(DISABLE_XCTWAITER)
 endif()
 
 find_package(SwiftTesting CONFIG)
-target_link_libraries(XCTest PRIVATE
-  _TestingInterop)
-target_compile_definitions(XCTest PRIVATE
-  XCT_BUILD_WITH_INTEROP)
+if(SwiftTesting_FOUND)
+  target_link_libraries(XCTest PRIVATE
+    _TestingInterop)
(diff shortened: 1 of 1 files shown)


rjmansfield:
@swift-ci Please test

grynspan:
Before we merge, we should figure out under what conditions we expect XCTest is available but Swift Testing is not (if any).

rjmansfield:
> Before we merge, we should figure out under what conditions we expect XCTest is available but Swift Testing is not (if any).

I hit this building without `--swift-testing`. The code is conditional with `XCT_BUILD_WITH_INTEROP`, this just reflects the linking in cmake to be similar. Or if it's now a hard dependency, the cmake should be `find_package(SwiftTesting CONFIG REQUIRED)`

grynspan:
I understand, I just want to quantify the effects of the change. :)

stmontgomery:
Swift's `build-script` does have a large number of presets and it's reasonable to expect that some of them don't include `swift-testing`, so I'm comfortable proceeding with this.

stmontgomery (approved):

jerryjrchen (approved):
Tested this against some of my in progress changes and it looks good. Thanks!

PR #542, opened 2026-04-14
Send interop event if XCTest is not the active test library

jerryjrchen:
### Motivation:

Enable reporting of XCTAssert failures in a Swift Testing test.

### Modifications:

* Update XCTAssert impl to encode issues and send to installed fallback event handler if no active XCTestCase
* Test both directions of interop as well as the more general case of issue handling without an active test case.

### DIFF
--- a/Sources/XCTest/Public/XCTAssert.swift
+++ b/Sources/XCTest/Public/XCTAssert.swift
@@ -95,14 +95,117 @@ private func _XCTEvaluateAssertion(_ assertion: _XCTAssertion, message: @autoclo
     case .success:
         return
     default:
+        // XCTest assertions record failures in a few scenarios which require separate treatment.
+        //
+        // 1) Within the body of an XCTestCase test: the test case itself records the failure.
+        //
+        // 2) While Swift Testing* tests are running: the other test library records the failure.
+        // We encode the failure into an event and call the fallback event handler.
+        // (*or other test library that participates in interop)
+        //
+        // 3) Neither: recorded failure outside of test case.
+        // We drop this kind of failure since there is currently no affordance for recording it.
+        //
+        // This is timing-dependent on global XCTCurrentTestCase state.
+        // For example, the following test _may_ lose the error depending on when the task executes
+        // relative to the end of test case execution.
+        //
+        // func testDetached() {
+        //   Task.detached {
+        //     XCTFail("Might be lost")
+        //   }
+        // }
+        let description = "\(result.failureDescription(assertion)) - \(message())"
         if let currentTestCase = XCTCurrentTestCase {
             currentTestCase.recordFailure(
-                withDescription: "\(result.failureDescription(assertion)) - \(message())",
+                withDescription: description,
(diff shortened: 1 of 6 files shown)


jerryjrchen:
@swift-ci Please test

Sources/XCTest/Public/XCTAssert.swift
```diff
@@ -95,14 +95,113 @@ private func _XCTEvaluateAssertion(_ assertion: _XCTAssertion, message: @autoclo
...
+        // (*or other test library that participates in interop)
+        //
+        // 3) Neither: an XCTest test is running, but we aren't in the body of a test case.
+        // We drop this kind of failure since there is currently no affordance for recording it.
+        //
+        // An example is error recording within Task.detached, e.g.
+        //
+        // Task.detached {
+        //   XCTFail("this is lost")
+        // }
```
stmontgomery: This doesn't _always, deterministically_ drop the issue; it _may_, but in XCTest, since the current XCTestCase is tracked using global mutable state instead of Task-local state, it also may record the issue by coincidence of timing. So it might be better to phrase this example in terms of timing relative to the end of a test.

Sources/XCTest/Public/XCTAssert.swift
```diff
@@ -95,14 +95,113 @@ private func _XCTEvaluateAssertion(_ assertion: _XCTAssertion, message: @autoclo
...
+        return Interop.Event.Instant(
+            absolute: d / .seconds(1),
+            since1970: Date().timeIntervalSince1970
+        )
+    }()
+
+    let sourceLocation = {
+        let fileName = (filePath as NSString).lastPathComponent
+        // Module name is not available to us, so unconditionally fill with <unknown>
+        let fileID = "<unknown>/\(fileName)"
```
stmontgomery, line 177: I suppose if we plumbed down the `StaticString` we could inspect it to find the module name 🤔

Sources/XCTest/Public/XCTAssert.swift
jerryjrchen: Yeah, good point. I used semaphores in the integration test to force the behaviour since I noticed it wasn't consistent without that

Sources/XCTest/Public/XCTAssert.swift
jerryjrchen: I could add module parsing logic, but I think we're preferring `filePath` now, right? In which case I'd rather keep the code simpler and fill in a placeholder.

stmontgomery: Agreed; this probably isn't worthwhile even if it were possible. Dropping this suggestion

stmontgomery (approved):

jerryjrchen:
@swift-ci Please test

jerryjrchen:
CI blocked by https://github.com/swiftlang/swift-testing/pull/1676

jerryjrchen:
@swift-ci Please test

PR #549, opened 2026-06-16
Access Testing SPI in CMake built of corelibs XCTest

jerryjrchen:
Access Testing SPI for interop when performing CMake build

Begin sharing some code between XCTest and Swift Testing in the interop
implementation.

## Motivation

In a follow-up, XCTest will favour using ABI.Record in Testing over the
Interop.OutputRecord, which will lead to lower maintenance following
future record schema changes.

## Modifications

* Decode event version with ABI.version in the fallback handler

  Essentially no functional change, as Swift Testing still always
  encodes the fallback events as version 6.3.

  This is only available when building in CMake against the Swift
  Testing module, e.g. a full toolchain build. Otherwise, continue to
  check event version is exactly version 6.3.

* Define XCT_CAN_USE_TESTING_SPI in CMake

  The Swift Testing module now shares SPI following
  https://github.com/swiftlang/swift-testing/pull/1755

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -85,9 +85,11 @@ find_package(SwiftTesting CONFIG)
 find_package(SwiftTestingMacros CONFIG)
 if(SwiftTesting_FOUND)
   target_link_libraries(XCTest PRIVATE
-    _TestingInterop)
+    _TestingInterop
+    Testing)
   target_compile_definitions(XCTest PRIVATE
-    XCT_BUILD_WITH_INTEROP)
+    XCT_BUILD_WITH_INTEROP
+    XCT_CAN_USE_TESTING_SPI)
 endif()
 
 if(USE_FOUNDATION_FRAMEWORK)
--- a/Sources/XCTest/Private/Interop/InteropHandler.swift
+++ b/Sources/XCTest/Private/Interop/InteropHandler.swift
@@ -101,11 +101,8 @@ extension Interop.Handler {
         // When building for the toolchain, support ABI versions 6.3 to 6.4.
 #if XCT_CAN_USE_TESTING_SPI
         guard
-            let schemaVersionNumber = String(
-                validatingCString: recordJSONSchemaVersionNumber
-            ).flatMap(ABI.VersionNumber.init),
-            (ABI.v6_3.versionNumber...ABI.v6_4.versionNumber).contains(schemaVersionNumber),
-            ABI.version(forVersionNumber: schemaVersionNumber) != nil else {
+            let schemaVersionNumber = ABI.VersionNumber.init(validatingCString: recordJSONSchemaVersionNumber),
+            (ABI.v6_3.versionNumber...ABI.v6_4.versionNumber).contains(schemaVersionNumber) else {
             return
         }
 #else
--- a/Sources/XCTest/Private/Interop/InteropHandler.swift
+++ b/Sources/XCTest/Private/Interop/InteropHandler.swift
@@ -19,6 +19,9 @@
 #else
     import Foundation
 #endif
+#if XCT_CAN_USE_TESTING_SPI
+@_spi(ForToolsIntegrationOnly) private import Testing
+#endif
 
 enum Interop {}
 
@@ -94,10 +97,23 @@ extension Interop.Handler {
     /// It can only report test issues if there is an active XCTest test case.
     static let ourFallbackEventHandler: FallbackEventHandler = {
         recordJSONSchemaVersionNumber, recordJSONBaseAddress, recordJSONByteCount, _ in
+        // Check that this schema version number is supported.
+        // When building for the toolchain, support ABI versions 6.3 to 6.4.
+#if XCT_CAN_USE_TESTING_SPI
+        guard
+            let schemaVersionNumber = String(
+                validatingCString: recordJSONSchemaVersionNumber
+            ).flatMap(ABI.VersionNumber.init),
+            (ABI.v6_3.versionNumber...ABI.v6_4.versionNumber).contains(schemaVersionNumber),
+            ABI.version(forVersionNumber: schemaVersionNumber) != nil else {
+            return
+        }
+#else
         guard let schemaVersion = String(validatingCString: recordJSONSchemaVersionNumber),
                 schemaVersion == "6.3" else {
             return
         }
+#endif
 
         // Memory is managed by the caller of the fallback event handler, so do
         // not attempt to deallocate when done.
--- a/Sources/XCTest/Private/Interop/InteropHandler.swift
+++ b/Sources/XCTest/Private/Interop/InteropHandler.swift
@@ -101,7 +101,7 @@ extension Interop.Handler {
         // When building for the toolchain, support ABI versions 6.3 to 6.4.
 #if XCT_CAN_USE_TESTING_SPI
         guard
-            let schemaVersionNumber = ABI.VersionNumber.init(validatingCString: recordJSONSchemaVersionNumber),
+            let schemaVersionNumber = ABI.VersionNumber(validatingCString: recordJSONSchemaVersionNumber),
             (ABI.v6_3.versionNumber...ABI.v6_4.versionNumber).contains(schemaVersionNumber) else {
             return
         }


jerryjrchen:
@swift-ci Please test

jerryjrchen:
@swift-ci Please test

Sources/XCTest/Private/Interop/InteropHandler.swift
```diff
@@ -94,10 +97,23 @@ extension Interop.Handler {
     /// It can only report test issues if there is an active XCTest test case.
     static let ourFallbackEventHandler: FallbackEventHandler = {
         recordJSONSchemaVersionNumber, recordJSONBaseAddress, recordJSONByteCount, _ in
+        // Check that this schema version number is supported.
+        // When building for the toolchain, support ABI versions 6.3 to 6.4.
+#if XCT_CAN_USE_TESTING_SPI
+        guard
+            let schemaVersionNumber = String(
```
grynspan, line 104: I'd suggest breaking this `guard` up a bit as it's somewhat hard to read.

Do you want an `ABI.VersionNumber.init(validatingCString:)` convenience?

Sources/XCTest/Private/Interop/InteropHandler.swift
```diff
@@ -94,10 +97,23 @@ extension Interop.Handler {
     /// It can only report test issues if there is an active XCTest test case.
     static let ourFallbackEventHandler: FallbackEventHandler = {
         recordJSONSchemaVersionNumber, recordJSONBaseAddress, recordJSONByteCount, _ in
+        // Check that this schema version number is supported.
+        // When building for the toolchain, support ABI versions 6.3 to 6.4.
```
grynspan, line 101: If you use `ABI.decodeEvent(fromRecordJSON:in:)`, I think we can simplify this code significantly.

Sources/XCTest/Private/Interop/InteropHandler.swift
jerryjrchen: `ABI.VersionNumber.init(validatingCString:)` would help!

I can also drop the `ABI.version(forVersionNumber: schemaVersionNumber) != nil` since the version isn't used right now.

Sources/XCTest/Private/Interop/InteropHandler.swift
jerryjrchen: Makes sense, although I'm going to save that for a follow-up where we actually integrate the Swift Testing Event type. I figure this PR is mostly focusing on the build script changes and proving that we can indeed `import Testing` in corelibs XCTest

Sources/XCTest/Private/Interop/InteropHandler.swift
```diff
@@ -19,6 +19,9 @@
 #else
     import Foundation
 #endif
+#if XCT_CAN_USE_TESTING_SPI
+@_spi(ForToolsIntegrationOnly) @_implementationOnly private import Testing
```
stmontgomery, line 23: I've not seen `@_implementationOnly` combined with `private` on an `import`; sounds like the compiler accepts it, but I'm surprised `private` alone is insufficient

CMakeLists.txt
```diff
@@ -85,9 +85,13 @@ find_package(SwiftTesting CONFIG)
...
 if(SwiftTesting_FOUND)
   target_link_libraries(XCTest PRIVATE
-    _TestingInterop)
+    _TestingInterop
+    Testing)
   target_compile_definitions(XCTest PRIVATE
-    XCT_BUILD_WITH_INTEROP)
+    XCT_BUILD_WITH_INTEROP
+    # After https://github.com/swiftlang/swift-testing/pull/1755,
+    # we can import @_spi visible symbols from this module
```
stmontgomery, line 93: This feels like a comment we could omit from the CMake file and only place here in the PR, since it's describing the logistics of landing cross-repo work

jerryjrchen: Sg, I'll make sure to include it in the commit message if it's not already there

Sources/XCTest/Private/Interop/InteropHandler.swift
jerryjrchen: I agree that it seems like only private should be needed. Let me update that and check the build results

jerryjrchen:
@swift-ci Please test

jerryjrchen:
@swift-ci Please test

jerryjrchen:
@swift-ci Please test Windows

jerryjrchen:
@swift-ci Please test Windows

Sources/XCTest/Private/Interop/InteropHandler.swift
jerryjrchen: Fails to build on Windows, and I think this is related?

```
    C:\Users\swift-ci\jenkins\workspace\swift-corelibs-xctest-PR-windows\build\tmp\lsp-test\D625109B\test.swift:1:8: error: missing required module 'Testing'


    1 | import XCTest


      |        `- error: missing required module 'Testing'
```

jerryjrchen: sourcekit-lsp tests compile with an `import XCTest` but don't link Testing. I think this can be fixed with `@_implementationOnly` at compile time, but someone linking `XCTest ` will now be required to also have `Testing` available at load time. Is that something we can guarantee?

jerryjrchen:
@swift-ci Please test

jerryjrchen:
@swift-ci Please test Windows

Sources/XCTest/Private/Interop/InteropHandler.swift
jerryjrchen: Based on CI results, the `@_implementationOnly` attribute is required to compile this change on Windows

grynspan: `@_implementationOnly` shouldn't be needed anymore, so this is weird. Might want to ask the compiler folks what's up.

Sources/XCTest/Private/Interop/InteropHandler.swift
```diff
@@ -94,10 +97,20 @@ extension Interop.Handler {
     /// It can only report test issues if there is an active XCTest test case.
     static let ourFallbackEventHandler: FallbackEventHandler = {
         recordJSONSchemaVersionNumber, recordJSONBaseAddress, recordJSONByteCount, _ in
+        // Check that this schema version number is supported.
+        // When building for the toolchain, support ABI versions 6.3 to 6.4.
+#if XCT_CAN_USE_TESTING_SPI
+        guard
+            let schemaVersionNumber = ABI.VersionNumber.init(validatingCString: recordJSONSchemaVersionNumber),
```
grynspan:
```suggestion
            let schemaVersionNumber = ABI.VersionNumber(validatingCString: recordJSONSchemaVersionNumber),
```

Sources/XCTest/Private/Interop/InteropHandler.swift
stmontgomery: This was discussed earlier: https://github.com/swiftlang/swift-corelibs-xctest/pull/549#discussion_r3858472407

stmontgomery (approved):

Sources/XCTest/Private/Interop/InteropHandler.swift
grynspan: I saw, but I didn't see any resolution?

grynspan: @compnerd Any idea why Windows wants `@_implementationOnly` here?

jerryjrchen:
@swift-ci Please test

jerryjrchen:
@swift-ci Please test Windows

jerryjrchen:
@swift-ci Please test

jerryjrchen:
I updated the sourcekit LSP tests, so now just need to verify that it unblocks this change.

Sources/XCTest/Private/Interop/InteropHandler.swift
jerryjrchen: This is no longer needed after my change here: https://github.com/swiftlang/sourcekit-lsp/pull/2760

